Issue #46, opened 2018-04-04
Provide stable API for setting context, etc.


axw:
The elasticapm.Transaction, Span, and Error types currently expose the underlying model types of the same name. This will pose a problem if/when the model changes. Instead of exposing those types directly, we should provide a stable API for adding context, setting error details, etc.

- [x] elasticapm.Error
- [x] elasticapm.Transaction
- [x] elasticapm.Span
- [x] introduce elasticapm.Context type/methods
- [x] introduce stacktrace.Frame

~~elasticapm.Tracer.Service~~ For now we'll continue to expose model.Service, as it's relatively simple and unlikely to change significantly, and we don't anticipate that modifying it will be a common thing anyway.

axw:
I think we're good now, will close this and open more specific issues as they're found.

PR #49, opened 2018-04-05
all: introduce and use stacktrace.Frame

axw:
Introduce stacktrace.Frame, a distilled
version of runtime.Frame, with only the
information we care about. Errors may
now implement

  type interface {
      StackTrace() []stacktrace.Frame
  }

as an alternative to pkg/errors, rather
than directly referencing the model
stacktrace frame type. This moves us a
step closer to not exposing the model
types in any if the public/stable API.

Stacktrace functions have been changed
to append to provided slices, enabling
memory reuse.

### DIFF
--- a/contrib/apmlambda/lambda.go
+++ b/contrib/apmlambda/lambda.go
@@ -5,7 +5,6 @@ import (
 	"net"
 	"net/rpc"
 	"os"
-	"path/filepath"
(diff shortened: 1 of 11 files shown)


axw:
Partially addresses https://github.com/elastic/apm-agent-go/issues/46

PR #60, opened 2018-04-28
resolve composite literal uses unkeyed fields

graphaelli:


axw:
Thanks! I should probably set up the prechecks to run vet and lint too.

graphaelli:
That's how this was caught - I vendored apm-agent-go in another project and it failed the build.

PR #92, opened 2018-05-17
elasticapm: StartSpan always returns non-nil

axw:
Transaction.StartSpan and elasticapm.StartSpan now
always return non-nil, even if the transaction is nil.
This simplifies callsites so they don't have to nil
checking of spans.


axw:
Note that this is not all roses. With this, non-sampled transactions will now incur a bit more overhead due to the allocations. We could return a singleton for dropped spans, but then we could have multiple goroutines racing on updates to the span object. The overhead is kept to a minimum by using a sync.Pool, so I think this is a reasonable trade-off for this safer and neater API.

PR #95, opened 2018-05-22
add func to set and get transaction id

ckahi:
add func to set and get transaction id

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -83,6 +83,14 @@ type Transaction struct {
 	spans []*Span
 }
 
+func (tx *Transaction) GetID() string {
+	return tx.model.ID
+}
+
+func (tx *Transaction) SetID(id string) {
+	tx.model.ID = id
+}
+
 func (tx *Transaction) setID() {
 	if tx.model.ID != "" {
 		return


axw:
@ckahi can you please explain what problem this solves?

axw:
Expanding on my question:

When a transaction is sent to the server, an ID is automatically generated by the agent code. I don't see a need for explicitly getting or setting the ID. If you have a use case, please describe it to help me understand the motivation behind this PR.

ckahi:
@axw sorry for my previous simple description

In an API request, the current service needs to call more children, then I want to take advantage of span in the management backstageto show the time line function to the calling process as an invocation chain , so I need to get and set the current transaction ID and span ID  in each service in order to make sure with the request of the source belongs to a transaction;

In addition, I found that it is not enough to only add transaction ID Settings and access methods, and also need to do the same for span func.

Do you have a better idea for my idea?

axw:
@ckahi thanks for the update. I think I understand.

The current API and data model does not support distributed tracing -- we're working on it now. You can see some progress here: https://github.com/elastic/apm-server/issues/913.

Transactions and spans are specific to a single service. What we'll most likely be doing is introducing trace IDs, and adding a parent ID to transactions. Trace IDs are a common ID associated with all transactions and spans within the same trace (it gets assigned at the root, and passed down to each child and their ancestors).

When you create a transaction, you'll have the option of passing in [trace context](https://w3c.github.io/distributed-tracing/report-trace-context.html): trace ID, parent transaction/span ID, and trace options (e.g. sampling). If you don't pass in a trace ID, then one will be generated by the library. I haven't yet thought too deeply about how the API will look, but I expect something like this:

```go
traceContext := elasticapm.ParseTraceContext("...")
tx := tracer.StartTransaction("name", "type", elasticapm.WithTraceContext(traceContext))
```

The apmhttp module (and other web framework modules) will be updated to do this, by extracting and parsing HTTP headers. We'll also provide functions for capturing outgoing HTTP requests as spans, and injecting HTTP headers into them.

Will these changes meet your needs?

ckahi:
@axw  I'm very glad to receive your reply. What you plan to develop is exactly what I need.

axw:
> @axw I'm very glad to receive your reply. What you plan to develop is exactly what I need.

Great! I'm going to close this PR then. Thank you for sending it anyway. I've just opened https://github.com/elastic/apm-agent-go/issues/99, in case you want to follow along or contribute.

Finally, if you haven't already seen it, there's a survey out asking people about their requirements: https://docs.google.com/forms/d/e/1FAIpQLScbW7D8m-otPO7cxqeg7XstWR8vMnxG6brnXLs_TFVSTHuHvg/viewform?usp=sf_link. If you have a few minutes, I'd appreciate your input!

PR #106, opened 2018-05-28
elasticapm: Done->End, Span.Start->Timestamp

### LINKED ISSUE #104: Rename {Span,Transaction}.Done to End, drop duration
The other APM agents call it "end", rather than "done". We will follow suit.

Also, we'll drop the time.Duration parameter from those methods, since in most cases developers will want to have Done/End calculate the duration. Instead, expose a Duration field on the types.

### LINKED ISSUE #105: Rename Span.Start to Span.Timestamp
In the intake API it's called start, but that's an offset from the transaction's timestamp. We should call the field Timestamp for consistency with the Transaction type, to tidy up the API.

axw:
Rename Span.Done and Transaction.Done methods to End. Drop the time.Duration parameter, and instead expose a time.Duration field on the types, initialised to a negative value. If it is negative at the time End is called, set it to the time elapsed since the Timestamp field.

Also, rename Span.Start to Span.Timestamp to be consistent with the Transaction type.

Closes #104 
Closes #105

### DIFF
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -69,18 +69,20 @@ transaction.Context.SetCustom("key", "value")
 See <<context-api>> for more details on setting transaction context.
 
 [float]
-[[transaction-done]]
-==== `func (*Transaction) Done(time.Duration)`
+[[transaction-end]]
+==== `func (*Transaction) End()`
 
-Done sets the transaction's duration to the specified value, and enqueues it for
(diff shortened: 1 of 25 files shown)


axw:
Fair warning, the name may change from End to something else in the future, so we can align the agent APIs. Ruby can't have "end" as it's a keyword.

PR #129, opened 2018-07-03
module/apmsql: parse SQL (WIP)

axw:
Still TODO:
 - PostgreSQL $tag$ quotes
 - more tests (see https://github.com/elastic/apm-agent-python/blob/master/tests/instrumentation/psycopg2_tests.py for a good assortment)


axw:
Added some more tests based on those in apm-agent-python. Also added a simple fuzz test, and benchmark.

PR #135, opened 2018-07-11
 Report system and process CPU/memory stats

axw:
Add builtin metrics which report the system and process CPU usage, and memory usage and availability.

We vendor go-sysinfo and friends. Note that these packages are only ever used internally, never showing up in the public API. Thus, vendoring them will not interfere with the program's dependencies.


axw:
_Maybe_ it's just a matter of education, but I feel like providing the ".norm" values is worse than providing non-normalized values. The interpretation still depends on the number of CPUs on the machine: 25% of 1 CPU is very different to 25% of 4 CPUs.

axw:
> Maybe it's just a matter of education, but I feel like providing the ".norm" values is worse than providing non-normalized values. The interpretation still depends on the number of CPUs on the machine: 25% of 1 CPU is very different to 25% of 4 CPUs.

Either way we should probably report the number of CPUs, and then it doesn't really matter; multiply or divide.

PR #151, opened 2018-07-20
Pare back the metrics API

axw:
For the initial implementation of metrics, we won't support specifying metric types or units, nor will we support summary metrics (in the API, schema, or wire protocol; bridges may still transform and send them). This is in line with existing metrics in the Elastic Stack (namely Metricbeat), and how they are today interpreted in Kibana.

Also, rename some of the builtin metrics to align with Metricbeat.


axw:
> code lgtm. It looks like the gc stats are either wrong here or in the metricbeat module - still tracking that down.

As discussed live, will land this as-is and sort that out in a followup.

PR #170, opened 2018-08-07
 module/apmgorm: introduce GORM instrumentation

### LINKED ISSUE #169: module/apmgorm: instrumentation module for jinzhu/gorm
See #168. gorm is very popular; we will provide a module to simplify instrumenting calls into it.

axw:
Package apmgorm provides a means of instrumenting [gorm](http://gorm.io)
database operations.

By using apmgorm.Open instead of gorm.Open, we obtain a *gorm.DB which
can be used with apmgorm.WithContext. The WithContext function can be
used to propagate context (i.e. containing a transaction) to callbacks
which report spans.

We provide "dialects" packages which import the gorm/dialects namesake
packages, as well as register the apmsql drivers. The latter is required
in order to parse DSNs.

Closes elastic/apm-agent-go#169


Azuka:
That was fast!

axw:
> That was fast!

:)

If you try it out and find any issues, I'll be very happy to hear your feedback. I desperately wanted to find a way to sneak a context through to apmsql, but to no avail. Hopefully one day gorm will add context propagation.

Azuka:
Oh, I'm already using it ;-). I can verify that it works, and that update and insert queries don't expose query params.

PR #202, opened 2018-09-05
Update import path prefix, top-level package name

### LINKED ISSUE #162: Update import paths to go.elastic.co/apm

axw:
The import path prefix is now "go.elastic.co/apm",
and the top-level package for this repo's name is
"apm" to match the directory name.

Closes elastic/apm-agent-go#162


axw:
apm-ci won't succeed until we update the jjb definition to clone into `go.elastic.co/apm`
can't really do that until we're ready to stop building the 0.5 branch, which is probably a bit premature

axw:
Updated PR to target v2.

PR #250, opened 2018-10-03
model: add leading slash to URL path if missing

axw:
If a URL path is missing a leading slash, add it
in while encoding. We do it here rather than when
building the model object to avoid allocation.


axw:
I found this while investigating adding instrumentation to https://github.com/ipfs/go-ipfs. Requests were showing up with no leading slash on the paths. We were just reporting whatever path was supplied via the request URI line.

Issue #260, opened 2018-10-16
Change SANITIZE_FIELD_NAMES and IGNORE_URLS to wildcard syntax


axw:
To align with other agents, we will implement the simpler wildcard syntax used by the Java agent.

PR #261, opened 2018-10-16
ELASTIC_APM_SANITIZE_FIELD_NAMES and ELASTIC_APM_IGNORE_URLS now use wildcard matching

axw:
Implement wildcard matching, and update `ELASTIC_APM_SANITIZE_FIELD_NAMES` and `ELASTIC_APM_IGNORE_URLS` to use that instead of regular expressions. This will bring is in alignment with the Java agent.


axw:
Closed #260

PR #292, opened 2018-10-29
[Jenkins] Go Agent Pipeline version 0

kuisathaverat:
image


Jenkinsfile
```diff
@@ -0,0 +1,335 @@
...
+              }
+              
+              on_merge {
+                echo "build cause a merge"
+              }
+              
+              on_pull_request {
+                echo "build cause PR"
+              }
+              gitCreateTag()
```
axw: What's the tag going to be? I'd generally prefer that CI does not modify the repo.

Jenkinsfile
```diff
@@ -0,0 +1,335 @@
...
+          agent { label 'linux' }
+          environment {
+            PATH = "${env.PATH}:${env.HUDSON_HOME}/go/bin/:${env.WORKSPACE}/bin"
+            GOPATH = "${env.WORKSPACE}"
+          }
+          
+          when { 
+            beforeAgent true
+            allOf { 
+              //branch 'master';
```
axw:
```suggestion
```
?

Jenkinsfile
```diff
@@ -0,0 +1,335 @@
...
+          agent { label 'linux && docker' }
+          environment {
+            PATH = "${env.PATH}:${env.HUDSON_HOME}/go/bin/:${env.WORKSPACE}/bin"
+            GOPATH = "${env.WORKSPACE}"
+          }
+          
+          when { 
+            beforeAgent true
+            allOf { 
+              //branch 'master';
```
axw:
```suggestion
```
?

Jenkinsfile
```diff
@@ -0,0 +1,335 @@
...
+        
+        /**
+         run Go integration test with the commit version on master branch.
+        */
+        stage('Integration test master') { 
+          agent { label 'linux' }
+          when { 
+            beforeAgent true
+            allOf { 
+              //branch 'master';
```
axw:
```suggestion
```
?

Jenkinsfile
```diff
@@ -0,0 +1,335 @@
...
+              environment name: 'integration_test_master_ci', value: 'true' 
+            }
+          }
+          steps {
```
axw: Is this going to run integration tests with the server and all the agents on each build of the Go agent? That sounds like it's going to be very slow. Can we make that bit periodic, rather than triggering on each build?

scripts/jenkins/docker-test.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+set -ex
+export GOPATH=$WORKSPACE
+eval "$(gvm $GO_VERSION)"
+go get -v -u github.com/jstemmer/go-junit-report
+go get -v -u github.com/axw/gocov/gocov
```
axw:
```suggestion
```

(github.com/axw/gocov/... further down is sufficient)

scripts/jenkins/test.sh
```diff
@@ -0,0 +1,19 @@
...
+eval "$(gvm $GO_VERSION)"
+go get -v -u github.com/jstemmer/go-junit-report
+go get -v -u github.com/axw/gocov/gocov
+go get -v -u github.com/matm/gocov-html
+go get -v -u github.com/axw/gocov/...
+go get -v -u github.com/AlekSi/gocov-xml
+
+go get -v -t ./...
+
+export COV_FILE="build/coverage.cov"
```
axw: I don't think we need coverage for both this and the docker tests. Having it for the docker tests should be enough, since those tests are a superset.

scripts/jenkins/docker-test.sh
```diff
@@ -0,0 +1,27 @@
...
+
+go get -v -t ./...
+
+export COV_FILE="build/coverage.cov"
+export OUT_FILE="build/test-report.out"
+mkdir -p build
+
+./scripts/docker-compose-testing up -d --build
+./scripts/docker-compose-testing run -T --rm go-agent-tests make coverage | tee ${COV_FILE}.raw
+echo "mode: atomic" > ${COV_FILE}
```
axw: I think gocov (which I wrote before `go test -cover` existed) can be replaced to make this a bit simpler, cutting down the number of external dependencies.

- https://github.com/t-yuki/gocover-cobertura can convert ${COV_FILE}.raw into XML output for the Cobertura plugin.
- `go tool cover -html=${COV_FILE}.raw -o build/coverage-apm-agent-go-docker-report.xml` will produce an HTML report

Jenkinsfile
```diff
@@ -0,0 +1,335 @@
...
+                  doGenerateSubmoduleConfigurations: false, 
+                  extensions: [],
+                  submoduleCfg: [],
+                  userRemoteConfigs: [[credentialsId: "${JOB_GIT_CREDENTIALS}", 
+                  url: "${GIT_URL}"]]])
+              }
+              env.JOB_GIT_COMMIT = getGitCommitSha()
+              env.JOB_GIT_URL = "${GIT_URL}"
+              
+              github_enterprise_constructor()
```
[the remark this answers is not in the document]
kuisathaverat: yep, it is a utility step to easily manage changes on Github, it creates some environment variables

Jenkinsfile
kuisathaverat: it creates a tag to use in the integration test to test against the exact version, it is deleted at the end of the job. It is easy to work with tags instead commit sha1, and all the integration test that we have right now are based on branchs, so you can use also tags.

Jenkinsfile
kuisathaverat: this step will executed only on the master branch, for the moment is comments until we test all the pipeline.

Jenkinsfile
kuisathaverat: ditto

Jenkinsfile
kuisathaverat: ditto

Jenkinsfile
kuisathaverat: no, only on master branch, on a PR it does not make sense

scripts/jenkins/test.sh
kuisathaverat: this one thing I have to ask you, there are differences between two coverage reports, the docker one has more files, these are they:

```
go.elastic.co/apm/apmtest/httpsuite.go
go.elastic.co/apm/apmtest/testlogger.go
go.elastic.co/apm/apmtest/withtransaction.go
go.elastic.co/apm/internal/apmcontext/context.go
go.elastic.co/apm/internal/apmdebug/debug.go
go.elastic.co/apm/internal/apmschema/schema.go
go.elastic.co/apm/internal/radix/radix.go
go.elastic.co/apm/transport/transporttest/err.go
go.elastic.co/apm/transport/transporttest/recorder.go
```

What is it the good coverage test?
I did not check the unit test executed but probably are different too.

scripts/jenkins/docker-test.sh
kuisathaverat: IMHO the html coverage output of [gocov-xml](https://github.com/matm/gocov-html) than the default one, we can open an issue related with that and let the whole team decided.

what is the difference/benefit to use [gocover-cobertura](https://github.com/t-yuki/gocover-cobertura) instead [gocov-xml](https://github.com/AlekSi/gocov-xml)? I saw that gocover-cobertura is updated more recently, I do not have a strong opinion the tool that you said is better, it is better for me ;)

kuisathaverat:
yep, as soon as Infra reprovision the Jenkins instance this pipeline will appear in the apm-agent-go-mbp project as a PR, then we would test it on apm-ci. I will try to it happens in the next two days.

Jenkinsfile
axw: So this just gets passed down to apm-integration-testing, to specify the branch? If I update apm-integration-testing to accept a commit instead of (or as well as) a branch, can we get rid of the tagging?

The branch is currently ignored by the Go service in apm-integration-testing anyway, so I need to make updates either way.

scripts/jenkins/docker-test.sh
axw: > IMHO the html coverage output of gocov-xml than the default one, we can open an issue related with that and let the whole team decided.

Sure, we can put it up for a vote. I'd just prefer to use more of the standard tooling, makes it easier for newcomers. Having fewer tools means less to maintain, too.

> what is the difference/benefit to use gocover-cobertura instead gocov-xml? I saw that gocover-cobertura is updated more recently, I do not have a strong opinion the tool that you said is better, it is better for me ;)

The main benefit is that gocover-cobertura doesn't require gocov, it uses the output of the built-in "go tool -cover".

scripts/jenkins/test.sh
axw: It's because of the "-coverpkg" flag.

`go test -cover ./...` will run the tests for each package in subdirectories of `./`, but for each package it only includes coverage information of its own package. i.e. test coverage for `go.elastic.co/apm/module/apmhttp` only contains coverage information for `go.elastic.co/apm/module/apmhttp`.

In the docker case, it runs "make coverage", which runs scripts/test_coverage.sh. In that script you can see we add a flag, `-coverpkg=go.elastic.co/apm/...`. This instructs the go command to include coverage for all packages under `go.elastic.co/apm/...`

That script is slightly more complex than needed here, since it also works with older versions of Go. You could just run `go test ./... -v -coverprofile="${COV_FILE}" -coverpkg=go.elastic.co/apm/...` to get the same effect, for recent versions of Go.

Jenkinsfile
```diff
@@ -0,0 +1,335 @@
...
+              
+              on_change{
+                echo "build cause a change (commit or PR)"
```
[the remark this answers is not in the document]
kuisathaverat: For the moment I do not have a real use for those methods, but I want to see which ones are triggered on each event, "on_change" for example will be triggered on a PR creation and on any commit push.

Jenkinsfile
kuisathaverat: I made the changes to manage branches, as do all the modules, we can open a team discussion about use tags or commits, but it only push tags so does not change anything else I do not see it as a problem, tags are for this kind of thing.

scripts/jenkins/test.sh
kuisathaverat: so if we add `-coverpkg=go.elastic.co/apm/...` to the test.sh (test stage) both will be the same, ok, in that case the docker test can be removed.

scripts/jenkins/docker-test.sh
kuisathaverat: fair enough, you convinced me ;)

scripts/jenkins/test.sh
axw: > so if we add -coverpkg=go.elastic.co/apm/... to the test.sh (test stage) both will be the same, ok, in that case the docker test can be removed.

They might cover the same files, but not the same lines. The docker tests run a super-set of tests: integration tests that only run when certain environment variables are set (e.g. PGHOST). The docker one should be kept, and this one removed.

Jenkinsfile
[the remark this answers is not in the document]
axw: OK. Might need to update the wording, it's a little bit confusing. Can be done later.

Jenkinsfile
axw: Let's bring it up with the team.

Go's tooling uses tags for identifying module versions, and GitHub uses tags for releases; using them for transient state feels odd to me if/when we can propagate the commit. Also, perhaps I'm a bit overcautious, but I like CI to have as few permissions as possible. Being able to write, even if it does not write, makes me nervous.

scripts/jenkins/test.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -e
+export GOPATH=$WORKSPACE
+eval "$(gvm $GO_VERSION)"
+go get -v -u github.com/jstemmer/go-junit-report
+go get -v -u code.google.com/p/go.tools/cmd/cover
```
axw: This shouldn't be necessary, I think the install instructions for gocover-cobertura are out of date. "go tool cover" is part of the Go toolchain, no need to install anything from go.tools. Just keep the line belowto install gocover-cobertura

axw:
```suggestion
```

I think the install instructions for gocover-cobertura are out of date. "go tool cover" is part of the Go toolchain, no need to install anything from go.tools. Just keep the line below to install gocover-cobertura

scripts/jenkins/test.sh
kuisathaverat: gotcha, it is the other way around

Jenkinsfile
kuisathaverat: Sure 😄

kuisathaverat:
@axw The creation of tags was removed, so the integration tests are disabled for now, the other stuff I think that was resolved, thus I think we are ready to Go ;)

Jenkinsfile
```diff
@@ -0,0 +1,335 @@
...
+          }
+        }
+        stage('Docker tests') {
+          agent { label 'linux && docker && immutable' }
+          environment {
+            PATH = "${env.PATH}:${env.HUDSON_HOME}/go/bin/:${env.WORKSPACE}/bin"
+            GOPATH = "${env.WORKSPACE}"
+          }
+          
+          when {
```
axw: Can we run docker-test after the parallel stages? I have it like that in Travis CI so that coverage reports aren't published until the unit tests pass on all platforms.

Jenkinsfile
```diff
@@ -0,0 +1,335 @@
...
+  triggers {
+    cron('0 0 * * 1-5')
+    githubPush()
+  }
+  options {
+    timeout(time: 1, unit: 'HOURS') 
+    buildDiscarder(logRotator(numToKeepStr: '3', artifactNumToKeepStr: '2', daysToKeepStr: '30'))
+    timestamps()
+    preserveStashes()
+    //ansiColor('xterm')
```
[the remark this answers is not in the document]
kuisathaverat: There is a bug right now and it only works on BO, or classic UI, so for the moment we only can see colors in the BO UI, when the PR will be merged I will enable it again.

see https://issues.jenkins-ci.org/browse/JENKINS-11752, https://issues.jenkins-ci.org/browse/JENKINS-39536, https://issues.jenkins-ci.org/browse/JENKINS-54133 and jenkinsci/ansicolor-plugin#132

Jenkinsfile
```diff
@@ -0,0 +1,335 @@
...
+              not { branch 'master' };
+              environment name: 'integration_test_pr_ci', value: 'true' 
+            }
+          }
+          steps {
+            build(
+              job: 'apm-server-ci/apm-integration-test-pipeline', 
+              parameters: [
+                string(name: 'BUILD_DESCRIPTION', value: "${BUILD_TAG}-INTEST"),
+                string(name: 'APM_AGENT_GO_PKG', value: "${BUILD_TAG}"),
```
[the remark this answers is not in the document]
kuisathaverat: I also did something for the AgentGoNetHttp but it only supports branches and tags, I am gonna change it to support SHA1

https://github.com/elastic/apm-integration-testing/pull/188/files#diff-d0df5dd5f0019c3b878fde30f4b7341bR762

Jenkinsfile
```diff
@@ -0,0 +1,335 @@
...
+            environment name: 'test_ci', value: 'true' 
+          }
+          steps {
+            withEnvWrapper() {
+              unstash 'source'
+              dir("${BASE_DIR}"){    
+                sh """#!/bin/bash
+                ./scripts/jenkins/test.sh
+                """
+                codecov('apm-agent-go')
```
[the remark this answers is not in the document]
kuisathaverat: sure

scripts/jenkins/build.sh
```diff
@@ -0,0 +1,13 @@
...
+srcdir=`dirname $0`
+test -z "$srcdir" && srcdir=.
+. ${srcdir}/common.bash
+
+jenkins_setup
+
+go get -u -v golang.org/x/tools/cmd/goimports
+go get -u -v golang.org/x/lint/golint
+
+make install check
```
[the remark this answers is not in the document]
kuisathaverat: ok

scripts/jenkins/common.bash
```diff
@@ -0,0 +1,113 @@
...
+    get_go_version
+  fi
+
+  # Setup Go.
+  export GOPATH=${WORKSPACE}
+  export PATH=${GOPATH}/bin:${PATH}
+  eval "$(gvm ${GO_VERSION})"
+
+  # Workaround for Python virtualenv path being too long.
+  export TEMP_PYTHON_ENV=$(mktemp -d)
```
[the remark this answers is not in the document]
kuisathaverat: yep, it a common script coming from beats, when I finish with all agents I have to come back to it and see how to share it between projects

scripts/jenkins/docker-test.sh
```diff
@@ -0,0 +1,30 @@
...
+./scripts/docker-compose-testing up -d --build
+./scripts/docker-compose-testing run -T --rm go-agent-tests make coverage | tee ${COV_FILE}.raw
+
+echo "mode: atomic" > ${COV_FILE}
+grep -v "mode\: atomic" ${COV_FILE}.raw >> ${COV_FILE}
+
+go tool cover -html="${COV_FILE}" -o build/coverage/coverage-apm-agent-go-docker-report.html
+gocover-cobertura < "${COV_FILE}" > build/coverage/coverage-apm-agent-go-docker-report.xml
+
+./scripts/docker-compose-testing run -T --rm go-agent-tests go test -race ./... -v 2>&1 | tee ${OUT_FILE}
```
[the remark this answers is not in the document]
kuisathaverat: nice pipe ;)

Jenkinsfile
kuisathaverat: in that case, we can move out the test stage, so we will run build, then test, and finally in parallel benchmarks and docker-test

kuisathaverat:
The only decision pending, it is to set which steps only run on the master branch (or other branches) and which steps only run on PRs, at the moment this is may suggestion

checkout - all
build - all
test - all
benchmarks - master
docker-test - all
integration test (PR vs server stable) - PR
integration test Axis (agent vs the versions matrix) - master
documentation - master

The integration test and axis test are disabled until we implemented the SHA1 support on the integration tests

Jenkinsfile
```diff
@@ -0,0 +1,333 @@
...
+          dir("${BASE_DIR}"){    
+            sh """#!/bin/bash
+            ./scripts/jenkins/test.sh
+            """
+          }
+        }
+      }
+      post { 
+        always {
+          coverageReport("${BASE_DIR}/build/coverage")
```
axw: coverageReport isn't needed here anymore right?

Jenkinsfile
```diff
@@ -0,0 +1,335 @@
...
+            }
+          }
+          steps {
+            withEnvWrapper() {
+              unstash 'source'
+              dir("${BASE_DIR}"){    
+                sh """#!/bin/bash
+                ./scripts/jenkins/bench.sh
+                """
+                sendBenchmarks(file: 'build/bench.out', index: "benchmark-go")
```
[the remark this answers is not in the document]
axw: OK. Will be interested to see if we can come up with a common schema.

Jenkinsfile
axw: SGTM

scripts/jenkins/test.sh
```diff
@@ -0,0 +1,22 @@
...
+
+go get -v -u github.com/jstemmer/go-junit-report
+go get -v -u github.com/t-yuki/gocover-cobertura
+go get -v -t ./...
+
+export COV_FILE="build/coverage/coverage.cov"
+export OUT_FILE="build/test-report.out"
+mkdir -p build/coverage
+
+(go test -race ./... -v -coverprofile="${COV_FILE}" -coverpkg=go.elastic.co/apm/... 2>&1 | tee ${OUT_FILE}) || echo -e "\033[31;49mTests FAILED\033[0m"
```
axw: still running coverage here, this should just be `go test -v -race ./...`, and the cover/gocover-cobertura bits below removed

axw:
Please squash before merging. Maybe hold off until we set the default branch to master?

kuisathaverat:
sure, we have to squash it before merge, there is no rush, so we can merge it next week.

PR #299, opened 2018-11-01
Introduce Transaction.EnsureParent

### LINKED ISSUE #298: Add support for RUM agent correlation
We need a way of generating parent IDs for transactions, and providing them to the RUM agent. We may also want to provide some templating utilities for sending transaction IDs to the RUM agent.

axw:
Introduce a new method, Transaction.EnsureParent, which returns
the transaction's parent ID. If the transaction does not have a
parent span recorded, then one is recorded based on the trace ID.

This change is in support of correlation with the RUM agent, for
applications where the initial page load is served by the backend
agent.

Closes elastic/apm-agent-go#298

### DIFF
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -109,6 +109,51 @@ transaction.End()
 TraceContext returns the transaction's <<trace-context, trace context>>.
 
 [float]
+[[transaction-ensureparent]]
+==== `func (*Transaction) EnsureParent() SpanID`
+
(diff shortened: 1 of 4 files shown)


felixbarny:
How can users convert the IDs to a string when rendering a template?

axw:
> How can users convert the IDs to a string when rendering a template?

There's an example in the PR of using it in a template. The ID types have String methods which are called when printing them in a template.

PR #313, opened 2018-11-09
 Introduce default internal logger

### LINKED ISSUE #278: Add default, operator-configurable logger
We currently have an API for enabling logging, but this requires code changes. We should also provide a default logger which can be configured using environment variables.

axw:
Tracer is now initialized with an internal logger
implementation if ELASTIC_APM_LOG_FILE is specified.
Without setting this environment variable, logging
is disabled as before.

Setting ELASTIC_APM_LOG_FILE to stdout or stderr will
cause logging to write to stdout/stderr respectively.
Otherwise, specifying a file path will cause that file
to be created/truncated and written to.

By default, only errors will be logged. Setting
ELASTIC_APM_LOG_LEVEL to "debug" will cause debug
messages to be printed in addition.

Closes #278


internal/apmlog/logger.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	var logWriter io.Writer
+	switch strings.ToLower(fileStr) {
+	case "stdout":
+		logWriter = os.Stdout
+	case "stderr":
+		logWriter = os.Stderr
+	default:
+		f, err := os.Create(fileStr)
```
graphaelli: I think this is the right call

axw:
> Is the log guaranteed to be flushed before the process exits?

This is just meant for debugging, so I don't expect that's necessary. We'd have to either flush after every log message (_maybe_ OK given the expected low volume, but preferable not to), or say flush when the tracer is closed. The user would then be responsible for closing the tracer before exiting the process. I'm going to leave this for now.

> Any plans to promote 404s to apm-server to error? Had to turn on debug to find those.

That would be sensible wouldn't it? I'll fix that before landing.

PR #326, opened 2018-11-22
StartSpanOptions ignored options parameter

ChristophPech:


axw:
@ChristophPech can you please explain what problem this is addressing? Preferably by adding a test case which fails without / passes with the change.

We already do this in Transaction.StartSpanOptions, which is eventually called by apm.StartSpan:

https://github.com/elastic/apm-agent-go/blob/797c31c516c1935ae3ed0d1281777f8a81da46a6/span.go#L68-L73

ChristophPech:
I am using go-pg and wrote a callback to trace postgres calls. I have to set the StartTime retroactively because there is only a "finished" callback so I used apm.StartSpanOptions(...)

Yes, in hindsight I could have uses Transaction.StartSpanOptions(), but the function apm.StartSpanOptions() completely ignores the opts parameter, which is a problem.

axw:
> I am using go-pg and wrote a callback to trace postgres calls. I have to set the StartTime retroactively because there is only a "finished" callback so I used apm.StartSpanOptions(...)

Understood.

> Yes, in hindsight I could have uses Transaction.StartSpanOptions(), but the function apm.StartSpanOptions() completely ignores the opts parameter, which is a problem.

Sorry, I misread the issue initially, and I was looking at apm.StartSpan and not apm.StartSpanOptions. I understand now.

There's a couple of issues with the patch - I'll respond with review comments.

gocontext.go
```diff
@@ -47,7 +47,10 @@ func StartSpan(ctx context.Context, name, spanType string) (*Span, context.Conte
 // when the span completes.
 func StartSpanOptions(ctx context.Context, name, spanType string, opts SpanOptions) (*Span, context.Context) {
 	tx := TransactionFromContext(ctx)
-	span := tx.StartSpan(name, spanType, SpanFromContext(ctx))
+	if opts.Parent == (TraceContext{}) {
```
axw: I'm rather embarrassed about this bug, thanks for picking it up!

There's a couple of issues here. As it is, this code will panic if ctx does not contain a transaction. This can be observed in the test failures. Second, StartSpanOptions is supposed to use the parent span from ctx, if it contains one. It would have done that before, but with this change it won't.

I think how it should be is this:
 - If opts.Parent is non-zero, leave opts alone
 - Otherwise, set opts.Parent to the trace context of the span in ctx, if any

This should work:

```go
if opts.Parent == (TraceContext{}) {
    opts.Parent = SpanFromContext(ctx).TraceContext()
}
```

(*Span.TraceContext() will return a zero trace context value if the span is nil.)

The call to tx.StartSpanOptions will already set the parent to tx.traceContext if it's zero.

ChristophPech: I added a test, but I had to expose Spandata.Timestamp for it, I'm not sure if this is the right way.
I also changed apm.StartSpanOptions to your suggestions, but that was not enough because Transaction.StartSpan is not equivalent to Transaction.StartSpanOptions, which I wouldn't know how to fix without breaking the API so added the ended => dropped logic to apm.StartSpanOptions which isn't really great either.

axw: > I added a test, but I had to expose Spandata.Timestamp for it, I'm not sure if this is the right way.

The timestamp is meant to be immutable, and exposing it would open up the API to misuse.

The right way to test this would be to get the resulting span event from the recorder transport, and check its timestamp matches what you provided. Take a look at TestTracerStartSpan for an example.

> I also changed apm.StartSpanOptions to your suggestions, but that was not enough because Transaction.StartSpan is not equivalent to Transaction.StartSpanOptions, which I wouldn't know how to fix without breaking the API so added the ended => dropped logic to apm.StartSpanOptions which isn't really great either.

Good point. It's not great repeating the code, but I can find a way to fix it later.

axw: I started writing a test and then got carried away and made the other changes too: https://github.com/ChristophPech/apm-agent-go/compare/master...axw:startspanoptions-options

ChristophPech: This looks great, I'll test it immediately.

axw:
@ChristophPech did you have a chance to test this out yet?

Let me know if you'd like me to take it over. There's no rush though, I'm happy to wait if you're keen to finish it off.

ChristophPech:
+axw Yes, I am running your changes now on my staging server for several days and it works as expexted. I used go mod to test it and have now merged it.

axw:
Thank you very much @ChristophPech! You're the first contributor outside of @elastic :1st_place_medal:

PR #333, opened 2018-11-28
module/apmecho: Set *apm.Transaction on echo.Context

KimMachineGun:
Now, we can't get `*apm.Transaction` in the handler.
It is hard to track some behavior in the handler.

So, I create some code to set `*apm.Transaction` on `echo.Context`.
I think that it makes me track some behavior in the handler easily.

### DIFF
--- a/module/apmecho/middleware.go
+++ b/module/apmecho/middleware.go
@@ -11,6 +11,9 @@ import (
 	"go.elastic.co/apm/module/apmhttp"
 )
 
+// TxContextKey is used as a key to get *apm.Transaction from echo.Context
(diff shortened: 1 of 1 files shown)


axw:
@KimMachineGun did you mean to close this?

The transaction is in the request context, which you can get with `c.Request().Context()`, where `c` is an `echo.Context`. I'll create an issue to add a code example for adding spans when using echo.

KimMachineGun:
@axw
Sorry, I saw that a minute ago. So I closed it. 😂

axw:
No problem! Thanks for confirming :)

PR #342, opened 2018-11-30
Add Kubernetes metadata

### LINKED ISSUE #329: Add k8s pod metadata
See https://github.com/elastic/apm/issues/21.

axw:
Add Kubernetes metadata. We sniff pod metadata from `/proc/self/cgroup`, and then add or override any values specified through the environment variables `KUBERNETES_NAMESPACE`, `KUBERNETES_NODE_NAME`, `KUBERNETES_POD_NAME`, and `KUBERNETES_POD_UID`. These can be populated using the Kubernetes Downward API.

Through `/proc/self/cgroup` we are able to sniff the container ID and pod UID. We also assume that the hostname is the pod name, which is the default. If the hostname has been explicitly set, then the user can use the Downward API as mentioned above to set the correct pod name.

Closes elastic/apm-agent-go#329


axw:
While testing, I realised that the container.id wasn't being populated. That's because the cgroup paths are different under k8s. They look like this:

`1:name=systemd:/kubepods/besteffort/pode9b90526-f47d-11e8-b2a5-080027b9f4fb/d87d73916efef57277e02f38eea58495a5e3687a494e34ffa8e14fc2b8a52dc3`

The last segment is the Docker container ID. Obviously only holds if Docker is the container runtime being used. The second-to-last segment is the pod UID, prefixed with "pod". I also realised we can get the pod _name_ just from the hostname.

Need to run some more tests, with the following variables:
 - kubelet cgroup driver: cgroupfs vs. systemd
 - alternative container runtime?

PR #358, opened 2018-12-05
[APM-CI][JENKINS-41929] use params for env variables default values

kuisathaverat:
* use params also in when conditions
* grab default value params


kuisathaverat:
replaced by https://github.com/elastic/apm-agent-go/pull/359

PR #398, opened 2018-12-24
Feature/391 implement error interface

### LINKED ISSUE #391: apm.Error implements error interface
At the moment there is no possibility to pass Error as error interface. 
We have to redeclare type as `type ApmError apm.Error` and define methods there.

`func (e *Error) Error() string` should be implemented

Is there any reason, why Error does not implement error interface? 
I can prepare pull request for that feature if you're interested in.

psrebniak:
Fixes #391 
Had to change `apm.captureError` behavior a little bit - instead of returning nil, `Error` with nil `ErrorData` is returned. Comments are adjusted. 

I think `Error.sent()` name may be adjusted as there are more cases when `ErrorData` is nil. 

Tests for `causer` and `error` interfaces added.

### DIFF
--- a/error.go
+++ b/error.go
@@ -62,7 +62,7 @@ func (t *Tracer) NewError(err error) *Error {
 	if err == nil {
 		panic("NewError must be called with a non-nil error")
 	}
-	e := t.newError()
+	e := t.newError(err)
 	rand.Read(e.ID[:]) // ignore error, can't do anything about it
 	initException(&e.exception, err)
 	initStacktrace(e, err)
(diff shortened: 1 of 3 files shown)


elasticcla:
Hi @psrebniak, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/apm-agent-go/pull/398.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PR #399, opened 2018-12-24
#391; implement error interface

psrebniak:
Fixes elastic/apm-agent-go#391
Had to change apm.captureError behavior a little bit - instead of returning nil, Error with nil ErrorData is returned. Comments are adjusted.

I think Error.sent() name may be adjusted as there are more cases when ErrorData is nil.

Tests for causer and error interfaces added.

### DIFF
--- a/error.go
+++ b/error.go
@@ -89,6 +89,8 @@ func (t *Tracer) NewError(err error) *Error {
 		panic("NewError must be called with a non-nil error")
 	}
 	e := t.newError()
+	e.cause = err
+	e.err = err.Error()
 	rand.Read(e.ID[:]) // ignore error, can't do anything about it
 	initException(&e.exception, err)
 	initStacktrace(e, err)
@@ -116,6 +118,8 @@ func (t *Tracer) NewErrorLog(r ErrorLogRecord) *Error {
 	if e.log.Message == "" {
 		e.log.Message = "[EMPTY]"
 	}
+	e.cause = r.Error
+	e.err = e.log.Message
 	rand.Read(e.ID[:]) // ignore error, can't do anything about it
 	if r.Error != nil {
 		initException(&e.exception, r.Error)
@@ -150,6 +154,14 @@ type Error struct {
 	// ErrorData holds the error data. This field is set to nil when
 	// the error's Send method is called.
 	*ErrorData
+
+	// cause holds original error.
+	// It is accessible by Cause method
+	// https://godoc.org/github.com/pkg/errors#Cause
+	cause error
+
+	// string holds original error string
+	err string
 }
 
 // ErrorData holds the details for an error, and is embedded inside Error.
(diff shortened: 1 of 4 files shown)


error.go
```diff
@@ -186,6 +194,25 @@ func (e *Error) SetTransaction(tx *Transaction) {
...
+// https://godoc.org/github.com/pkg/errors#Cause
+func (e *Error) Cause() error {
+	if e != nil {
+		return e.cause
+	}
+	return nil
+}
+
+// Error returns string message for error.
+// if Error or Error.cause is nil, "[EMPTY]" will be used.
```
axw: I think for log-type errors, the log message should be returned by `Error()`, while the associated error is returned by `Cause()` (as you do already.)

Perhaps change NewError and NewErrorLog to set another error message field at construction time, alongside the cause?

gocontext.go
```diff
@@ -64,13 +64,15 @@ func StartSpanOptions(ctx context.Context, name, spanType string, opts SpanOptio
 // set either from err, or from the caller.
 //
 // If there is no span or transaction in the context, CaptureError returns
-// nil. As a convenience, if the provided error is nil, then CaptureError
-// will also return nil.
+// Error with nil ErrorData field. As a convenience, if the provided error is nil,
```
axw: Can you please explain the need for this change? It's not clear to me, and I'm reluctant to change this behaviour.

psrebniak: Wanted to avoid situation when returned type is nil pointer to interface - it can be tricky sometimes https://stackoverflow.com/questions/13476349/check-for-nil-and-nil-interface-in-go

Second thing is to keep original error even if error was already sent.

If you like, I can revert that change - was made to provide stable interface, no matter error was sent previously or not.

axw: I think I can see reason in that: we would be saying that if the result is nil if and only if the input error is nil. If the input error is non-nil, but there is no span or transaction in the context, then we'll get a non-nil Error that wraps the input error, only its (Send, SetTransaction, SetSpan) methods will be no-ops.

I'm fine with keeping this, but I'd like a change to the doc comment: it should be stated in terms of functionality, rather than whether the ErrorData field is nil or not. Maybe something like this?

> If the provided error is nil, then CaptureError will also return nil; otherwise a non-nil Error will always be returned. If there is no transaction or span in the context, then the returned Error's Send method will have no effect.

error.go
psrebniak: Hmm, you mean I should add another `error` field to both `ErrorLogRecord` and `ErrorData` structures?
Not sure when should I use them. Could you explain?

axw: I mean add an error message field to `Error`, so it looks something like:

```go
type Error struct {
    *ErrorData
    cause error
    err string
}

func (e *Error) Error() string {
    return e.err
}

func (e *Error) Cause() error {
    return e.cause
}
```

`NewErrorLog` would set the ErrorLogRecord.Message value, and `NewError` would just copy across the value from the provided error.

What do you think?

axw:
@psrebniak do you expect to finish this off soon? I'd love to get it in for the next minor release, which I'm hoping to release in the next week or two.

If you're unsure about the changes I requested, we can just land it as it is (after rebasing of course) and I can look into making minor adjustments later.

psrebniak:
@axw I'll finish changes on Monday, sorry for delay

axw:
> @axw I'll finish changes on Monday, sorry for delay

No worries, thank you!

error.go
psrebniak: Good idea - I just wonder about scenario when `NewErrorLog` is called with nil error. Do you think it is possible on production? 
This scenario exists in `validation_test.go:236 TestValidateErrorLog` and not sure should I change test to have non-nil error value or not.
At the moment I bypassed problem by settings `Error.err` manually afterwards but I really don't like this solution.

psrebniak: Thank you for detailed explanation - I revert changes made to newError function and initialize fields later.

axw:
@psrebniak thank you, looks good. I think this is ready to merge, with a few minor things:

 - run "make fmt"
 - add a brief description of the change under "Unreleased" in CHANGELOG.md
 - rebase on master, squash down to one commit

psrebniak:
@axw done - thanks for your help and patience :)

axw:
@psrebniak not at all, thank you very much for your contribution.

PR #403, opened 2019-01-03
fix apmhttp code example when wrapping a function

jonathanbeber:
`apmhttp.Wrap` function receive a `http.Handler` type and in this example was receiving a function. Changing the example to use the `http.HandlerFunc` function which will return a `http.Handler` type.


axw:
Thanks!

Issue #384, opened 2018-12-07
CI: Jenkins isn't uploading coverage for merge commits to codecov


axw:
The builds keep showing misleading coverage changes. Looking at https://codecov.io/gh/elastic/apm-agent-go/branch/master/commits, the merge commits have lower coverage than the non-merge commits. The most obvious difference between them is that there are no coverage reports from Jenkins on the merge commits.

kuisathaverat:
I am not sure why, we report the same info on PRs and on merges, but the measure seems about 3% different, I will keep an eye on it.

axw:
> The most obvious difference between them is that there are no coverage reports from Jenkins on the merge commits. e.g. see the difference between these:

- https://codecov.io/gh/elastic/apm-agent-go/commit/ccc9f2b184d02238351c8ce4df7fa9153732dce3/build
- https://codecov.io/gh/elastic/apm-agent-go/commit/db8b0c624960218f76fdc823a293d3bb304178b3/build

kuisathaverat:
Jenkins PR build
https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-go%2Fapm-agent-go-mbp/detail/PR-387/2/pipeline

Jenkins merge Master build 
https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-go%2Fapm-agent-go-mbp/detail/master/25/pipeline/73

Jenkins report
https://codecov.io/gh/elastic/apm-agent-go/commit/f61a7f13aa02b1cba4e823807d9d53cbb3a764c7/build

Travis Report
https://codecov.io/gh/elastic/apm-agent-go/commit/c4a71d917a09679badfa025e9bb26d212f120d39/build

I saw on the Jenkins logs that Jenkins report to the same commit for the PR and the merge on master (https://codecov.io/gh/elastic/apm-agent-go/commit/f61a7f13aa02b1cba4e823807d9d53cbb3a764c7/build) I dunno why Travis change the report commit, the commit SHA1 is `f61a7f13aa02b1cba4e823807d9d53cbb3a764c7` Tavis points to `c4a71d917a09679badfa025e9bb26d212f120d39`

https://github.com/elastic/apm-agent-go/commit/f61a7f13aa02b1cba4e823807d9d53cbb3a764c7

kuisathaverat:
[...]

PR #419, opened 2019-01-09
CI: coverage improvements

axw:
- Update scripts/test_coverage.sh to remove repeat "mode: atomic" lines,
  and drop the code from scripts/jenkins/docker-test.sh which did that.
- Under Jenkins, write the "go test -cover" output to build/codecov
  instead of build/coverage, so that we can prevent the codecov script
  from picking up the Cobertura report.

Hopefully addresses #384


axw:
Did not work, the codecov script looks in the git repo root regardless of CWD. Will need to make some changes to the shared Jenkins pipeline repo to enable specifying the report file.

PR #442, opened 2019-02-06
[APM-CI] trigger by a comment, set fixed lib version

kuisathaverat:
* trigger by a comment
* set fixed lib version

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,7 +18,7 @@ pipeline {
     durabilityHint('PERFORMANCE_OPTIMIZED')
   }
   triggers {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -15,6 +17,9 @@ pipeline {
     disableResume()
     durabilityHint('PERFORMANCE_OPTIMIZED')
   }
+  triggers {
+    issueCommentTrigger('.*(?:jenkins\\W+)?run\\W+(?:the\\W+)?tav\\W+tests(?:\\W+please)?.*')
```
axw: "tav" is not a thing for the Go agent, I guess this was copied from the Node.js agent? So just remove the tav bit, and then LGTM.

kuisathaverat: yep C&P

axw: Thanks, merge at will

Issue #449, opened 2019-02-19
apmelasticsearch: capture additional request bodies


axw:
See https://github.com/elastic/apm/issues/55

PR #470, opened 2019-03-12
 module/apmelasticsearch: capture more request bodies

axw:
Capture the request body for more search requests.
Previously we would capture the body only for requests
where the path ended with "/_search". We will now
match the following suffixes:

 - "/_search"
 - "/_msearch"
 - "/_search/template"
 - "/_msearch/template"
 - "/_rollup_search"


axw:
Closes #449

PR #477, opened 2019-03-16
Updates echo package version to v4.

1995parham:
Echo has released its new version with Go module support. Its import path has been changed to `github.com/labstack/echo/v4` and I change it in the `apmecho`.


axw:
@1995parham thanks for opening a PR for this!

This breaks backwards compatibility, so it would require a major version bump of the agent. However, I think we should maintain support for both versions for some time, as not everyone is using modules yet.

I'm not sure what the best thing to do is. Perhaps introduce `apmecho/v4`, which is a copy of `apmecho` but with the import paths changed?

1995parham:
I agree with your idea, we can create `apmecho/v4` that works with version 4 of echo and save `apmecho` for backward compatibility. I try to do it in this PR and let you know.

1995parham:
@axw I tried to provide a separate module for version 4 of Echo.

module/apmecho/go.mod
```diff
@@ -1,17 +1,13 @@
-module go.elastic.co/apm/module/apmecho
```
axw: The changes to the existing go.mod and go.sum can be reverted now, right?

module/apmecho/v4/example_test.go
```diff
@@ -0,0 +1,39 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package apmecho_test
+
+import (
+	"github.com/labstack/echo/v4"
```
axw: Because of how echo adds the version in the module definition, rather than as a directory, this package won't build with Go 1.8 (see https://travis-ci.org/elastic/apm-agent-go/jobs/507736180).

Can you please add a build constraint for 1.9+ to each of the Go files in this new package? Look for "// +build go1.9" in the files in apmgocql for an example, if you're unfamiliar with doing that.

module/apmecho/go.mod
1995parham: Yes, I forgot that.

module/apmecho/v4/example_test.go
1995parham: That is an awesome idea 👍 I added the build tags in `v4` files.

module/apmecho/go.mod
axw: go.mod and go.sum are still modified, can you please remove the changes from the diff altogether? You can do this by resetting those files to their contents in the master branch:

```
git reset master go.mod go.sum
git checkout go.mod go.sum
git commit -a -m "Revert changes to go.mod/go.sum"
```

axw:
Thanks @1995parham! There's something going wrong with CI to do with module verification. I think it's a bug in the verification code, and not your changes. I will try to look into it tomorrow.

1995parham:
@axw Thanks for your time and your reviews. I am available if any changes will need.

axw:
@1995parham the go tool is getting confused by the "v4" path, thinking it signifies a major version path for the apmecho module.

The best idea I've come up with is to rename it to "go.elastic.co/apm/module/apmechov4", but I'm open to other ideas if you have any.

1995parham:
@axw Personally I prefer `/v4` but because of issues I think creates another top-level module is the only way. I will try it and let you know. Again thanks for your time.

module/apmechov4/doc.go
```diff
@@ -0,0 +1,22 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+// +build go1.9
+
+// Package apmecho provides middleware for the Echo framework,
```
axw: This should now be "Package apmechov4 ..."
(golint is complaining)

scripts/Dockerfile-testing
```diff
@@ -6,6 +6,7 @@ COPY go.mod go.sum /go/src/go.elastic.co/apm/
 COPY internal/tracecontexttest/go.mod internal/tracecontexttest/go.sum /go/src/go.elastic.co/apm/internal/tracecontexttest/
 COPY module/apmbeego/go.mod module/apmbeego/go.sum /go/src/go.elastic.co/apm/module/apmbeego/
 COPY module/apmecho/go.mod module/apmecho/go.sum /go/src/go.elastic.co/apm/module/apmecho/
+COPY module/apmecho/v4/go.mod module/apmecho/v4/go.sum /go/src/go.elastic.co/apm/module/apmecho/v4/
```
axw: This (and below) needs to be changed from `apmecho/v4` to `apmechov4`

axw:
Thanks again for your contribution @1995parham!

Issue #486, opened 2019-03-22
Ensure consistent transaction name for unknown-route HTTP requests


axw:
We should ensure consistency with other agents, and set the transaction name to "\<METHOD\> unknown route" for incoming HTTP requests, when the route cannot be determined.

axw:
This is a little bit complicated for gorilla/mux, since its middleware isn't invoked at all when a route cannot be matched. To support this we'll need to override the Router.NotFoundHandler and Router.MethodNotAllowedHandler fields (see https://godoc.org/github.com/gorilla/mux#Router).

axw:
Likewise for httprouter.

PR #495, opened 2019-04-02
Add apmchi module to support github.com/go-chi/chi

### LINKED ISSUE #494: Support github.com/go-chi/chi
Out of the box, `apmhttp` supports https://github.com/go-chi/chi, but does not recognize parameterized routes.

There should exist a module `apmchi` that leverages middleware to report each request as a transaction, correctly specifying the server request name. This could meet essentially the same public interface that `apmgorilla` is using, and be used as such:
```go
r := chi.Router()
r.Use(apmchi.Middleware())
```

People could manually configure `apmhttp` along with `apmhttp.WithServerRequestName`, then wrap the router, which is what an `apmchi` module would do automatically.

**Additional context**
I'd be happy to build this out, shouldn't take too long.

rliebz:
Create an `apmchi` module in `modules/`, which allows users to create middleware to log to Elastic APM. Does the standard `apmhttp` stuff, but also handles parametrized paths.

It can be used as follows:
```go
r := chi.NewRouter()
r.Use(apmchi.Middleware())
```

Resolves elastic/apm-agent-go#494


module/apmchi/middleware.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	routePath := r.URL.Path
+	if r.URL.RawPath != "" {
+		routePath = r.URL.RawPath
+	}
+
+	tctx := chi.NewRouteContext()
+	if !rctx.Routes.Match(tctx, r.Method, routePath) {
+		return routePath
```
axw: To match the behaviour of other agents, I intend to change the Go agent to set the transaction name to `"<METHOD> unknown route"` when the route cannot be determined (https://github.com/elastic/apm-agent-go/issues/486), such as in this case. Would you mind making that change here?

module/apmchi/middleware.go
```diff
@@ -0,0 +1,104 @@
...
+	return r.Method + " " + getRequestPattern(r)
+}
+
+func getRequestPattern(r *http.Request) string {
+	rctx := chi.RouteContext(r.Context())
+	if pattern := rctx.RoutePattern(); pattern != "" {
+		return pattern
+	}
+
+	routePath := r.URL.Path
```
axw: I'm unfamiliar with Chi, but it appears that if you use `Router.Use`, then the middleware will be run before routing; and if you use `Router.With`, then middleware will run after. Are there any other circumstances in which the route pattern not set above? It doesn't seem ideal to run the route pattern matching twice.

According to https://godoc.org/github.com/go-chi/chi#Context.RoutePattern, the route pattern should be queried after executing the handler. Should the transaction name be set initially to `""` (or perhaps `"\<METHOD\> unknown route"`...), and then updated before ending the transaction?

Another thing to consider is that we may later want to add sampling based on transaction names. The earlier we know the transaction name, the better; that may mean restricting (or recommending) people to add the middleware inline, using `Router.With`.

axw:
jenkins run the tests please

module/apmchi/middleware.go
rliebz: > it appears that if you use Router.Use, then the middleware will be run before routing; and if you use Router.With, then middleware will run after. 

I believe that's true with some caveats. Because of how subroutes are set up, using either `Router.Use` and `Router.With` will have routed to a particular point, but not necessarily the whole thing:

```go
middleware := // ...
r := chi.NewRouter()
r.Use(middleware) // RoutePattern(): ""

r.With(middleware). // RoutePattern(): "/prefix/*"
  Route("/prefix", func(r chi.Router) {
    r.Use(middleware) // RoutePattern(): "/prefix/*"
    r.With(middleware).Get("/{thing}", handler)  // RoutePattern(): "/prefix/{thing}"
  })
```

What should be happening here is:
- If you do `Use` at the top level, there's no route metadata, so it'll find that with `Match`. 
- If you do `With` or do `Use` further down into the routing, it will capture the part of the route pattern so far, but not attempt to find the rest. So you could use it on each route to get the full thing, or on a subrouter if all you care about is the prefix.

I think that's a relatively sane behavior, but I'm not sure if it's desirable/consistent. Other options I can think of are to always match on the full pattern, or start with `unknown route` and swap in the actual route after, like you mentioned.

Either way, let me know, and I'll get some more proper test cases for these situations and update the documentation.

> It doesn't seem ideal to run the route pattern matching twice.

Agreed, but that's [the recommended pattern](https://github.com/go-chi/chi/issues/270#issuecomment-335570862) as far as I can tell, and it's [used in the official middleware](https://github.com/go-chi/chi/blob/master/middleware/get_head.go#L29) for similar purposes.

> Should the transaction name be set initially to "" (or perhaps "\<METHOD\> unknown route"...), and then updated before ending the transaction?

I'm not sure exactly what the trade-offs here are from an APM perspective, but if this is the preferred behavior, it should be as easy as re-implementing a slightly modified version of the `handler.ServeHTTP` function from `apmhttp`. Of course, that would mean not knowing the transaction name until the transaction is basically over.

So I guess there's options—let me know what you think is the way forward.

axw: > Agreed, but that's the recommended pattern as far as I can tell, and it's used in the official middleware for similar purposes.

OK, fair enough. Thanks for the detailed reply.

The difference in behaviour between `Use` and `With` seems a little subtle to me; generally transactions are named after the full matched route. I would just go with always matching the full route pattern, but if you have a use case for naming transactions by a route prefix, then I'm fine with keeping that flexibility.

> Either way, let me know, and I'll get some more proper test cases for these situations and update the documentation.

Thank you, that would be excellent.

rliebz: > I would just go with always matching the full route pattern

Yeah I think this is the behavior that makes the most sense. Done.

module/apmchi/middleware.go
rliebz: Done and tested.

rliebz:
Not sure what makes Jenkins fail whenever I push, but otherwise I think it's good to go.

axw:
> Not sure what makes Jenkins fail whenever I push, but otherwise I think it's good to go.

Our Jenkins CI only runs when instructed to by @elastic folks, preventing damage by malicious PRs. I'll get it going.

axw:
jenkins run the tests please

Issue #489, opened 2019-03-31
Feature: github.com/go-redis/redis tracing (already implemented)


aspacca:
**Is your feature request related to a problem? Please describe.**
I'm evaluating adoption of Elastic APM I a project and the only component I use missing auto instrumentations was github.com/go-redis/redis

**Describe the solution you'd like**
I already implemented tracing for `redis.Client` and `redis.ClusterClient` on github.com/aspacca/apmgoredis. I'd like to have a review and some guidance in order to add tests (I didn't spot tests for go agent modules, at a first glance).
The same mechanism could be use for `redis.Ring` but it misses an interface so I cannot use the composition solution. I will discuss with the library's developers how to solve.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

axw:
@aspacca thank you for offering this. The code looks good. A couple of thoughts:

- The contextClient.ctx field isn't used, since it's captured in the process/processPipeline closures. You could perhaps make those methods instead, or otherwise just remove the field.

- I'm not convinced errors should be captured, since they may be expected by the application. Perhaps this could be configurable? In the [apmredigo](https://github.com/elastic/apm-agent-go/tree/master/module/apmredigo) module, errors are not captured, it is left to the application.

Regarding tests: for apmredigo, there are unit tests in https://github.com/elastic/apm-agent-go/blob/master/module/apmredigo/conn_test.go, and integration tests in https://github.com/elastic/apm-agent-go/blob/master/module/apmredigo/integration_test.go.

The integration tests run inside Docker Compose, with Redis already set up. If you can replicate the integration tests for apmredigo, then you can test it with `make docker-test` in the top level directory. Alternatively, you could start Redis manually and set the environment variable "REDIS_URL" to point at it.

For the unit tests, you can again copy the unit tests from apmredigo, but instead of mocking the connection, I think you can just call WrapProcess on the original client to avoid connecting to Redis. You probably know better here; I haven't used this particular client.

> The same mechanism could be use for redis.Ring but it misses an interface so I cannot use the composition solution. I will discuss with the library's developers how to solve.

It also appears that the two functions WrapClient and WrapClusterClient are only needed because redis.UniversalClient doesn't have the WrapProcessPipeline method. If you're proposing changes to go-redis, perhaps suggest adding that method to UniversalClient? Then you could get away with just one function, `func WrapUniversalClient(redis.UniversalClient) Client`.

aspacca:
[...]

PR #505, opened 2019-04-13
Auto instrumentation for go-redis

aspacca:
addresses #489


scripts/moduledirs.sh
```diff
@@ -6,5 +6,5 @@
 if test -z "$(go env GOMOD)"; then
     pwd
 else
-    find -type f -name go.mod \! -path './vendor/*' -printf "%h\n"
+    find . -type f -name go.mod \! -path './vendor/*' -exec dirname '{}' \;
```
aspacca: this is just a fix to be not-GNU compatible

module/apmgoredis/integration_test.go
```diff
@@ -0,0 +1,391 @@
...
+}
+
+func redisClient(t *testing.T) *redis.Client {
+	redisURL := os.Getenv("GOREDIS_URL")
+	if redisURL == "" {
+		t.Skipf("GOREDIS_URL not specified")
+	}
+
+	var err error
+	redisURL, err = lookupIPFromHostPort(redisURL)
```
axw: could you use https://godoc.org/github.com/go-redis/redis#ParseURL here?

module/apmgoredis/integration_test.go
```diff
@@ -0,0 +1,391 @@
...
+	redisURLs := strings.Split(os.Getenv("GOREDIS_CLUSTER_URLS"), " ")
+	if len(redisURLs) == 0 {
+		if t != nil {
+			t.Skipf("GOREDIS_CLUSTER_URLS not specified")
+		}
+	}
+
+	for i, redisURL := range redisURLs {
+		var err error
+		redisURLs[i], err = lookupIPFromHostPort(redisURL)
```
axw: and here, just pulling the Addrs field out of the options?

module/apmgoredis/integration_test.go
```diff
@@ -0,0 +1,391 @@
...
+		return nil
+	}
+
+	closeConn := true
+	client := redis.NewClient(&redis.Options{
+		Addr: redisURL,
+	})
+
+	defer func() {
+		if closeConn {
```
axw: Since closeConn is set to false immediately after the defer, this will never be reached. May as well lose the closeConn and defer.

module/apmgoredis/integration_test.go
```diff
@@ -0,0 +1,391 @@
...
+		}
+	}
+
+	closeConn := true
+	client := redis.NewClusterClient(&redis.ClusterOptions{
+		Addrs: redisURLs,
+	})
+
+	defer func() {
+		if closeConn {
```
axw: same as in redisClient: lose the closeConn and defer, as this will never be reached.

module/apmgoredis/client.go
```diff
@@ -0,0 +1,129 @@
...
+	"go.elastic.co/apm"
+)
+
+// Client is the interface returned by ContextClient.
+//
+// Client implements redis.UniversalClient
+type Client interface {
+	redis.UniversalClient
+
+	// ClusterClient returns the eventually embedded redis.ClusterClient or nil
```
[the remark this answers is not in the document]
aspacca: actually passing `Wrap` receives a no `redis.UniversalClient` it will be caught at compile time. I will refactor to never return `nil`

module/apmgoredis/client.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// WithContext returns a shallow copy of the client with
+	// its context changed to ctx and will add instrumentation
+	// with client.WrapProcess and client.WrapProcessPipeline
+	//
+	// To report commands as spans, ctx must contain a transaction or span.
+	WithContext(ctx context.Context) Client
+}
+
+// Wrap according to client type. The context can be changed using Client.WithContext.
```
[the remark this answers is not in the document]
aspacca: it is still valid to call `WrapProcess/WrapProcessPipeline` and it should not interfere with the instrumentation. The only thing is if I `Wrap` the client two times with the same context, in this case the content of the spans will be duobled

module/apmgoredis/client.go
```diff
@@ -0,0 +1,129 @@
...
+			for i, cmd := range cmds {
+				cmdName := strings.ToUpper(cmd.Name())
+				if cmdName == "" {
+					cmdName = "(flush pipeline)"
+				}
+
+				cmdNames[i] = cmdName
+			}
+
+			spanName := "(pipeline) " + strings.Join(cmdNames, " ")
```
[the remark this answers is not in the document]
aspacca: the second solution I like more but I'm not sure I see how to achieve an encompassing span

module/apmgoredis/client.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+}
+
+func processPipeline(ctx context.Context) func(oldProcess func(cmds []redis.Cmder) error) func(cmds []redis.Cmder) error {
+	return func(oldProcess func(cmds []redis.Cmder) error) func(cmds []redis.Cmder) error {
+		return func(cmds []redis.Cmder) error {
+			cmdNames := make([]string, len(cmds))
+			for i, cmd := range cmds {
+				cmdName := strings.ToUpper(cmd.Name())
+				if cmdName == "" {
```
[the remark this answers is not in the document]
aspacca: yes, I will change to `(empty command)`

module/apmgoredis/client_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+		_, spans, _ := apmtest.WithTransaction(func(ctx context.Context) {
+			client := apmgoredis.Wrap(client).WithContext(ctx)
+			pipe := client.TxPipeline()
+			pipe.Do("")
+			pipe.Do("")
+			pipe.Exec()
+		})
+
+		switch testCase.isTxWrapped {
```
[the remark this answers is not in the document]
aspacca: I added in that PR that was merged but `go.mod` still pin tagged version (while my fix in only in master): tests should not fail. I will check

module/apmgoredis/integration_test.go
aspacca: taken from https://github.com/elastic/apm-agent-go/blob/master/module/apmredigo/integration_test.go#L122 without realizing that there's no `require` in my code and so the logic is not needed

module/apmgoredis/client.go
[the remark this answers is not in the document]
axw: Something like this:

```go
func processPipeline(ctx context.Context) func(oldProcess func(cmds []redis.Cmder) error) func(cmds []redis.Cmder) error {
	return func(oldProcess func(cmds []redis.Cmder) error) func(cmds []redis.Cmder) error {
		return func(cmds []redis.Cmder) error {
                        pipelineSpan, ctx := apm.StartSpan(ctx, "(pipeline)", "db.redis")
                        defer pipelineSpan.End()

			for i, cmd := range cmds {
				cmdName := strings.ToUpper(cmd.Name())
 				span, _ := apm.StartSpan(ctx, cmdName)
                                defer span.End()
			}
 			return oldProcess(cmds)
		}
	}
}
```

Note that the statement `pipelineSpan, ctx := apm.StartSpan` shadows "ctx" with a new context containing the pipeline span. When that is passed into the calls to apm.StartSpan below, those spans end up as children of the pipeline span.

module/apmgoredis/client_test.go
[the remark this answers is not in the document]
aspacca: tests actually don't use go modules, but go get, that fetches the latest master. I updated anyway `go.mod` to use the commit with my fix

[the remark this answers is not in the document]
axw: :+1:
Looks like there's a Go 1.8 test failure unrelated to your changes, I'm looking into this now.

axw:
@aspacca I've just merged a change into master which fixes the test failure in https://travis-ci.org/elastic/apm-agent-go/jobs/520141217. You'll need to rebase or merge master to get that.

There are some other CI failures due to formatting, e.g. https://travis-ci.org/elastic/apm-agent-go/jobs/520141216. You can fix that with `make fmt`.

Finally, you will need to sign the [CLA](https://www.elastic.co/contributor-agreement) before I can merge this.

aspacca:
@axw everything should be fine now

axw:
Thanks @aspacca. I'm away on vacation at the moment, I will finish reviewing next week.

module/apmgoredis/client.go
```diff
@@ -0,0 +1,163 @@
...
+	// its context changed to ctx and will add instrumentation
+	// with client.WrapProcess and client.WrapProcessPipeline
+	//
+	// To report commands as spans, ctx must contain a transaction or span.
+	WithContext(ctx context.Context) Client
+}
+
+// Wrap wraps client such that executed commands are reported as spans to Elastic APM,
+// using the client's associated context.
+// A context-specific client may be obtained by using Client.WithContext.
```
axw: It appears from the tests that not everything works with Ring. Can you please add a comment here explaining what isn't supported, linking to issues if there are any?

aspacca: `Ring` has not `TxPipeline/TxPipelined` implemented, it's not related to the instrumentation with the apm agent, do you want me to add a comment for this?

While, related to instrumentation, it works with `Ring` on `Pipeline/Pipelined` but not for direct command on the client: this PR https://github.com/go-redis/redis/pull/1017 will address

For me it's ok to wait for that PR to be merged, and then remove the special case in the test here

axw: I see, thanks. I'd prefer to wait for go-redis/redis#1017 to merge first, to avoid a little bit of churn.

aspacca:
go-redis/redis#1017 merged, pushed final commit here

axw:
I ran the docker tests locally, and I'm seeing this error occur in the integration tests:

    Received unexpected error:      
    CLUSTERDOWN Hash slot not served

Any ideas what's going on there?

aspacca:
redis cluster was not setup properly: all the cluster nodes must be up before we start the cluster. with docker compose 2.2 you can set healthcheck to be sure that containers dependencies are satisfied upon readiness and not only bootstrap. but you cannot do anymore in version 3 (that’s used by the apm-agent-go). I tried to overcome the issue setting a restart “on_failure” for the redis cluster container but you may have hit a corner case where this is not enough 
if the cluster is not setup properly in general go-redis cluster client will fails

axw:
> redis cluster was not setup properly

This is the case, but unrelated to timing. Logs from redis-cluster:

    redis-cluster_1          | Unrecognized option or bad number of args for: '--cluster'

Looks like an issue with the entrypoint commant. I'll have to wait until I have some time to debug this before merging, as I don't want local testing to be blocked.

aspacca:
it’s really strange because it passed jenkins build, unless they pushed a new redis:latest with different behavior in the meanwhile

aspacca:
the case could be also that you don’t have updated redis:latest image :)

axw:
> the case could be also that you don’t have updated redis:latest image :)

I tried that already, it wasn't it. The `--cluster` flag is there, it's something to do with the args being passed into it. Possibly related to the order of services coming up? Really just guessing at the moment.

axw:
Looks like it was me after all, I've run it a bunch of times without any failures.

axw:
Thanks again for your work on this, @aspacca.

PR #523, opened 2019-05-03
tracecontexttest: build service binary

axw:
Build the service first, rather than as the CMD, so building doesn't hold up starting the service.


axw:
Because of the broken state of Go 1.8, I'll need to merge this into #522.

PR #532, opened 2019-05-19
Add row_query callback to apmgorm

Vilsol:
The RowQuery callback is used when objects are queried with functions like row, rows, count, etc are executed.

### DIFF
--- a/module/apmgorm/context.go
+++ b/module/apmgorm/context.go
@@ -87,6 +87,10 @@ func registerCallbacks(db *gorm.DB, dsnInfo apmsql.DSNInfo) {
 			spanType:  execSpanType,
 			processor: func() *gorm.CallbackProcessor { return db.Callback().Update() },
 		},
+		"gorm:row_query": {
+			spanType:  querySpanType,
+			processor: func() *gorm.CallbackProcessor { return db.Callback().RowQuery() },
+		},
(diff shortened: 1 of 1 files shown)


axw:
@Vilsol thanks for the PR!

Would you please add a test?

Also, in case you haven't already noticed, there is a [CLA](https://www.elastic.co/contributor-agreement) which requires signing. Let me know if you're having trouble with that.

Vilsol:
@axw Thanks, I assumed that commits signed by Github's key also count as signed? If not, how can I re-sign the commit with my own key?

axw:
@Vilsol the message "Pull request contains unsigned commits" is perhaps a bit unclear. Your commits are signed and that's fine, but it appears that you have not signed the CLA.

Did you already sign the CLA at https://www.elastic.co/contributor-agreement? Maybe with another email?

axw:
(Ah, I just got what you meant. I'll have to check about using GitHub's key.)

Vilsol:
Ah, I got it now, the signed commits are related to the contributor agreement not GPG signing. Might want to make the error message more descriptive.

axw:
I've picked up the task of adding a test. I'll close this and continue over in #543. Thanks for the fix @Vilsol!

PR #556, opened 2019-06-09
module/apmgopg: unexpose QueryHook, add Instrument

axw:
Unexpose the QueryHook type, and add the Instrument
function. This hides the implementation details so
that we can potentially support go-pg v9.x.x without
introducing a new package.


axw:
@macnibblet Before including apmgopg in a new release of the agent, I'd like to revise the API a little bit.

go-pg master has changed the QueryEvent structure and QueryHook interface, so the current hook implementation won't compile against it. Hiding the implementation would enable us to support both the current API and the new API when released.

Any concerns about this approach?

macnibblet:
Sounds good to me

PR #559, opened 2019-06-11
Introduce ELASTIC_APM_STACK_TRACE_LIMIT

### LINKED ISSUE #526: Add ELASTIC_STACK_TRACE_LIMIT config
Add ELASTIC_STACK_TRACE_LIMIT config, which would control the maximum number of stack frames to capture in a stack trace. This is implemented by the Node.js and Java agents. See https://www.elastic.co/guide/en/apm/agent/java/current/config-stacktrace.html#config-stack-trace-limit

axw:
We introduce a new configurable limit for the
number of stack frames to capture, defaulting
to 50.

stacktrace.AppendStacktrace has been improved
for a limit of N > 10, don't allocate space for
N callers up front; that would be wasteful if
there are mostly fewer. Instead, allocate space
for 10 and work our way up, like we do when there
is no specified limit.

Closes elastic/apm-agent-go#526

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
  - module/apmgrpc: added WithServerRequestIgnorer server option (#531)
  - Introduce `ELASTIC_APM_GLOBAL_LABELS` config (#539)
  - module/apmgorm: register `row_query` callbacks (#532)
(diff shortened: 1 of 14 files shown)


axw:
@bmorelli25 FYI added a new config entry to the docs

PR #563, opened 2019-06-13
feat(jenkins): GH Checks for the PR

v1v:
## Highlights
- Create GH checks with the most important stages in the CI Pipeline.
- This particular change might help to debug when there are errors in the CI and identify them easily within then PR itself.
- Use the notifyBuildStep

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -56,10 +56,12 @@ pipeline {
         */
         stage('build') {
           steps {
-            deleteDir()
-            unstash 'source'
-            dir("${BASE_DIR}"){
-              sh './scripts/jenkins/build.sh'
+            withGithubNotify(context: 'Build') {
+              deleteDir()
+              unstash 'source'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -197,27 +205,19 @@ pipeline {
         }
       }
       steps {
-        deleteDir()
-        unstash 'source'
-        dir("${BASE_DIR}"){
-          buildDocs(docsDir: "docs", archive: true)
+        withGithubNotify(context: 'Documentation', tab: 'artifacts') {
```
axw: Is this step still needed, seeing as there's the separate elasticsearch-ci docs-build check?

v1v: That's actually a good point, it's completely duplicated and might mislead. Even though I initially thought it might be interesting to keep in case the notification from elasticsearch-ci docs-build is disabled for whatever reason. So I'll remove that check from now on

Issue #568, opened 2019-06-26
request body is missing sometimes on APM


ikropotov:
I have wrapped with apmhttp one of our go services, to track requests on kibana. And for some requests body was missing within apm logs. 

I've dived deeper into your plugin and explored what was going on there, and it turned out that on this line https://github.com/elastic/apm-agent-go/blob/master/capturebody.go#L115 the error appeared, complaining that bc.originalBody is already closed, and cannot be read.

Going further down the hole, I've discovered that we actually use that tool as a little proxy, so when request passed down to our tool here https://github.com/elastic/apm-agent-go/blob/master/module/apmhttp/handler.go#L87 we are doing a further request to other endpoints, and when they reply we write answer to `w http.ResponseWriter` to handle it back.

Looking at `w http.ResponseWriter` definition at http lib https://github.com/golang/go/blob/master/src/net/http/server.go#L127, I came across the comment `handlers should read any needed request body data before writing the response`, so it came out that writing to ResponseWriter sometimes marks body as being read, and afterwards apm plugin is trying to read it again...

So I suppose a proper behaviour would be to read original request body beforehand, and attach the copy to be sent to APM - exactly that I've done in my fork, and will submit a PR.

May be you can merge it or suggest some other solution/improvement.

Thanks, Ivan

axw:
@ikropotov thanks for opening the issue and PR!

`Tracer.CaptureHTTPRequestBody` wraps the body so that whatever the underlying handler reads is buffered. So we're relying on the handler to read the body. Is your application code not doing that?

ikropotov:
@axw our handler is reading it and it's not the question about the Teed part, the issue is that here
https://github.com/elastic/apm-agent-go/blob/master/capturebody.go#L114 
it tries to read `bc.originalBody` in addition to teed body, which is already closed by writing to `w http.ResponseWriter`, and returns an error here https://github.com/elastic/apm-agent-go/blob/master/capturebody.go#L117
however if you remove it reading from originalBody, it will be missing in apm request, in case teed part is not being read in handler (that's why I am reading it fully first in the PR)

axw:
@ikropotov ahh I see, because that error we bail out and don't set anything. Got it.

I'll have to think on what the ideal solution is. As mentioned on the PR, I don't think that reading it all up front is the right way, because of the possibility of a streaming body.

axw:
[...]

PR #569, opened 2019-06-26
fix closed body issue

ikropotov:
PR related to the issue elastic/apm-agent-go#568

### DIFF
--- a/capturebody.go
+++ b/capturebody.go
@@ -70,13 +70,19 @@ func (t *Tracer) CaptureHTTPRequestBody(req *http.Request) *BodyCapturer {
 		io.Reader
 		io.Closer
 	}
+	var err error
+	copiedBody := req.Body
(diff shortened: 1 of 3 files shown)


capturebody.go
```diff
@@ -70,13 +70,19 @@ func (t *Tracer) CaptureHTTPRequestBody(req *http.Request) *BodyCapturer {
 		io.Reader
 		io.Closer
 	}
+	var err error
+	copiedBody := req.Body
+	copiedBody, req.Body, err = drainBody(req.Body)
```
axw: If the client streams its body, this will cause the middleware to block until the whole body is received. I don't think that's desirable behaviour. This is the reason why we don't eagerly read the body, but instead wrap the body so that it is tee'd.

axw:
Thank you for sending the PR. Closing as this is superseded by #575.

PR #575, opened 2019-07-03
 Fix body capture to ignore read errors

### LINKED ISSUE #568: request body is missing sometimes on APM

axw:
Ignore request body read errors when capturing the remaining body. Also, limit the amount we read from the body.

Closes elastic/apm-agent-go#568

### DIFF
--- a/capturebody.go
+++ b/capturebody.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 	"net/url"
 
+	"go.elastic.co/apm/internal/apmstrings"
 	"go.elastic.co/apm/model"
 )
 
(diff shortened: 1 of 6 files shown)


axw:
Supersedes #569

PR #582, opened 2019-07-09
Wrong/Missing information about config variable.

emret:
Tags are still added to context for unsampled txn's.
Developer should check for txn.Sampled() to decide add or not to add to context.


axw:
@emret thank you for the PR!

This is actually a bug in the code, the agent should not be doing that. I've opened https://github.com/elastic/apm-agent-go/issues/583 to track fixing it.

PR #591, opened 2019-07-19
Implement central config polling

### LINKED ISSUE #512: Remote config
See https://github.com/elastic/apm/issues/76. We will update the agent to periodically query config from the APM Server.

axw:
We introduce a new package, apmconfig, which
defines a Watcher interface that can be used
for watching for central config changes.

When creating a new Tracer, it will use its
Transport for watching config if it implements
apmconfig.Watcher. The default HTTP transport
implements that interface.

The default behaviour of watching config can
be disabled by setting the new environment
variable ELASTIC_APM_CENTRAL_CONFIG to "false".

For now only the sampler can be updated via
central config. While there is central sampling
config defined, any locally defined Sampler will
be disregarded, whether it be specified via
environment variable or via SetSampler.

Closes elastic/apm-agent-go#512

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
  - Introduce transaction/span breakdown metrics (#564)
  - Optimised HTTP request body capture (#592)
  - Fixed transaction encoding to drop tags (and other context) for non-sampled transactions (#593)
+ - Introduce central config polling (#591)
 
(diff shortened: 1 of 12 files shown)


axw:
This branch has a reference to docs that are not yet current, so we can't merge this yet.

PR #598, opened 2019-07-25
docs: add mentions of central config

axw:
Add mentions of central config:
- Order of precedence
- ELASTIC_APM_CENTRAL_CONFIG config

Splitting this out of https://github.com/elastic/apm-agent-go/pull/591, so we can land that PR and let it soak a bit while we wait for the server docs to be updated.


axw:
Must not be merged until the server docs are updated.

PR #602, opened 2019-07-29
 module/apmgrpc: fix trace propagation vs. sampling

### LINKED ISSUE #601: module/apmgrpc: traces are broken by sampling
**Describe the bug**

As seen in https://discuss.elastic.co/t/only-common-trace-id-for-sampled-spans/192459, sampling is breaking distributed tracing in `apmgrpc`.

**Expected behavior**

The apmgrpc client interceptor should include the transaction's trace context when spans are dropped (e.g. due to sampling or the max spans limit has been reached), like in the apmhttp client instrumentation.

axw:
Fix a bug in the client interceptor which would break
trace propagation when facing dropped spans or non-sampled
transactions. In these cases we should propagate the
transaction's trace context, rather than sending nothing
in the outgoing metadata which would restart tracing.

Fixes #601

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
  - Optimised HTTP request body capture (#592)
  - Fixed transaction encoding to drop tags (and other context) for non-sampled transactions (#593)
  - Introduce central config polling (#591)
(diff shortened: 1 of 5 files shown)


axw:
jenkins run the tests please

axw:
jenkins run the tests please

PR #613, opened 2019-08-16
added RedisClient() to return underlaying redis.Client

wins1908:
Our code is using *redis.Client, we need to add a method to return underlaying redis.client after called WithContext()

### DIFF
--- a/module/apmgoredis/client.go
+++ b/module/apmgoredis/client.go
@@ -34,6 +34,10 @@ import (
 type Client interface {
 	redis.UniversalClient
 
+	// Client() returns the wrapped *redis.Client,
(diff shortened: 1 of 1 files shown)


elasticcla:
Hi @wins1908, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/apm-agent-go/pull/613.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

wins1908:
Hi,
I added my company email
Could u pls check again?

Thanks

axw:
jenkins run the tests please

PR #621, opened 2019-08-28
reducing cyclomatic complexity when parsing opentracing tags.

codingsince1985:


axw:
@codingsince1985 sorry for the delay in responding - thanks for the PR!

I like the simplicity, but this is going to impact performance: particularly the `clone` call, but also the fmt.Sprintf-then-parse for `http.status_code` and `error`. Do you have any ideas of how to simplify the code without sacrificing performance?

codingsince1985:
Thank you @axw. My plan was to get rid of the O(n^2) complexity when parsing tags. It definitely introduces a cost, which may hardly to be justified. Locally it takes ~2,000ns to clone 10 tag entries, which is ~0.5% increase when measuring existing test cases.
Please feel free to close this PR. I may submit another one this weekend, which will follow current way of doing thing, just faster.

axw:
OK. I admit I've never liked this bit of the code, but it was done for performance. We're careful to minimise overhead added to applications, as minor as it may be.

I'll close it - thank you anyway!

PR #622, opened 2019-09-01
stringify only when necessary

codingsince1985:
Benchmark setting span and transaction context.
Before
```
$ go test -bench=.
goos: linux
goarch: amd64
pkg: go.elastic.co/apm/module/apmot
BenchmarkSetSpanContext-4          	  300000	      4201 ns/op
BenchmarkSetTransactionContext-4   	  300000	      4593 ns/op
PASS
ok  	go.elastic.co/apm/module/apmot	2.881s
```
After
```
$ go test -bench=.
goos: linux
goarch: amd64
pkg: go.elastic.co/apm/module/apmot
BenchmarkSetSpanContext-4          	 1000000	      2148 ns/op
BenchmarkSetTransactionContext-4   	  500000	      2297 ns/op
PASS
ok  	go.elastic.co/apm/module/apmot	3.803s
```


module/apmot/span_test.go
```diff
@@ -0,0 +1,58 @@
+package apmot
```
axw: This file is failing the `goimports` check in CI, as the imports aren't grouped. Also, it'll fail the license check, as it's missing a header. You can fix these two things by running `make update-licenses fmt`

module/apmot/span_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"go.elastic.co/apm"
+	"testing"
+)
+
+func BenchmarkSetSpanContext(b *testing.B) {
```
axw: I think these benchmarks are a bit too low level, seeing as it's only possible for `setSpanContext` and `setTransactionContext` to be called via `Span.Finish`/`FinishWithOptions`.

How about changing them to start and finish real OT spans? i.e.

```go
tracer := apmot.New(apmot.WithTracer(apmtest.DiscardTracer))
for i := 0; i < b.N; i++ {
    span := tracer.StartSpanWithOptions("span", opentracing.StartSpanOptions{
        Tags: map[string]interface{}{/* tags go here */},
    })
    span.Finish()
}
```

codingsince1985: Unfortunately, `tracer` in this case is of `opentracing.Tracer` type, which doesn't have `StartSpanWithOptions()`, I may have to go with `StartSpan()`. Will look into it tomorrow.

axw: Sorry, I was conflating the implementation with the interface. You could instead pass an `opentracing.Tags` into `tracer.StartSpan`.

codingsince1985: Benchmark span finish.
Before
```
$ go test -bench=. tracer_test.go
goos: linux
goarch: amd64
pkg: go.elastic.co/apm/module/apmot
BenchmarkSpanSetSpanContext-4   	  200000	     10252 ns/op
BenchmarkSpanSetTxContext-4     	  100000	     11307 ns/op
PASS
ok  	go.elastic.co/apm/module/apmot	3.552s
```
After
```
$ go test -bench=. tracer_test.go
goos: linux
goarch: amd64
pkg: go.elastic.co/apm/module/apmot
BenchmarkSpanSetSpanContext-4   	  200000	      7656 ns/op
BenchmarkSpanSetTxContext-4     	  200000	      8064 ns/op
PASS
ok  	go.elastic.co/apm/module/apmot	3.480s
```

axw: Nice, thanks! Will you delete span_test.go? I think the benchmarks you added to tracer_test.go are sufficient?

codingsince1985: done.

axw:
jenkins run the tests please

PR #635, opened 2019-09-17
Introduce TraceFormatter, add log correlation docs

axw:
Introduce TraceFormatter, a function taking a context and returning a fmt.Formatter that can format trace, transaction, and span IDs.

Also, add documentation around log correlation.


docs/log-correlation.asciidoc
```diff
@@ -0,0 +1,119 @@
...
+There are two main approaches you can take, depending on whether you are using structured
+or unstructured logging.
+
+[float]
+[[log-correlation-manual-structured]]
+==== Manual log correlation (structured)
+
+For correlating structured logs with traces, the following fields should be added to your logs:
+
+ - trace.id
```
bmorelli25: Any reason why these aren't formatted as code? For example, `trace.id` like you do [here](https://github.com/elastic/apm-agent-go/pull/635/files#diff-9b5a72acb7c2dd95c30074560cccc3beR199).

docs/log-correlation.asciidoc
```diff
@@ -0,0 +1,119 @@
...
+
+Given a transaction object, you can obtain its trace ID and transaction ID using
+the <<transaction-tracecontext, apm.Transaction.TraceContext>> method. Similarly,
+given a span object, you can obtain its span ID using <<span-tracecontext, apm.Span.TraceContext>>.
+
+If you use the context APIs to start transactions and spans, then you can obtain
+the context's current transaction using <<apm-transaction-from-context, apm.TransactionFromContext>>,
+and current span using <<apm-span-from-context, apm.SpanFromContext>>. Note that if
+a transaction is not sampled, `apm.TransactionFromContext` will return `nil`.
+Similarly, spans may be dropped by the agent, so `apm.SpanFromContext` may also return nil.
```
bmorelli25:
```suggestion
Similarly, spans may be dropped by the agent, so `apm.SpanFromContext` may also return `nil`.
```

To match formatting from the line above

docs/log-correlation.asciidoc
```diff
@@ -0,0 +1,119 @@
...
+	labels["trace.id"] = traceContext.Trace.String()
+	labels["transaction.id"] = traceContext.Span.String()
+	if span := apm.SpanFromContext(ctx); span != nil {
+		labels["span.id"] = span.TraceContext().Span
+	}
+}
+----
+
+You can follow this article in order to ingest JSON-encoded logs with Filebeat:
+{blog-ref}how-to-instrument-your-go-app-with-the-elastic-apm-go-agent#logs[How to instrument your Go app with the Elastic APM Go agent].
```
bmorelli25: I agree that the content of this link is better suited for a blog post than documentation, but I'm wondering if you think it's possible/useful to include some of this information in the docs?

docs/log-correlation.asciidoc
```diff
@@ -0,0 +1,119 @@
...
+[[log-correlation-manual-unstructured]]
+==== Manual log correlation (unstructured)
+
+For correlating unstructured logs (e.g. basic printf-style logging, like the standard library's
+`log` package), then you will need to need to include the trace IDs in your log message, and then
+extract them using Filebeat.
+
+If you already have a transaction and/or span object, then you can use the
+<<transaction-tracecontext, Transaction.TraceContext>> and <<span-tracecontext, Span.TraceContext>>
+methods. The trace, transaction and span ID types all provide `String` methods that yield
```
bmorelli25: Adds the oxford comma
```suggestion
methods. The trace, transaction, and span ID types all provide `String` methods that yield
```

docs/log-correlation.asciidoc
axw: None at all, will fix.

docs/log-correlation.asciidoc
axw: I would love to see more guided walkthroughs in our docs in general. For the specific case of how to ingest JSON logs using Filebeat, to me it seems more appropriate to add it to FIlebeat docs than here. WDYT?

bmorelli25: Yeah, it's a tough line to toe. Our docs are so difficult to navigate cross-stack that I worry about users getting lost along the way. With this in mind, I've added a small amount of Filebeat information to the central documentation for this feature -- with the hopes that it will be helpful enough for users, and not be too difficult to maintain.

https://github.com/elastic/apm-server/pull/2672

axw: Cool, I'll link to that instead then.

axw: ... of course that's not actually merged yet, so I can't. I'll update the Go docs once that's merged and released.

Issue #632, opened 2019-09-05
initStackTrace doesn't return the deepest stacktrace for an error


Azuka:
**Describe the bug**
I've noticed that wrapped errors from `github.com/pkg/errors` do not always carry the full stack trace when re-wrapped downstream, ie only one level is checked.

https://github.com/elastic/apm-agent-go/blob/master/error.go#L411. For comparison, here's the Newrelic implementation: https://github.com/newrelic/go-agent/blob/c680931e5ab45efd53e933df0350ff469d8c3d24/_integrations/nrpkgerrors/nrpkgerrors.go#L28

which does a loop to find the deepest implementation of the interface 

```go
interface {
    Cause() error
}
```

**To Reproduce**
```go
func a() error {
    return errors.Errorf("something happened here")
}

func main() {
    // ... context initialized somewhere with a transaction or span
    err := errors.WithStack(a())
    apm.CaptureError(ctx, err).Send()
}
```

**Expected behavior**
The trace should show the lines from `func a()`. I can take a crack at it next week if you have any recommendations on the best way to implement.

axw:
@Azuka thanks for opening this! That is certainly a problem.

I think this will be addressed by some changes that are underway now to add support for reporting error causes, so we can see the full error chain.

There's some progress being made on the UI/design front in https://github.com/elastic/kibana/issues/40375, and there's an existing issue for the Go agent in https://github.com/elastic/apm-agent-go/issues/423.

I'd prefer to wait for for this rather than just taking the deepest error, since that may also throw away valuable information. What do you think?

Azuka:
Thanks @axw. I'm always amazed at how quickly you respond to issues here.

axw:
:)

I'll close this one and make a note of the issue in #423

PR #638, opened 2019-09-24
Report error causes

### LINKED ISSUE #423: Report complete error chains
## Describe the bug
```go
// err   generated by the net/http
apmError := apm.CaptureError(apmCtx, err)
apmError.Send()
```
net err will not execute here, exceptionData.typeName can't get the correct value.
```go
func initException(e *exceptionData, err error) {
     .......
    // Set Module, Type, Attributes, and Code.
    switch err := err.(type) {     // net err 
    case *net.OpError:
        e.typePackagePath, e.typeName = "net", "OpError"
        setAttr("op", err.Op)
        setAttr("net", err.Net)
        setAttr("source", err.Source)
        setAttr("addr", err.Addr)
    ........
}

golang net err relation:

error ----> net.Error ---> net.OpError
                  net.OpError.Err|
                                 |--------> *net.DNSError
                                 |--------> *net.InvaildAddrError
                                 |--------> *net.UnkonwnNetworkError
                                 |---------> *net.AddrError
                                 |---------> *net.DNSConfigError
                                 |---------> *os.SyscallError -----> *os.SyscallError.Err---> sys.Errno
```

This place will affect the error classification. [Please see here](https://github.com/elastic/apm-server/blob/558480cab0d98450658510891f7461c0425d9a8c/processor/error/event.go#L236-L269).I hope to get a more accurate error classification.

axw:
We will now report complete error trees in
error.exception: the top-most exception is
the provided error, and any cause/wrapped
error will be reported in an array of causes
to the server.

We will only report at most 50 errors in a tree,
and we break cycles by just ignoring references
to previously seen errors.

The error culprit will now be calculated by
traversing the tree depth-first, returning
the first non-library frame's function name.
This will resolve issues such as #632, where
an error with a complete stack trace is
wrapped by one with only a limited stack trace.

Closes #423

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
  - module/apmhttp: add WithPanicPropagation function (#611)
  - module/apmgoredis: add Client.RedisClient (#613)
  - Introduce apm.TraceFormatter, for formatting trace IDs (#635)
+ - Report error cause(s), add support for errors.Unwrap (#638)
 
 ## [v1.5.0](https://github.com/elastic/apm-agent-go/releases/tag/v1.5.0)
 
--- a/error.go
+++ b/error.go
@@ -33,6 +33,14 @@ import (
 	"go.elastic.co/apm/stacktrace"
 )
 
+const (
+	// maxErrorGraphSize is the maximum number of errors
+	// to report in an error tree. Once this number of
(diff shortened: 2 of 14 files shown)


axw:
jenkins run the tests please

PR #642, opened 2019-09-26
Add SetLabel, deprecate SetTag

### LINKED ISSUE #572: Add support for numeric/bool label values
Since APM Server 6.7, labels can have numerical or boolean values. See: https://www.elastic.co/guide/en/apm/get-started/7.2/metadata.html. We should add support for these in the agent's API, which currently supports only strings. This can be done in conjunction with  #467.

axw:
Add Context.SetLabel and SpanContext.SetLabel,
which accept interface{} values. The value may
be a nil, string, bool, or numerical type. Any other
type will be sent as a string using fmt.Sprint.

The SetTag methods remain, but marked as being
deprecated, with a notice that they will be removed
in a future major version.

Closes #572

### DIFF
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -63,7 +63,7 @@ further describe the transaction.
 [source,go]
 ----
 transaction.Result = "Success"
-transaction.Context.SetTag("region", "us-east-1")
+transaction.Context.SetLabel("region", "us-east-1")
 ----
 
 See <<context-api>> for more details on setting transaction context.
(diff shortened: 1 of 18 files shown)


docs/api.asciidoc
```diff
@@ -328,15 +328,31 @@ custom context and tags.
...
 ==== `func (*Context) SetTag(key, value string)`
 
-SetTag tags the transaction or error with the given key and value. If the
-key contains any special characters (`.`, `*`, `"`), they will be replaced
-with underscores. Values longer than 1024 characters will be truncated.
-Tags will be indexed in Elasticsearch as keyword fields.
+SetTag is equivalent to calling SetLabel with a string value.
 
-TIP: Before using custom tags, ensure you understand the different types of
+NOTE: This function is deprecated, and will be removed in a future major
```
bmorelli25: Just as an FYI, we do have fun _deprecated_ tags you can use here as well. As an example:
```
deprecated[4.1.x,Tags have been replaced with labels], Please use <<apm-add-labels,`apm.addLabels()`>> instead
```
You can see what this renders as [here](https://www.elastic.co/guide/en/apm/agent/rum-js/4.x/agent-api.html#apm-add-tags).

axw: Thanks, I'll make this change in a follow-up.

bmorelli25: I don't know that you necessarily have to update this, just wanted to show you in case you didn't know about it.

PR #643, opened 2019-09-26
Improve span duration test resilience

axw:


axw:
Failing on Travis when testing against go tip/master: reported https://github.com/golang/go/issues/34552.

PR #644, opened 2019-10-04
ci(jenkins): reuse top level agent

v1v:
### Highlights
- Let's reuse the top-level agent to reduce the number of fo requested agents and speed up just a bit the builds.
- Reset labels to be `linux && immutable` to help with what kind of OS is required to be used.
- An average improvement of ~4 minutes per build.
- Also fixed the tag regex when running the Benchmarking and some minor changes.

### Test Cases
- Benchmark


v1v:
So far some issues when running the Tests for certain go versions:

```
[2019-10-11T13:40:43.552Z] + make install check
[2019-10-11T13:40:43.552Z] $GOPATH/go.mod exists but should not
[2019-10-11T13:40:43.552Z] go get -v -t ./...
[2019-10-11T13:40:43.810Z] go get: -t flag is a no-op when using modules
[2019-10-11T13:40:43.810Z] $GOPATH/go.mod exists but should not
[2019-10-11T13:40:43.811Z] Makefile:44: recipe for target 'install' failed
```

But that particular go.mod files will exist as it's already within the project... 

No idea what to do with this particular issue

axw:
> $GOPATH/go.mod exists but should not

That means there's a `go.mod` file in the workspace root, which I don't think should be there. The repo should be cloned into `$WORKSPACE/src/go.elastic.co/apm`, right?

I don't know why your changes would cause this though.

v1v:
I've found what's causing that particular issue, there was a corner case when reusing the same worker between stages, the `checkout` stage didn't have the `skipDefaultCheckout` which caused two checkouts:
- the default one provided by the MultiBranchPipeline engine. This is not required.
- the customized one, called `gitCheckout`, which it's the one we do care. This is the required one.

So this is fixed with https://github.com/elastic/apm-agent-go/pull/644/commits/623375aa041be6a2eddb4ff57f4d9213fd4d133f

PR #645, opened 2019-10-06
Ignore sql.ErrNoRows

vutung2311:
There are some cases that gorm can return sql.ErrNoRows when using with postgres and ON CONFLICT DO NOTHING RETURNING *. I think we can safely ignore this error as we're ignoring gorm.ErrRecordNotFound

Also add second argument of redis command trace to understand more clearly about the trace.

### DIFF
--- a/module/apmgorm/apmgorm_test.go
+++ b/module/apmgorm/apmgorm_test.go
@@ -81,6 +81,7 @@ func testWithContext(t *testing.T, dsnInfo apmsql.DSNInfo, dialect string, args 
 		defer db.Close()
 		db = apmgorm.WithContext(ctx, db)
 
+		db.DropTableIfExists(&Product{})
 		db.AutoMigrate(&Product{})
 		db.Create(&Product{Code: "L1212", Price: 1000})
 
@@ -124,6 +125,7 @@ func TestWithContextNoTransaction(t *testing.T) {
 	defer db.Close()
 	db = apmgorm.WithContext(context.Background(), db)
 
+	db.DropTableIfExists(&Product{})
 	db.AutoMigrate(&Product{})
 	db.Create(&Product{Code: "L1212", Price: 1000})
 
@@ -138,6 +140,7 @@ func TestWithContextNonSampled(t *testing.T) {
 	db, err := apmgorm.Open("sqlite3", ":memory:")
 	require.NoError(t, err)
 	defer db.Close()
+	db.DropTableIfExists(&Product{})
 	db.AutoMigrate(&Product{})
 
 	_, spans, _ := apmtest.WithTransaction(func(ctx context.Context) {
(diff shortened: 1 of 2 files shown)


axw:
@vutung2311 thanks for the PR!

The Gorm change seems reasonable, but can you please add a minimal test that fails before the change?

The Redis change should be made separately, if at all - please remove that from this PR so we can focus on one change at a time. If we do make the Redis change, it should be made across all of the agents - so please open a discussion at https://discuss.elastic.co/c/apm first.

vutung2311:
> @vutung2311 thanks for the PR!
> 
> The Gorm change seems reasonable, but can you please add a minimal test that fails before the change?
> 
> The Redis change should be made separately, if at all - please remove that from this PR so we can focus on one change at a time. If we do make the Redis change, it should be made across all of the agents - so please open a discussion at https://discuss.elastic.co/c/apm first.

Understood. I'm gonna work on it.

vutung2311:
I tried quite a few ways for sqlite3 but making that test to have sql.ErrNoRows was not possible for me. I modified the test to use Postgres. I hope that won't hurt anything.

axw:
> I tried quite a few ways for sqlite3 but making that test to have sql.ErrNoRows was not possible for me. I modified the test to use Postgres. I hope that won't hurt anything.

@vutung2311 thanks for the updates. I'd like to keep running the test against sqlite3, so it can be run without starting up PostgreSQL. How about this?

```go
func TestCaptureErrors(t *testing.T) {
        t.Run("sqlite3", func(t *testing.T) {
                db, err := apmgorm.Open("sqlite3", ":memory:")
                require.NoError(t, err)
                defer db.Close()
        })

        if os.Getenv("PGHOST") == "" {
                t.Logf("PGHOST not specified, skipping")
        } else {
                t.Run("postgres", func(t *testing.T) {
                        db, err := apmgorm.Open("postgres", "user=postgres password=hunter2 dbname=test_db sslmode=disable")
                        require.NoError(t, err)
                        defer db.Close()
                })
        }
}

func testCaptureErrors(t *testing.T, db *gorm.DB, f func(ctx context.Context, db *gorm.DB)) {
        db.SetLogger(nopLogger{})
        db.AutoMigrate(&Product{})
        require.NoError(t, db.Unscoped().Delete(&Product{}).Error)

        _, spans, errors := apmtest.WithTransaction(func(ctx context.Context) {
                db = apmgorm.WithContext(ctx, db)

                // gorm.ErrRecordNotFound should not cause an error
                db.Where("code=?", "L1212").First(&Product{})

                // The postgres dialect will use QueryRow when inserting, in order to fetch
                // the inserted row ID. By using "ON CONFLICT (id) DO NOTHING", no row will
                // be inserted, and QueryRow will return sql.ErrNoRows. This should not be
                // reported as an error.
                product := Product{
                        Model: gorm.Model{
                                ID: 1001,
                        },
                        Code:  "1001",
                        Price: 1001,
                }
                require.NoError(t, db.Create(&product).Error)
                db.Set("gorm:insert_option", "ON CONFLICT (id) DO NOTHING").Create(&product)

                // invalid SQL should
                db.Where("bananas").First(&Product{})
        })
        assert.Len(t, spans, 4)
        require.Len(t, errors, 1)
        assert.Regexp(t, `.*bananas.*`, errors[0].Exception.Message)
}
```

module/apmgorm/apmgorm_test.go
```diff
@@ -148,30 +151,63 @@ func TestWithContextNonSampled(t *testing.T) {
...
+	})
+
+	if os.Getenv("PGHOST") == "" {
+		t.Logf("PGHOST not specified, skipping")
+	} else {
+		t.Run("postgres", func(t *testing.T) {
+			db, err := apmgorm.Open("postgres", "user=postgres password=hunter2 dbname=test_db sslmode=disable")
+			require.NoError(t, err)
+			testCaptureErrors(t, db)
+			defer db.Close()
```
axw: `defer db.Close()` needs to be just above `testCaptureErrors(t, db)`, in case testCaptureErrors panics

module/apmgorm/apmgorm_test.go
```diff
@@ -148,30 +151,63 @@ func TestWithContextNonSampled(t *testing.T) {
...
 
 func TestCaptureErrors(t *testing.T) {
-	db, err := apmgorm.Open("sqlite3", ":memory:")
-	require.NoError(t, err)
-	defer db.Close()
+	t.Run("sqlite3", func(t *testing.T) {
+		db, err := apmgorm.Open("sqlite3", ":memory:")
+		require.NoError(t, err)
+		testCaptureErrors(t, db)
+		defer db.Close()
```
axw: `defer db.Close()` needs to be just above `testCaptureErrors(t, db)`, in case testCaptureErrors panics

module/apmgorm/apmgorm_test.go
vutung2311: That's right. I've just fixed it.

module/apmgorm/apmgorm_test.go
vutung2311: Fixed :D

axw:
jenkins, run the tests please

axw:
Thanks again for your contribution @vutung2311!

PR #670, opened 2019-11-09
Fix deadlock in Span.reportSelfTime()

nogoegst:
A child tries acquire lock of its parent while the parent waits on transaction lock which is acquired by the child.

### DIFF
--- a/span.go
+++ b/span.go
@@ -330,15 +330,16 @@ func (s *Span) reportSelfTime() {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


axw:
Hi @nogoegst 

Can you please provide more specific details about where you think the deadlock is? I'm not seeing it. Ideally, provide a test case which illustrates the issue.

axw:
As I've not heard back, and I'm not convinced there's an issue, I'm closing. Feel free to reopen if you can answer my question above.

PR #671, opened 2019-11-11
module/apmnegroni

earlzo:
apmnegroni and apmhttp have a lot of duplicate code.

shall we extract some common parts for dry code?

@axw


axw:
> apmnegroni and apmhttp have a lot of duplicate code.
>
> shall we extract some common parts for dry code?

How about we instead use `apmhttp` internally? One way to do this would be to use `apmhttp.Wrap`, calling a top-level http.HandlerFunc`, which would get the "next" handler from the request context. Something like this:

```go
func Middleware(o ...Option) negroni.Handler {
    opts := apmhttpServerOptions(o...) // translate []Option to []apmhttp.ServerOption
    return &middleware{handler: apmhttp.Wrap(http.HandlerFunc(nextHandler), o...)}
}

type middleware struct {
        handler http.Handler
}

func (m middleware) ServeHTTP(w http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
        r = r.WithContext(context.WithValue(r.Context(), nextKey{}, next))
        m.handler.ServeHTTP(w, r)
}

type nextKey struct{}

func nextHandler(w http.ResponseWriter, r *http.Request) {
        next := r.Context().Value(nextKey{}).(http.HandlerFunc)
        next(w, r)
}
```

earlzo:
LGTM.

If we use `apmhttp` internally, why not use `apmhttp.ServerOption` for `Middleware` rather than write a`apmhttpServerOptions ` for `Option`?

axw:
> If we use apmhttp internally, why not use apmhttp.ServerOption for Middleware rather than write aapmhttpServerOptions for Option?

This could become a problem if we wanted to add an option specific to Negroni, which doesn't fit in the apmhttp package. I don't like repeating the options either, but I think it provides the cleanest API.

earlzo:
how about that:

1. 

```go
type Option apmhttp.ServerOption

type RecoveryFunc apmhttp.RecoveryFunc

type RequestNameFunc apmhttp.RequestNameFunc

var WithServerRequestName = apmhttp.WithServerRequestName
```

or 

2.

```
type Option func(*middleware)

type RecoveryFunc apmhttp.RecoveryFunc

type RequestNameFunc apmhttp.RequestNameFunc

func WithServerRequestName(r RequestNameFunc) Option {
	if r == nil {
		panic("r == nil")
	}
	return func(m *middleware) {
		m.requestName = r
	}
}
```

I prefer the first.

axw:
> var WithServerRequestName = apmhttp.WithServerRequestName

I'd prefer not to have package level `var`s like this.

We could combine them a bit:

```go
type Option apmhttp.ServerOption

func WithServerRequestName(r apmhttp.RequestNameFunc) Option {
	return Option(apmhttp.WithServerRequestName(r))
}
```

earlzo:
Yes,  you are right. 

I will start to rewrite this module.

Do I need other steps to complete this pr? Like documenting or something else.

module/apmnegroni/go.mod
```diff
@@ -0,0 +1,11 @@
...
+
+go 1.13
+
+require (
+	github.com/google/go-cmp v0.3.1 // indirect
+	github.com/stretchr/testify v1.3.0
+	github.com/urfave/negroni v1.0.0
+	go.elastic.co/apm v1.5.0
+	go.elastic.co/apm/module/apmhttp v1.5.0
+)
```
axw:
```suggestion
)

replace go.elastic.co/apm/module/apmhttp => ../apmhttp

replace go.elastic.co/apm => ../..
```

earlzo:
@axw you can review my code now

go.mod
```diff
@@ -2,15 +2,17 @@ module go.elastic.co/apm
 
 require (
 	github.com/armon/go-radix v1.0.0
-	github.com/davecgh/go-spew v1.1.1 // indirect
```
axw: There are a lot of go.mod changes unrelated to apmnegroni. Can you please run `go mod tidy` to revert them? There's a script to list the modules which you can run from the repo root dir to make this easier:

```sh
for dir in $(./scripts/moduledirs.sh); do (cd $dir && go mod tidy -v); done
```

docs/instrumenting.asciidoc
```diff
@@ -291,6 +291,36 @@ func main() {
 }
 ----
 
+[[builtin-modules-apmnegroni]]
```
axw: Thanks for adding the docs!

module/apmnegroni/middleware.go
```diff
@@ -0,0 +1,122 @@
...
+// Middleware returns a new negroni middleware handler for tracing
+// requests and reporting errors.
+//
+// This middleware will recover and report panics, so it can
+// be used instead of the standard negroni.Recovery middleware.
+//
+// By default, the middleware will use apm.DefaultTracer.
+// Use WithTracer to specify an alternative tracer.
+func Middleware(o ...Option) negroni.Handler {
+	m := &middleware{
```
axw: Make this a non-pointer type (`m := middleware{...}`)? The ServeHTTP method is on the non-pointer type middleware

module/apmnegroni/middleware.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+type nextKey struct{}
+
+func nextHandler(w http.ResponseWriter, r *http.Request) {
+	next := r.Context().Value(nextKey{}).(http.HandlerFunc)
+	next(w, r)
+}
+
+var _ negroni.Handler = middleware{}
```
axw: I think we can remove this, it's already covered by the tests.

axw:
jenkins run the tests please

axw:
Seems to be something wrong with the CI docs builder, but I've verified the docs build locally.

Thanks for the PR @earlzo - merging now!

PR #673, opened 2019-11-12
[apm-ci] branches/tags with the format ^module/ are filtered

### LINKED ISSUE #608: Review which tags we should build
right now, we build every tag is in the repository, but maybe it is not needed there are a bunch of tags like `module/*` that probably can be skipped.

v1v:
The pipeline should care only for those branches/tags which are not ^module based

Closes #608 

### Changes
image

image

### DIFF
--- a/.ci/jobs/defaults.yml
+++ b/.ci/jobs/defaults.yml
@@ -11,21 +11,19 @@
     view: APM-CI
     project-type: multibranch
     logrotate:
-      daysToKeep: 30
       numToKeep: 100
-    number-to-keep: '5'
-    days-to-keep: '1'
     concurrent: true
     node: linux
     script-path: Jenkinsfile
     scm:
     - github:
-        branch-discovery: all
+        branch-discovery: no-pr
         discover-pr-forks-strategy: merge-current
         discover-pr-forks-trust: permission
         discover-pr-origin: merge-current
         discover-tags: true
         notification-context: 'apm-ci'
+        head-filter-regex: '^(?!module/).*$'
         repo: apm-agent-go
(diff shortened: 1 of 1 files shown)


.ci/jobs/defaults.yml
```diff
@@ -11,21 +11,19 @@
...
       numToKeep: 100
-    number-to-keep: '5'
-    days-to-keep: '1'
     concurrent: true
     node: linux
     script-path: Jenkinsfile
     scm:
     - github:
-        branch-discovery: all
+        branch-discovery: no-pr
```
v1v: This will avoid any branches within this repo which were raised as PRs too.

.ci/jobs/defaults.yml
```diff
@@ -11,21 +11,19 @@
     view: APM-CI
     project-type: multibranch
     logrotate:
-      daysToKeep: 30
       numToKeep: 100
```
v1v: Let's only care about the last 100 builds

.ci/jobs/defaults.yml
```diff
@@ -11,21 +11,19 @@
...
     scm:
     - github:
-        branch-discovery: all
+        branch-discovery: no-pr
         discover-pr-forks-strategy: merge-current
         discover-pr-forks-trust: permission
         discover-pr-origin: merge-current
         discover-tags: true
         notification-context: 'apm-ci'
+        head-filter-regex: '^(?!module/).*$'
```
v1v: This is the regex to ensure branches/tags which are not matching `^module/` should not be ignored (sorry about my double negation!)  :)

PR #675, opened 2019-11-14
Use/send "traceparent" header

### LINKED ISSUE #503: Implement full support for Trace-Context
See https://github.com/elastic/apm/issues/71

axw:
We will now look for and propagate the W3C "traceparent" header. We still look for "Elastic-Apm-Traceparent" and use it if present, and otherwise we use the "traceparent" header. We will always send "traceparent" in outgoing requests, and optionally send the "Elastic-Apm-Traceparent" header.

We introduce the `ELASTIC_APM_USE_ELASTIC_TRACEPARENT_HEADER` environment variable, which controls whether or not we send the "Elastic-Apm-Tracparent" header; currently this defaults to true, but we will later flip the value and eventually remove the config altogether. The value is accessed
through the new method, `Transaction.ShouldPropagateLegacyHeader`.

Part of #503 (still need to implement support for "tracestate" to close it)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@
  - Server URL path is cleaned/canonicalizsed in order to avoid 301 redirects (#658)
  - `context.request.socket.remote_address` now reports the peer address (#662)
  - Experimental support for periodic CPU/heap profiling (#666)
+ - module/apmnegroni: introduce tracing Negroni middleware (#671)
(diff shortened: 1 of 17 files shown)


beniwohli:
@axw did you leave the new setting undocumented on purpose?

(was hoping to lift the docs from you :D )

axw:
@beniwohli I don't recall now. I may have left it out on the basis that nobody should care about it at the moment, but we'll need to document it sooner or later...

PR #678, opened 2019-11-20
ci(jenkins): enable opbeans syncup

v1v:
When a new tag, aka a new release, is created then, the automation will kick the opbeans build which consists of the following tasks:
- [x] Bump dependency versions in the opbeans-go
- [x] Push changes. (Implementation within the opbeans-go but the call it's done within this pipeline)
- [x] Create a tag to trigger the release pipeline in the opbeans-go. That particular tag will match with the same one that the one in the agent itself.

This particular pipeline will push the changes into the `master` branch for the opbeans-go and tag it with the name of the tag version too. Therefore, the opbeans-go pipeline will trigger two builds, the one for its master branch and the other one for the just created tag. This particular flow is implemented within the opbeans-go.

### UTs

- PRs UI view

image

- Release tags UI view

A simplistic view

image

The above build bumped the version to `v1.6.0` and caused to trigger the opbeans-go pipeline

image

image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -262,7 +262,6 @@ pipeline {
         stage('Opbeans') {
           environment {
             REPO_NAME = "${OPBEANS_REPO}"
-            BRANCH_NAME = 'v1.6.0'
           }
           steps {
             deleteDir()
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -63,7 +63,12 @@ pipeline {
         stage('Tests') {
           options { skipDefaultCheckout() }
           when {
-            expression { return params.test_ci }
+            allOf {
+              expression { return params.test_ci }
+              not {
+                branch 'PR-678'
+              }
+            }
           }
           steps {
             withGithubNotify(context: 'Tests', tab: 'tests') {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -220,6 +221,32 @@ pipeline {
...
+        stage('Opbeans') {
+          environment {
+            REPO_NAME = "${OPBEANS_REPO}"
+          }
+          steps {
+            deleteDir()
+            dir("${OPBEANS_REPO}"){
+              git credentialsId: 'f6c7695a-671e-4f4f-a331-acdce44ff9ba',
+                  url: "git@github.com:elastic/${OPBEANS_REPO}.git"
+              sh script: ".ci/bump-version.sh ${env.BRANCH_NAME}", label: 'Bump version'
```
v1v: BRANCH_NAME will match the tag release, as the when condition https://github.com/elastic/apm-agent-go/pull/678/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R228 filters what tags should trigger this particular stage

Jenkinsfile
```diff
@@ -220,6 +221,32 @@ pipeline {
...
+    stage('Release') {
+      options { skipDefaultCheckout() }
+      when {
+        beforeAgent true
+        tag pattern: 'v\\d+\\.\\d+\\d+', comparator: 'REGEXP'
+      }
+      stages {
+        stage('Opbeans') {
+          environment {
+            REPO_NAME = "${OPBEANS_REPO}"
```
v1v: This is required to use the git*** steps as it does infer them under the hood

PR #687, opened 2019-12-03
Reduce memory on rand source in Transaction code

jasonmoo-we:
This PR addresses high memory use in the transaction code.  We found that large amounts of memory were being allocated to the `StartTransactionOptions` method when creating a random source on each transaction.  This is not required to achieve threadsafe, uniq ids.  

Screen Shot 2019-12-03 at 4 42 28 PM

I implemented a simple mutexted global random source that satisfies the needs of the code.  

The existing benchmark for Transaction was masking the memory in use.  For each new math/rand rng source, >5kb is allocated.  But since the benchmark was starting/ending inside the loop block it likely was getting elided or using stack memory.  Whatever the reason, when removing the end method, shows the expected memory use.  I have removed it for this PR.

Example benchmarks showing memory use after removing the end method and the performance of this PR.
```bash
$ ~/go/src/github.com/jasonmoo-we/apm-agent-go master: go test -v -run xxx -benchmem -bench Transaction
goos: darwin
goarch: amd64
pkg: go.elastic.co/apm
BenchmarkTransaction-8   	 5158528	       228 ns/op	      81 B/op	       1 allocs/op
PASS
ok  	go.elastic.co/apm	5.254s
$ ~/go/src/github.com/jasonmoo-we/apm-agent-go master: g df
diff --git a/transaction_test.go b/transaction_test.go
index 935cd82..f0be6d2 100644
--- a/transaction_test.go
+++ b/transaction_test.go
@@ -177,8 +177,8 @@ func BenchmarkTransaction(b *testing.B) {
 		rand := rand.New(rand.NewSource(globalRand.Int63()))
 		mu.Unlock()
 		for pb.Next() {
-			tx := tracer.StartTransaction(names[rand.Intn(len(names))], "type")
-			tx.End()
+			tracer.StartTransaction(names[rand.Intn(len(names))], "type")
+			// tx.End()
 		}
 	})
 }
$ ~/go/src/github.com/jasonmoo-we/apm-agent-go master: go test -v -run xxx -benchmem -bench Transaction
goos: darwin
goarch: amd64
pkg: go.elastic.co/apm
BenchmarkTransaction-8   	  527340	      2723 ns/op	    6465 B/op	       7 allocs/op
PASS
ok  	go.elastic.co/apm	3.755s
$ ~/go/src/github.com/jasonmoo-we/apm-agent-go jasonmoo-we/reduce-memory-on-rand-sourcing: go test -v -run xxx -benchmem -bench Transaction
goos: darwin
goarch: amd64
pkg: go.elastic.co/apm
BenchmarkTransaction-8   	 2269914	       553 ns/op	     896 B/op	       3 allocs/op
PASS
ok  	go.elastic.co/apm	2.009s
```

### DIFF
--- a/span.go
+++ b/span.go
@@ -108,7 +108,7 @@ func (tx *Transaction) StartSpanOptions(name, spanType string, opts SpanOptions)
 		if opts.SpanID.Validate() == nil {
 			span.traceContext.Span = opts.SpanID
 		} else {
-			binary.LittleEndian.PutUint64(span.traceContext.Span[:], tx.rand.Uint64())
+			binary.LittleEndian.PutUint64(span.traceContext.Span[:], globalRand.Uint64())
 		}
 		span.stackFramesMinDuration = tx.spanFramesMinDuration
 		span.stackTraceLimit = tx.stackTraceLimit
--- a/transaction.go
+++ b/transaction.go
@@ -25,6 +25,27 @@ import (
 	"time"
 )
 
+type threadSafeRand struct {
+	me   sync.Mutex
+	rand *rand.Rand
+}
+
+func newThreadSafeRand() *threadSafeRand {
+	var seed int64
+	if err := binary.Read(cryptorand.Reader, binary.LittleEndian, &seed); err != nil {
+		seed = time.Now().UnixNano()
+	}
+	return &threadSafeRand{rand: rand.New(rand.NewSource(seed))}
+}
+
+func (tsr *threadSafeRand) Uint64() uint64 {
+	tsr.me.Lock()
+	defer tsr.me.Unlock()
+	return tsr.rand.Uint64()
+}
+
+var globalRand = newThreadSafeRand()
+
 // StartTransaction returns a new Transaction with the specified
 // name and type, and with the start time set to the current time.
 // This is equivalent to calling StartTransactionOptions with a
@@ -45,11 +66,6 @@ func (t *Tracer) StartTransactionOptions(name, transactionType string, opts Tran
 			},
 			spanTimings: make(spanTimingsMap),
 		}
-		var seed int64
-		if err := binary.Read(cryptorand.Reader, binary.LittleEndian, &seed); err != nil {
-			seed = time.Now().UnixNano()
-		}
-		td.rand = rand.New(rand.NewSource(seed))
 	}
 	tx := &Transaction{tracer: t, TransactionData: td}
 
@@ -66,14 +82,14 @@ func (t *Tracer) StartTransactionOptions(name, transactionType string, opts Tran
 		if opts.TransactionID.Validate() == nil {
 			tx.traceContext.Span = opts.TransactionID
 		} else {
-			binary.LittleEndian.PutUint64(tx.traceContext.Span[:], tx.rand.Uint64())
+			binary.LittleEndian.PutUint64(tx.traceContext.Span[:], globalRand.Uint64())
 		}
 	} else {
 		// Start a new trace. We reuse the trace ID for the root transaction's ID
 		// if one is not specified in the options.
 		root = true
-		binary.LittleEndian.PutUint64(tx.traceContext.Trace[:8], tx.rand.Uint64())
-		binary.LittleEndian.PutUint64(tx.traceContext.Trace[8:], tx.rand.Uint64())
+		binary.LittleEndian.PutUint64(tx.traceContext.Trace[:8], globalRand.Uint64())
+		binary.LittleEndian.PutUint64(tx.traceContext.Trace[8:], globalRand.Uint64())
 		if opts.TransactionID.Validate() == nil {
 			tx.traceContext.Span = opts.TransactionID
 		} else {
@@ -300,7 +316,6 @@ type TransactionData struct {
 	spansDropped  int
 	childrenTimer childrenTimer
 	spanTimings   spanTimingsMap
-	rand          *rand.Rand // for ID generation
 	// parentSpan holds the transaction's parent ID. It is protected by
 	// mu, since it can be updated by calling EnsureParent.
 	parentSpan SpanID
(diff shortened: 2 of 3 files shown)


jasonmoo-we:
Just signed CLA.

axw:
I believe what you're seeing in the profile is expected due to high concurrency, but I'd just like to clear up a few things first before we get into possible solutions.

> We found that large amounts of memory were being allocated to the StartTransactionOptions method when creating a random source on each transaction.

You should only get a new random source on _every_ transaction if you don't end them. `Transaction` objects are pooled, and their random source along with them; when you end a transaction it gets put back into the pool, and it'll either be reused or (eventually) garbage collected.

> This is not required to achieve threadsafe, uniq ids

That's true, but goroutine-safety is not the whole story. The approach taken so far has been to avoid serialising callers of `StartTransaction*` and `StartSpan*` at the cost of additional memory usage, so as not to affect application throughput. We do serialise calls to `StartSpan*` within a transaction, but not across them.

> The existing benchmark for Transaction was masking the memory in use. For each new math/rand rng source, >5kb is allocated. But since the benchmark was starting/ending inside the loop block it likely was getting elided or using stack memory. Whatever the reason, when removing the end method, shows the expected memory use. I have removed it for this PR.

This code change leads to incorrect usage of the API (transactions aren't ended), and so it is expected to incur worst-case performance. Unless you've got some custom instrumentation (from your profile output it appears you're just using `apmhttp`), this isn't really the same thing.

When transactions are ended (i.e. in the original benchmark code), the memory usage should be proportional to the maximum number of Transaction objects active at any given time.

> I implemented a simple mutexted global random source that satisfies the needs of the code.

Probably goes without saying given the above statements, but just to be crystal clear: the problem with this approach is that it can have a significant impact on the throughput of the application, due to serialising all callers of `StartTransaction*` and `StartSpan*` through a single global mutex. This becomes more noticeable as you instrument more code, creating additional spans.

Having said that, for some applications (e.g. just using apmhttp with no additional spans) that overhead will probably not be very noticeable. If this level of concurrency is not an issue for you, and you're instead more concerned about keeping memory usage down, then we could introduce an option to override how a `math.Rand` is acquired when a new `Transaction` object is allocated. Something like this:

```go
// SetTransactionRand sets a function to use for acquiring a `math/rand.Rand`, when a
// new Transaction object is created. The Rand object will be used for generating the
// transaction's ID, as well as IDs for that transaction's spans.
//
// If a new Rand object is allocated on each call (the default behaviour), then the result
// does not need to be safe for concurrent use, as the object will be owned by a single
// transaction, and calls to start spans will be serialised within that transaction.
function (t *Tracer) SetTransactionRand(f func() *rand.Rand)
```

WDYT?

axw:
Another option would be to use a separate `sync.Pool` for the rand.Source, decoupling its lifetime from the transaction:

```diff
diff --git a/transaction.go b/transaction.go
index 77ca3862..4a2a64e3 100644
--- a/transaction.go
+++ b/transaction.go
@@ -25,6 +25,25 @@ import (
        "time"
 )
 
+var rngPool = sync.Pool{
+       New: func() interface{} {
+               var seed int64
+               if err := binary.Read(cryptorand.Reader, binary.LittleEndian, &seed); err != nil {
+                       seed = time.Now().UnixNano()
+               }
+               return rand.NewSource(seed).(rand.Source64)
+       },
+}
+
+type poolRand struct{}
+
+func (r poolRand) Uint64() uint64 {
+       rng := rngPool.Get().(randSource)
+       v := rng.Uint64()
+       rngPool.Put(rng)
+       return v
+}
+
 // StartTransaction returns a new Transaction with the specified
 // name and type, and with the start time set to the current time.
 // This is equivalent to calling StartTransactionOptions with a
@@ -45,11 +64,7 @@ func (t *Tracer) StartTransactionOptions(name, transactionType string, opts Tran
                        },
                        spanTimings: make(spanTimingsMap),
                }
-               var seed int64
-               if err := binary.Read(cryptorand.Reader, binary.LittleEndian, &seed); err != nil {
-                       seed = time.Now().UnixNano()
-               }
-               td.rand = rand.New(rand.NewSource(seed))
+               td.rand = poolRand{}
        }
        tx := &Transaction{tracer: t, TransactionData: td}
 
@@ -302,7 +317,7 @@ type TransactionData struct {
        spansDropped  int
        childrenTimer childrenTimer
        spanTimings   spanTimingsMap
-       rand          *rand.Rand // for ID generation
+       rand          randSource // for ID generation
        // parentSpan holds the transaction's parent ID. It is protected by
        // mu, since it can be updated by calling EnsureParent.
        parentSpan SpanID
@@ -321,3 +336,7 @@ func (td *TransactionData) reset(tracer *Tracer) {
        td.spanTimings.reset()
        tracer.transactionDataPool.Put(td)
 }
+
+type randSource interface {
+       Uint64() uint64
+}
```

This avoids synchronisation, and the memory usage of rand.Sources should be proportional to GOMAXPROCS rather than the number of concurrent transactions. It would probably be reasonable to make this the default.

jasonmoo-we:
Let's take a step back.  I'm familiar with contention and the approaches you mentioned.  I didn't include them because I was trying to get a quick fix into something that appears to be "good enough".  The likelyhood of having contention on such a cheap function seemed unrealistic and with each rand source costing 6kb it seemed costly to have a new one per transaction.  If you want to use a sync pool go for it.

But what I'm curious about is:  All we did was wrap 2 http handlers with the apmhttp lib.  These handlers are not failing or doing anything interesting at all.  And yet we're seeing 20% (160mb) on the heap just for math rand sources.  Why is that?  Or rather, what conditions would cause that with vanilla apmhttp lib use?  For reference this http application runs at about 17mb of memory under normal traffic without apmhttp instrumentation.

jasonmoo-we:
Ok.  I actually found the source of the memory "bloat" and it on our end.  Had a situation where an http handler was blocked from exiting which would have caused the transaction to never end and would have produced the output seen here.  I'm going to close this PR.  Thanks for the in-depth talk through the issue, none-the-less.

axw:
Glad you got it sorted!

Maybe the underlying issue would have been more obvious if we had a graph showing the number of running goroutines. This will be coming at some point: the data is already being recorded, we just need to update the Kibana APM app to graph it.

If you have any other ideas on what would have made it easier to spot the root cause, I'd be happy to hear them.

PR #690, opened 2019-12-16
Add "tracestate" propagation

### LINKED ISSUE #503: Implement full support for Trace-Context

### LINKED ISSUE #685: Document ELASTIC_APM_USE_ELASTIC_TRACEPARENT_HEADER config
See https://github.com/elastic/apm-agent-go/pull/675#issuecomment-560091774 - we need to document ELASTIC_APM_USE_ELASTIC_TRACEPARENT_HEADER.

axw:
Add the field TraceContext.State, which has the new
type TraceState. TraceState is an opaque type, which
internally contains a list of TraceStateEntry objects.

When starting a transaction with StartTransactionOptions,
the TraceContext.State field will only be considered if
both TraceContext.Trace and TraceContext.State are valid.
If this is true, then the state will be stored on the
transaction, and spans created using the transaction.

Server instrumentation in apmhttp, apmgrpc, and apmot
will extract and parse "tracestate" headers, while client
instrumentation will inject "tracestate" headers.

Add missing docs for `ELASTIC_APM_USE_ELASTIC_TRACEPARENT_HEADER`.

Closes #503
Closes #685

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -514,3 +514,19 @@ Activate APM Agent Configuration via Kibana. By default the agent will poll the 
 for agent configuration changes. This can be disabled by changing the setting to `false`.
 
 NOTE: This feature requires APM Server v7.3 or later and that the APM Server is configured with `kibana.enabled: true`.
+
+[float]
+[[config-use-elastic-traceparent-header]]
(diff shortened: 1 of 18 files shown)


axw:
jenkins run the tests

PR #706, opened 2020-01-23
Upgrading apmgopg dependency on gopg from v8 to v9

c16a:


axw:
@c16a thanks for the PR! This is a breaking change for existing users, so we can't do this without a major version bump of the agent - and I don't think this change warrants that. Instead of changing the existing module, I would suggest introducing "apmgopgv9", similar to how we have "apmecho" and "apmechov4".

c16a:
Raised #709 with independent module

PR #719, opened 2020-02-18
[docs] Add EOL table

bmorelli25:
## What does this pull request do?

This PR defines end of life and end of maintenance dates for the Go Agent. EOL dates were determined by going through old releases, finding the GA release date of the major Go Agent versions, and adding 18 months.

## Why is it important?

We love all our products, but sometimes we must say goodbye to a release so that we can continue moving forward on future development and innovation.

## What does EOL/Maintenance mean?

> When a version goes out of maintenance, it is no longer guaranteed to get bug fixes. Between the end of maintenance and EOL, bug fixes will be discretionary, e.g. critical security issues. Once it's EOL, it's no longer given any support attention; issues related to it will be closed with WONTFIX.

## Related issues

For elastic/apm#173.

## Screenshots

Screen Shot 2020-02-18 at 1 16 11 PM


bmorelli25:
Question. Does it make more sense to wait to define EOL for a major until after a new major has been released? Through opening these PRs in the Agent repos, I've discovered the release cadence of majors for most of our agents is much slower than the stack. Following the 18 month rule, we're likely to hit EOL for 1.x before 2.0 is released.

axw:
@bmorelli25 shouldn't the list have 1.0.x, 1.1.x, etc.? And since 1.7.0 was released on Jan 10 this year, then the EOL for that is much further out.

bmorelli25:
You're right! Phew. Those dates make a lot more sense.

docs/upgrading.asciidoc
```diff
@@ -0,0 +1,32 @@
...
+We love all our products, but sometimes we must say goodbye to a release so that we can continue moving
+forward on future development and innovation.
+Our https://www.elastic.co/support/eol[End of life policy] defines how long a given release is considered supported,
+as well as how long a release is considered still in active development or maintenance.
+The table below is a simplified description of this policy.
+
+[options="header"]
+|====
+|Agent version |EOL Date |Maintained until
+|1.7.x |2021-07-09 |1.8
```
axw:
```suggestion
|1.7.x |2021-07-09 |1.8.0
```

etc., for consistency with https://www.elastic.co/support/eol. I've yet to issue a patch release, but in theory we do semver.

PR #721, opened 2020-02-21
fixing wrong documents

insukcho:
- change gin gonic link
- add missing import in beego module snippet
- fix wrong controller name in beego module snippet


insukcho:
It seems my local machine git user name setting should be aligned with the corporate one.  I will raise it again. Sorry for the noise.

PR #737, opened 2020-03-26
Introduce "recording" config

axw:
We introduce the ELASTIC_APM_RECORDING configuration.
This is a boolean configuration that defaults to true,
controlling whether events are recorded and sent. When
recording is true there should be no change; when false:

 - Transactions will always be recorded as "unsampled",
   and will be silently discarded when ended, without
   affecting tracer statistics
 - Spans will all be dropped by virtue of transactions
   all being unsampled
 - Captured errors will not have details filled in, and
   will be silently discarded when "sent" without affecting
   tracer statistics
 - Breakdown metrics will not be updated
 - Metrics gathering will be disabled

Recording can be updated via central config and by using
the new Tracer.SetRecording method.

We also introduce a new Tracer.Recording method which
reports whether events are being recorded. If the tracer
is inactive (Tracer.Active returns false), then
Tracer.Recording will also return false. This new method
can be used by instrumentation to avoid expensive
instrumentation paths when recording is disabled. We have
updated all provided instrumentation modules to use the
new Tracer.Recording method instead of Tracer.Active.

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -213,6 +213,20 @@ Enable or disable the agent. If set to false, then the Go agent does not send
 any data to the Elastic APM server, and instrumentation overhead is minimized.
 
 [float]
+[[config-recording]]
+=== `ELASTIC_APM_RECORDING`
+
+[options="header"]
+|============
+| Environment             | Default | Example
+| `ELASTIC_APM_RECORDING` | true    | `false`
+|============
+
+Enable or disable recording of events. If set to false, then the Go agent does
+send any events to the Elastic APM server, and instrumentation overhead is
+minimized, but the agent will continue to poll the server for configuration changes.
+
+[float]
 [[config-global-labels]]
 === `ELASTIC_APM_GLOBAL_LABELS`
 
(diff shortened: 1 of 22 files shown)


axw:
@bmorelli25 could you please take a gander at the docs addition? https://github.com/elastic/apm-agent-go/pull/737/files#diff-01c1a79a5a1e1b67003a1885365f2cf5

axw:
Leaving this open until https://github.com/elastic/kibana/issues/61179 is done, so I can test e2e

docs/configuration.asciidoc
```diff
@@ -212,6 +212,20 @@ You must use the query bar to filter for a specific environment in versions prio
...
+[[config-recording]]
+=== `ELASTIC_APM_RECORDING`
+
+[options="header"]
+|============
+| Environment             | Default | Example
+| `ELASTIC_APM_RECORDING` | true    | `false`
+|============
+
+Enable or disable recording of events. If set to false, then the Go agent does
```
bmorelli25:
```suggestion
Enable or disable recording of events. If set to false, then the Go agent does not
```

axw: Thanks, we are indeed - sorry for the waste of time.

PR #738, opened 2020-03-27
ci(jenkins): disable stages for only docs

v1v:
### What

- Any builds for any branch or PR that only contains changes related to the asciidoc then it will skip all the stages but the `checkout`

### Why

Docs validation happens with the `elasticsearch-ci/docs` therefore there is no need to validate anything in our end.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -58,8 +58,8 @@ pipeline {
             stash allowEmpty: true, name: 'source', useDefaultExcludes: false
             script {
               dir("${BASE_DIR}"){
-                // Skip all the stages except docs for PR's with asciidoc changes only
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -201,8 +219,9 @@ pipeline {
       when {
         beforeAgent true
         allOf {
+          expression { return env.ONLY_DOCS == "false" }
           anyOf {
-            environment name: 'GIT_BUILD_CAUSE', value: 'pr'
+            changeRequest()
```
cachedout: I'm not sure I understand this change in the context of the goal of the PR. Could you explain? Thanks!

v1v: Good catch, it's unrelated to this one, GIT_BUILD_CAUSE is an env variable that we do generate on the fly with the gitCheckout, and the `changeRequest` is the builtin step to be used. Then we don't need any specific customisation.

PR #744, opened 2020-04-02
Add support for errors returning *runtime.Frames

rliebz:
The standard library has a built-in concept of stack traces in the runtime package. While there is no standard of how errors are tied to stack traces, the current implementation of requires either stack traces from this library directly, or from github.com/pkg/errors.

With this change, any library that implements the following interface would have their stack traces supported within APM:

```go
type runtimeStackTracer interface {
  StackTrace() *runtime.Frames
}
````

The advantage to this is that now an error-handling library with stack-trace support can make itself compatible with this library without requiring go.elastic.co/apm or github.com/pkg/errors as a dependency. They only have to import the "runtime" package.

For reference, here is one such library: <https://github.com/morningconsult/serrors>

### DIFF
--- a/error.go
+++ b/error.go
@@ -23,6 +23,7 @@ import (
 	"net"
 	"os"
 	"reflect"
+	"runtime"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


axw:
jenkins run the tests please

rliebz:
Thanks!

PR #750, opened 2020-04-21
Implement CloseIdleConnections on apmelasticsearch

pebrc:
The apmelasticsearch `roundTripper` did not implement `CloseIdleConnections` and we (@charith-elastic who did the investigation) believe we observed a goroutine leak as a consequence of that. 

This PR adds support for `CloseIdleConnections` and propagates it to the wrapped transport/ `RoundTripper` if it implements it.


axw:
Yikes. Thanks for the PR!

Can you please add a mention of this in https://github.com/elastic/apm-agent-go/blob/master/CHANGELOG.asciidoc, in the "unreleased" section?

axw:
Test failure is unrelated

PR #754, opened 2020-04-30
set url.full even when path is empty

graphaelli:
`url.full` is not set when path is empty but is when rootless.  This makes a difference in the APM UI:

## Empty Before

image

## Empty After

image

## Rootless

image

### DIFF
--- a/model/marshal_test.go
+++ b/model/marshal_test.go
@@ -401,12 +401,14 @@ func TestMarshalURL(t *testing.T) {
 }
 
 func TestMarshalURLPathEmpty(t *testing.T) {
-	var empty fastjson.Writer
-	(&model.URL{
+	in := model.URL{
 		Hostname: "example.com",
 		Path:     "",
 		Protocol: "http",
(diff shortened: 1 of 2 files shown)


model/marshal_test.go
```diff
@@ -400,6 +400,20 @@ func TestMarshalURL(t *testing.T) {
 	assert.Equal(t, in, out)
 }
 
+func TestMarshalURLPathEmpty(t *testing.T) {
+	var empty fastjson.Writer
+	(&model.URL{
+		Hostname: "example.com",
+		Path:     "",
+		Protocol: "http",
+	}).MarshalFastJSON(&empty)
```
axw, line 409:
```suggestion
	in := model.URL{
		Hostname: "example.com",
		Path:     "",
		Protocol: "http",
	}

	var w fastjson.Writer
	w.MarshalFastJSON(&w)
```

For consistency with other tests.

axw:
Go 1.8 issue is unrelated, merging. Thanks!

PR #779, opened 2020-05-28
Typo on ELASTIC_APM_API_REQUEST_TIME description

hbizunehe:
The ELASTIC_APM_API_REQUEST_TIME description has typo and may need to be updated from
Amount of time to wait before ending a request to the Elastic APM server.
to
Amount of time to wait before sending a request to the Elastic APM server.

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -349,7 +349,7 @@ The host name to use when sending error and transaction data to the APM server.
 | `ELASTIC_APM_API_REQUEST_TIME` | `10s`
 |============
 
-Amount of time to wait before ending a request to the Elastic APM server.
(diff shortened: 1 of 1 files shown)


axw:
@hbizunehe thanks for the PR!

Although it might sound a bit odd, it's correct as it is. The way the agent works is by streaming events: when the first event (transaction, span, etc.) occurs, a new stream is started. After `ELASTIC_APM_API_REQUEST_TIME` elapses the stream is ended, and a new stream will be started when the next event occurs, and so on.

I'll close this PR, but I'd be happy to accept a change that clarifies the above, if you have some ideas. If you do, please do [sign the CLA](https://www.elastic.co/contributor-agreement), as we cannot merge changes without it.

Thanks again!

PR #785, opened 2020-06-29
Duplicated word

daehli:
also also was duplicated


axw:
@daehli thanks for the PR! To merge this change (despite its size), we would need you to [sign the CLA](https://www.elastic.co/contributor-agreement). If you prefer not to do that, then I'll close this and recreate the change.

PR #788, opened 2020-07-19
add httptrace option

graphaelli:
image


module/apmhttp/client.go
```diff
@@ -70,6 +70,7 @@ type roundTripper struct {
 	r              http.RoundTripper
 	requestName    RequestNameFunc
 	requestIgnorer RequestIgnorerFunc
+	requestTracer  *requestTracer
```
graphaelli, line 73: considered adding two functions here - for start and end - instead of a pointer

module/apmhttp/clienttrace.go
```diff
@@ -0,0 +1,97 @@
...
+
+			// When connection is re-used, DNS/TCP/TLS hooks not called.
+			if r.Connect == nil {
+				// TODO
+			}
+		},
+
+		GotFirstResponseByte: func() {
+			r.Server.End()
+			r.Transfer = r.tx.StartSpan("Transfer", "http.transfer", nil)
```
axw, line 85: Maybe call this "Response"?

module/apmhttp/clienttrace.go
```diff
@@ -0,0 +1,97 @@
...
+
+	DNS,
+	Connect,
+	TLS,
+	Server,
+	Transfer *apm.Span
+}
+
+func (r *requestTracer) start(ctx context.Context) context.Context {
+	if r == nil {
```
axw, line 30: Can you please move the nil checks outside? That's how most of the Go agent code is written, and I would prefer to keep it consistent.

module/apmhttp/client.go
```diff
@@ -83,6 +84,7 @@ func (r *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	if tx == nil {
 		return r.r.RoundTrip(req)
 	}
+	ctx = r.requestTracer.start(ctx)
```
axw, line 87: Instead of creating spans as direct children of the transaction, we should make them children of the `external.http` span below. (And only if StartSpan returns a non-dropped span.)

module/apmhttp/client.go
```diff
@@ -114,6 +116,7 @@ func (r *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 
 	r.setHeaders(req, traceContext, propagateLegacyHeader)
 	resp, err := r.r.RoundTrip(req)
+	r.requestTracer.end()
```
axw, line 119: The response may not have been fully transferred at this point. I think we should end the requestTracer when we end the `external.http` span: either if RoundTrip returns an error (see below, line 122), or when the wrapped response body is consumed.

graphaelli:
Thanks for the review, i am still planning to follow up on this!

module/apmhttp/clienttrace.go
```diff
@@ -0,0 +1,97 @@
...
+
+		DNSDone: func(i httptrace.DNSDoneInfo) {
+			r.DNS.End()
+		},
+
+		ConnectStart: func(_, _ string) {
+			r.Connect = r.tx.StartSpan("Connect", "http.connect", nil)
+
+			if r.DNS == nil {
+				r.tx.Context.SetLabel("dns", false)
```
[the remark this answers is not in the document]
graphaelli: Intention was to quantify connections making (or not) unneeded frequent DNS requests

module/apmhttp/clienttrace.go
```diff
@@ -0,0 +1,97 @@
...
+			r.tx.Context.SetLabel("conn_reused", true)
+		},
+
+		WroteRequest: func(info httptrace.WroteRequestInfo) {
+			r.Server = r.tx.StartSpan("Server Processing", "http.server", nil)
+
+			// When connection is re-used, DNS/TCP/TLS hooks not called.
+			if r.Connect == nil {
+				// TODO
+			}
```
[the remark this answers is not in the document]
graphaelli: I forget, removed

graphaelli:
GH is not letting me convert this back to a draft - pushing up some commits now but still missing the multiple connection handling, will flag when really ready for another review.

module/apmhttp/clienttrace.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func (r *requestTracer) start(ctx context.Context) context.Context {
+	if r == nil {
+		return ctx
+	}
+	r.tx = apm.TransactionFromContext(ctx)
+	return httptrace.WithClientTrace(ctx, &httptrace.ClientTrace{
+		DNSStart: func(i httptrace.DNSStartInfo) {
+			r.DNS = r.tx.StartSpan("DNS Lookup", "http.dns", nil)
```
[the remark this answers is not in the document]
graphaelli: yes, done

module/apmhttp/clienttrace.go
```diff
@@ -0,0 +1,97 @@
...
+		},
+
+		DNSDone: func(i httptrace.DNSDoneInfo) {
+			r.DNS.End()
+		},
+
+		ConnectStart: func(_, _ string) {
+			r.Connect = r.tx.StartSpan("Connect", "http.connect", nil)
```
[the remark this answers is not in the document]
graphaelli: image

module/apmhttp/client_test.go
```diff
@@ -311,6 +312,23 @@ func TestWithClientRequestName(t *testing.T) {
...
+	server := httptest.NewServer(http.NotFoundHandler())
+	defer server.Close()
+
+	_, spans, _ := apmtest.WithTransaction(func(ctx context.Context) {
+		mustGET(ctx, server.URL, apmhttp.WithClientTrace())
+	})
+
+	require.Len(t, spans, 4)
+	assert.True(t, strings.HasPrefix(spans[0].Name, "Connect "),
+		fmt.Sprintf("want: Connect ... got: %s", spans[0].Name))
```
axw, line 325:
```suggestion
	assert.Regexp(t, "Connect .*", spans[0].Name)
```

axw: Or do an exact match with the server's listen address

module/apmhttp/clienttrace.go
```diff
@@ -0,0 +1,106 @@
...
+			r.Request.End()
+			r.Response = r.tx.StartSpan("Response", "http.response", parent)
+		},
+	})
+}
+
+func (r *requestTracer) end() {
+	if r == nil {
+		return
+	}
```
axw, line 102:
```suggestion
```

In keeping with the repo style
(Once we use `rt != nil` as a guard before calling `rt.end()`)

module/apmhttp/client.go
```diff
@@ -102,6 +103,10 @@ func (r *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 
 	name := r.requestName(req)
 	span := tx.StartSpan(name, "external.http", apm.SpanFromContext(ctx))
+	var rt requestTracer
```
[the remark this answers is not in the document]
graphaelli: Makes sense.  It seems weird for `newRequestTracer` to also return a `ctx`, thoughts on if/how to separate those?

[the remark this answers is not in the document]
axw: True. Maybe call it `ctx, clientTrace := withClientTrace(ctx, tx, span)`?

module/apmhttp/clienttrace.go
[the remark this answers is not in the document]
graphaelli: removed this label

module/apmhttp/clienttrace.go
```diff
@@ -0,0 +1,106 @@
...
+		},
+
+		DNSDone: func(i httptrace.DNSDoneInfo) {
+			r.DNS.End()
+		},
+
+		ConnectStart: func(network, addr string) {
+			span := r.tx.StartSpan(fmt.Sprintf("Connect %s", addr), "http.connect", parent)
+			if r.DNS == nil {
+				span.Context.SetLabel("dns", false)
```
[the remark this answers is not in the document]
graphaelli: We don't really need the labels for this, removed them.

graphaelli:
@axw I don't understand the go 1.9 test failure, this is ready for review otherwise.

module/apmhttp/clienttrace.go
```diff
@@ -0,0 +1,106 @@
...
+	return httptrace.WithClientTrace(ctx, &httptrace.ClientTrace{
+		DNSStart: func(i httptrace.DNSStartInfo) {
+			r.DNS = r.tx.StartSpan(fmt.Sprintf("DNS %s", i.Host), "http.dns", parent)
+		},
+
+		DNSDone: func(i httptrace.DNSDoneInfo) {
+			r.DNS.End()
+		},
+
+		ConnectStart: func(network, addr string) {
```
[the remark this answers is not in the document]
graphaelli: oh right, I forgot to do this

module/apmhttp/clienttrace.go
```diff
@@ -0,0 +1,97 @@
...
+		TLSHandshakeStart: func() {
+			r.TLS = tx.StartSpan("TLS", "http.tls", parent)
+		},
+
+		TLSHandshakeDone: func(_ tls.ConnectionState, _ error) {
+			r.TLS.End()
+		},
+
+		WroteRequest: func(info httptrace.WroteRequestInfo) {
+			r.Request = tx.StartSpan("Request", "http.request", parent)
```
axw, line 83: Isn't this when we want to _end_ the request span? I think we want to start it with GotConn

module/apmhttp/clienttrace.go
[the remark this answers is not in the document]
graphaelli: done now

module/apmhttp/client.go
```diff
@@ -178,6 +187,9 @@ func (b *responseBody) Read(p []byte) (n int, err error) {
 }
 
 func (b *responseBody) endSpan() {
+	if b.requestTracer != nil {
+		b.requestTracer.end()
+	}
```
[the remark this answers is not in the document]
graphaelli: I swear I considered doing that

module/apmhttp/clienttrace.go
graphaelli: You're right, this was not translated properly when converting it from `Server Processing`.  Looking at this again and seeing:

```
	// WroteRequest is called with the result of writing the
	// request and any body. It may be called multiple times
	// in the case of retried requests.
```

I think we should make multiple request spans in case of retries, so if `r.Request != nil r.Request.End()` - objections?

graphaelli: Cancel that, will test what happens in this case, not clear if gotconn is called again too

graphaelli: Now that the request span doesn't cover the server response time I'm missing that span representing the server time

image

I follow that a distributed trace into your own systems would show up there but this makes it non-trivial to monitor time in external services - any ideas here?  Maybe an option for `WithClientTrace` to create spans for that external time?  Is this not actually an issue?

axw: > I follow that a distributed trace into your own systems would show up there but this makes it non-trivial to monitor time in external services - any ideas here?

Fair point, I was really thinking about wholly-owned distributed systems.

Going back to https://w3c.github.io/resource-timing/#processing-model, "Request" is actually `responseStart - requestStart`. How about we do the same? `GotFirstResponseByte - GotConn`? If it's really interesting to know when the client finished writing the request, we could add "marks" like the RUM agent does.

module/apmhttp/clienttrace.go
```diff
@@ -0,0 +1,107 @@
...
+			r.TLS = tx.StartSpan("TLS", "http.tls", parent)
+		},
+
+		TLSHandshakeDone: func(_ tls.ConnectionState, _ error) {
+			r.TLS.End()
+		},
+
+		WroteRequest: func(info httptrace.WroteRequestInfo) {
+			r.Request.End()
+		},
```
axw, line 94:
```suggestion
```

replaced by below

axw:
Looks good! We can investigate the 1.9 failure separately... seems unrelated.

graphaelli:
woo, thanks for all of the guidance!

PR #789, opened 2020-07-22
docs: Add troubleshooting

bmorelli25:
## Summary

The Go Agent is the only APM Agent without a dedicated troubleshooting section. This PR adds a generic troubleshooting page with information on where users can get help, how to enable logging, and how to disable the Agent.

## SS

Screen Shot 2020-07-23 at 12 31 51 PM

## Related issues

For https://github.com/elastic/observability-docs/issues/35.


docs/troubleshooting.asciidoc
```diff
@@ -0,0 +1,29 @@
...
+
+In the unlikely event the Agent causes disruptions to a production application,
+you can disable the Agent while you troubleshoot.
+
+If you have access to <<dynamic-configuration,dynamic configuration>>,
+you can disable the recording of events by setting <<config-recording,`ELASTIC_APM_RECORDING`>> to `false`.
+When changed at runtime from a supported source, there's no need to redeploy your application for the change to take effect.
+
+If that doesn't work, or you don't have access to dynamic configuration, you can disable the Agent by setting
+<<config-active,`ELASTIC_APM_ACTIVE`>> to `false`.
```
bmorelli25, line 28: Does this approach make sense?

docs/troubleshooting.asciidoc
```diff
@@ -0,0 +1,29 @@
...
+You'll need to restart your application for the changes to apply.
```
bmorelli25, line 29: restart, redeploy, recompile -- what's the best verb?

axw: I think restart is fine, doesn't assume too much about how the application is being run. Definitely don't need to recompile.

docs/troubleshooting.asciidoc
axw: I think this is good.

docs/troubleshooting.asciidoc
```diff
@@ -0,0 +1,29 @@
...
+
+Is something not working as expected?
+Don't worry if you can't figure out what the problem is; we’re here to help!
+As a first step, ensure your app is compatible with the Agent's <<supported-tech,supported technologies>>.
+
+If you're an existing Elastic customer with a support contract, please create a ticket in the
+https://support.elastic.co/customers/s/login/[Elastic Support portal].
+Other users can post in the https://discuss.elastic.co/c/apm[APM discuss forum].
+
+IMPORTANT: *Please attach your debug logs* so that we can analyze the problem.
```
axw, line 12: I think we should call out here that the Go agent doesn't have any logging by default; you need to enable it by setting [ELASTIC_APM_LOG_FILE](https://www.elastic.co/guide/en/apm/agent/go/current/configuration.html#config-log-file). Perhaps add another section below?

bmorelli25: Added!

docs/troubleshooting.asciidoc
```diff
@@ -0,0 +1,57 @@
...
+
+With logging enabled, you can use <<config-log-level,`ELASTIC_APM_LOG_LEVEL`>> to increase the granularity of the agent's logging.
+For example: `ELASTIC_APM_LOG_FILE=debug`.
+
+Make sure to execute a few requests to your application before posting your log files.
+Each request should add lines similar to these in the logs:
+
+[source,log]
+----
+{"@timestamp":"2020-07-23T03:14:16Z","client-ip":"127.0.0.1","duration":31693200,"log.level":"info","message":"GET /api/orders (200)","method":"GET","path":"/api/orders","status":200,"trace.id":"9790410c7c9ff50ea1606094bf75946d","transaction.id":"9790410c7c9ff50e"}
```
axw, line 36: That log message isn't coming from the agent. Here's an example log:

```
{"level":"debug","time":"2020-07-23T11:46:32+08:00","message":"sent request with 100 transactions, 0 spans, 0 errors, 0 metricsets"}
```

bmorelli25: 🤦 Thanks as always, Andrew!

PR #794, opened 2020-08-05
Pin grpc dependency to v1.30.0 for Go 1.9

axw:


mdelapenya:
Thanks for the quick fix!

PR #795, opened 2020-08-05
chore(ci): separate retries into two separate steps

mdelapenya:
## What is this PR doing?
It separates the before_install and build scripts into two different retries on CI

## Why is it important?
With previous approach, if the second step (build) failed, then the before_install will be executed again, even if it already executed with success

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -280,10 +280,10 @@ def generateStep(version){
         withEnv(["GO_VERSION=${version}"]) {
           // Another retry in case there are any environmental issues
           // See https://issuetracker.google.com/issues/146072599 for more context
-          deleteDir()
-          unstash 'source'
           retry(3) {
             dir("${BASE_DIR}"){
+              deleteDir()
+              unstash 'source'
               sh script: './scripts/jenkins/before_install.sh', label: 'Install dependencies'
             }
           }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -280,11 +280,15 @@ def generateStep(version){
         withEnv(["GO_VERSION=${version}"]) {
           // Another retry in case there are any environmental issues
           // See https://issuetracker.google.com/issues/146072599 for more context
+          deleteDir()
+          unstash 'source'
           retry(3) {
-            deleteDir()
-            unstash 'source'
             dir("${BASE_DIR}"){
               sh script: './scripts/jenkins/before_install.sh', label: 'Install dependencies'
+            }
+          }
+          retry(3) {
+            dir("${BASE_DIR}"){
               sh script: './scripts/jenkins/build.sh', label: 'Build'
             }
           }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -280,11 +280,15 @@ def generateStep(version){
         withEnv(["GO_VERSION=${version}"]) {
           // Another retry in case there are any environmental issues
           // See https://issuetracker.google.com/issues/146072599 for more context
+          deleteDir()
+          unstash 'source'
```
axw, line 284: IIRC `before_install` will fail if you run it twice and it already fetched some stuff. That might be why the `deleteDir` is inside the retry block.

mdelapenya: Mmm, I see, let me move it back to the `before_install` step. Related to the test failures, which also reproduces in master:

```log
[2020-08-05T09:19:25.302Z] + make install precheck check-modules

[2020-08-05T09:19:25.562Z] go get -v -t ./...

[2020-08-05T09:19:26.542Z] google.golang.org/grpc/internal/credentials

[2020-08-05T09:19:26.542Z] # google.golang.org/grpc/internal/credentials

[2020-08-05T09:19:26.542Z] ../../google.golang.org/grpc/internal/credentials/spiffe.go:34:70: state.PeerCertificates[0].URIs undefined (type *x509.Certificate has no field or method URIs)

[2020-08-05T09:19:26.542Z] ../../google.golang.org/grpc/internal/credentials/spiffe.go:38:47: state.PeerCertificates[0].URIs undefined (type *x509.Certificate has no field or method URIs)

[2020-08-05T09:19:26.543Z] ../../google.golang.org/grpc/internal/credentials/spiffe.go:56:35: state.PeerCertificates[0].URIs undefined (type *x509.Certificate has no field or method URIs)

[2020-08-05T09:19:26.543Z] Makefile:44: recipe for target 'install' failed

[2020-08-05T09:19:26.543Z] make: *** [install] Error 2

script returned exit code 2
```

Jenkinsfile
```diff
@@ -198,11 +198,17 @@ pipeline {
...
+                dir("${BASE_DIR}"){
+                  sh script: './scripts/jenkins/before_install.sh', label: 'Install dependencies'
+                }
+              }
+              retry(3) {
+                dir("${BASE_DIR}"){
+                  sh script: './scripts/jenkins/build-test.sh', label: 'Build and test'
+                }
               }
             }
```
mdelapenya, line 213: This block is almost the same as the one produced `generateStep`, only differing in the labels for the node. @elastic/observablt-robots wdyt about refactoring this reusing the above step, parametrised by `version` and `nodeLabels`?

cachedout: You certainly could, but IMHO this is quite readable as-is and not much would be gained by a refactor.

mdelapenya: If it's readable then you have my sword :)

PR #739, opened 2020-03-30
ci(jenkins): add resilience when timeout issues

v1v:
### What

Split build-test.sh into two different scripts to retry if there are build errors when fetching the dependencies from github, see the below error:

```
[2020-03-27T14:35:15.622Z] Cloning into '/var/lib/jenkins/workspace/agent-go_apm-agent-go-mbp_master/src/github.com/prometheus/procfs'...
[2020-03-27T14:35:15.622Z] fatal: unable to access 'https://github.com/prometheus/procfs/': Failed to connect to github.com port 443: Connection timed out
[2020-03-27T14:35:15.622Z] package github.com/prometheus/procfs: exit status 128
```

### Why

Timeout issues are still a problem in the GCP, for such, we do use the retry/sleep approach. On the other hand, by decoupling the script in two different steps will help to retry the test step when there are genuine test failures, aka faster feedback.

PR #796, opened 2020-08-11
chore(ci): separate retries into two separate steps

mdelapenya:
Backports the following commits to 1.x:

- chore(ci): separate retries into two separate steps (#795)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -197,11 +197,17 @@ pipeline {
           }
           steps {
             withGithubNotify(context: 'Build-Test - OSX') {
-              deleteDir()
-              unstash 'source'
-              dir("${BASE_DIR}"){
-                sh script: './scripts/jenkins/before_install.sh', label: 'Install dependencies'
-                sh script: './scripts/jenkins/build-test.sh', label: 'Build and test'
+              retry(3) {
+                deleteDir()
+                unstash 'source'
+                dir("${BASE_DIR}"){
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -197,11 +197,17 @@ pipeline {
...
+                deleteDir()
+                unstash 'source'
+                dir("${BASE_DIR}"){
+                  sh script: './scripts/jenkins/before_install.sh', label: 'Install dependencies'
+                }
+              }
+              retry(3) {
+                dir("${BASE_DIR}"){
+                  sh script: './scripts/jenkins/build-test.sh', label: 'Build and test'
+                }
```
v1v, line 210: What's the reason of running the build/test scripts with a retry?

mdelapenya: I'm replicating the strategy for the linux workers, although linux scripts are decoupled into three different scripts. Maybe this is something that we can do here too for MacOS

mdelapenya:
Closing, as we do not need a backport to the 1.x branch

Jenkinsfile
v1v: I recall now:
- https://github.com/elastic/apm-agent-go/pull/739

It was split in three different scripts are you mentioned, so same behaviour for the MacOS, maybe we can refactor the common function to be reused for *nix?

mdelapenya: Like it!

PR #804, opened 2020-08-23
Record {transaction,sample}.sample_rate

axw:
Introduce the new ExtendedSampler interface, which
Samplers may implement to return the effective
sampling rate. This is implemented by the built-in
ratioSampler.

When starting a root transaction we now call the
ExtendedSampler method if implemented, and store
the effective sampling rate in the transaction's
tracestate under our "es" vendor key.

When receiving tracestate, we parse our "es" vendor
value and extract the sample rate.

When encoding transactions and spans we record the
sample rate (from tracestate) in the transaction and
span events.

### DIFF
--- a/sampler_test.go
+++ b/sampler_test.go
@@ -86,3 +86,23 @@ func TestRatioSamplerNever(t *testing.T) {
 		Span: apm.SpanID{255, 255, 255, 255, 255, 255, 255, 255},
 	}))
 }
+
+func TestRatioSamplerExtended(t *testing.T) {
+	s := apm.NewRatioSampler(0.5).(apm.ExtendedSampler)
+
+	result := s.SampleExtended(apm.SampleParams{
+		TraceContext: apm.TraceContext{Span: apm.SpanID{255, 0, 0, 0, 0, 0, 0, 0}},
+	})
+	assert.Equal(t, apm.SampleResult{
+		Sampled:    false,
+		SampleRate: 0.5,
+	}, result)
+
+	result = s.SampleExtended(apm.SampleParams{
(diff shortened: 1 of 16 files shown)


utils_go9.go
```diff
@@ -0,0 +1,33 @@
...
+
+// +build !go1.10
+
+package apm
+
+import "math"
+
+// Implementation of math.Round for Go < 1.10.
+//
+// Code shamelessly copied from pkg/math.
```
mikker, line 26: YOLO 😉

sampler_test.go
```diff
@@ -86,3 +86,23 @@ func TestRatioSamplerNever(t *testing.T) {
...
+
+func TestRatioSamplerExtended(t *testing.T) {
+	s := apm.NewRatioSampler(0.5).(apm.ExtendedSampler)
+
+	result := s.SampleExtended(apm.SampleParams{
+		TraceContext: apm.TraceContext{Span: apm.SpanID{255, 0, 0, 0, 0, 0, 0, 0}},
+	})
+	assert.Equal(t, apm.SampleResult{
+		Sampled:    false,
+		SampleRate: 0.5,
```
jalvz, line 98: Based on the spec, this result is wrong, right? I was expecting to find the special case https://github.com/elastic/apm-agent-go/pull/804/files#diff-ade132f78e2d3e9afadc8eabd0ac434bR105 inside `ExtendedSample` somehow

jalvz (approved):

sampler_test.go
axw: The special case is handled elsewhere (transaction.go), to keep the logic in one place, rather than in every implementation of ExtendedSampler.

axw: Having said that, I think we're missing a test where the sample rate is > 0, sampling decision is false, and es=s:0. I'll add one.

PR #813, opened 2020-09-02
group httptrace spans as external.http

graphaelli:
To align with existing HTTP spans.  Compare and old and new breakdowns in the UI as a result:

image


axw (approved):
LGTM! We should write some tests for this at some point.

graphaelli:
We have one test - https://github.com/elastic/apm-agent-go/pull/788/files#diff-28a415f0b4cd73c220a8624ecc44b179R315-R328, agreed on adding more.  I'd love to see what one for the dual stack connection spans looks like.

axw:
> We have one test

Ah right, sorry. I was just expecting to see tests changing as a result of changing the span type.

Issue #799, opened 2020-08-12
Feature: {Transaction,Span}.Outcome


axw:
Add {Transaction,Span}.Outcome fields, which accept one of three possible values: "success", "failure", "unknown". Instrumentation modules should set the Outcome, but if unspecified the server will default to setting these to "unknown".

See https://github.com/elastic/apm/pull/299

PR #820, opened 2020-09-24
Outcome

axw:
Add the Outcome field to Transactions and Spans.

If when a transaction or span is ended, its outcome has not been specified, we either use some protocol/instrumentation-specific logic to set a default, or otherwise set it to "unknown". For HTTP and gRPC, we use the status code to set the outcome.

- HTTP server side: "success" for status codes < 500, "failure" otherwise.
- HTTP client side: "success" for status codes < 400, "failure" otherwise.
- gRPC server side:
   - "unknown" for status code Unknown
   - "failure" for status codes DataLoss, Unavailable, Internal, and Unimplemented
   - "success" otherwise
- gRPC client side:
   - "unknown" for status code Unknown
   - "success" for status code OK
   - "failure" otherwise

Closes https://github.com/elastic/apm-agent-go/issues/799


simitt (approved):
LGTM

PR #823, opened 2020-09-30
Collect cloud metadata

### LINKED ISSUE #786: Collect cloud metadata
See elastic/apm#256

Reference implementation: https://github.com/elastic/apm-agent-python/issues/822

axw:
Add support for fetching cloud metadata for AWS, Azure, and GCP.
Based on the Python reference implementation: https://github.com/elastic/apm-agent-python/pull/826

By default we attempt to sniff metadata for all cloud providers. This
can be overridden by setting the environment variable
`ELASTIC_APM_CLOUD_PROVIDER` to one of:
"none", "aws", "azure", or "gcp".

We set a short (100ms) socket connect timeout, and a slightly longer
overall timeout (1s) for fetching the cloud metadata.

In tests we disable cloud metadata fetching by default. There are
currently no functional tests, as that would rely on us running in a
known cloud environment.

The PR has been manually tested in Google Compute Engine and
Google Cloud Run. Otherwise we're relying on the test data from
the Python agent.

Closes #786

### DIFF
--- a/internal/apmcloudutil/provider.go
+++ b/internal/apmcloudutil/provider.go
@@ -1,0 +1,116 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package apmcloudutil
+
+import (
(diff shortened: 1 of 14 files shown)


internal/apmcloudutil/provider.go
```diff
@@ -0,0 +1,116 @@
...
+// requests do not block normal operation in non-cloud environments.
+func (p Provider) GetCloudMetadata(ctx context.Context, logger apmlog.Logger, out *model.Cloud) bool {
+	return p.getCloudMetadata(ctx, defaultClient, logger, out)
+}
+
+func (p Provider) getCloudMetadata(ctx context.Context, client *http.Client, logger apmlog.Logger, out *model.Cloud) bool {
+	if p == None {
+		return false
+	}
+	for _, provider := range []Provider{AWS, Azure, GCP} {
```
jalvz, line 92: Im confused by this logic, why do you need to iterate over all providers? can't you just `switch` on `p` value?

jalvz (approved):
code looks great!

internal/apmcloudutil/provider.go
axw: This is for the "auto" case: in that case we loop around if one fails.

jalvz: got it

axw: good call, will add a comment

PR #825, opened 2020-10-04
Support gormio

### LINKED ISSUE #810: Add instrumentation for gorm v2
Gorm v2 has significant performance improvements and features.

More notably it has context support by default.

https://gorm.io/docs/v2_release_note.html#Major-Features

Deepak13245:
Resolves #810


axw:
@Deepak13245 thanks for the PR!

Looks pretty good, but I'm wondering if the callbacks are still necessary. If context is now first class, then perhaps we can instead just rely on `apmsql` to create the spans?

Also, I'd like to call this module `apmgormv2` to be more in line with other modules.

Deepak13245:
@axw renamed to apmgormv2
regarding the creation of spans by apmsql, gorm doesn't accept *sql.DB directly in params. There is a workaround so I raised a separate PR https://github.com/Deepak13245/apm-agent-go/pull/1. Please suggest if there is a better approach.

Thanks

axw:
A couple of things bother me about the plugin approach:
 - It closes and reopens the connection. Not the end of the world, just a bit annoying.
 - As it is at the moment, extractDsn forces a dependency on all the drivers. We could solve this by having the "_" dialect imports register a function though.

What do you think about providing "Open" functions inside the dialect packages instead? So instead of

```go
import (
    "gorm.io/driver/sqlite"
    "gorm.io/gorm"         
)

db, err := gorm.Open(sqlite.Open("gorm.db"), &gorm.Config{})
```

you could write

```go
import (
    "go.elastic.co/apm/module/apmgormv2/dialect/apmsqlite"
    "gorm.io/gorm"
)

db, err := gorm.Open(apmsqlite.Open("gorm.db"), &gorm.Config{})
```

Internally this would use the recently introduced support for custom drivers: https://github.com/go-gorm/sqlite/pull/11

Basically just:

```go
package apmsqlite // go.elastic.co/module/apmgormv2/dialect/apmsqlite

import (
    "gorm.io/driver/sqlite"
    "gorm.io/gorm"
    "go.elastic.co/module/apm/module/apmsql"
    _ "go.elastic.co/module/apm/module/apmsql/sqlite3"
)

func Open(dsn string) gorm.Dialector {
        return &sqlite.Dialector{
                DriverName: apmsql.DriverPrefix + sqlite.DriverName,
                DSN:        dsn,
        }
}
```

Deepak13245:
@axw creating dialect from with driver prefix makes it neat, updated the pr please check.

module/apmgormv2/driver/sqlite/init.go
```diff
@@ -0,0 +1,39 @@
...
+	_ "go.elastic.co/apm/module/apmsql/sqlite3" // register sqlite3 with apmsql
+)
+
+// Open creates a dialect with apmsql
+func Open(dsn string) gorm.Dialector {
+	driverName := sqlite.Dialector{}.Name()
+
+	dialect := &sqlite.Dialector{
+		DSN: dsn,
+		DriverName: driverName,
```
axw, line 35:
```suggestion
		DriverName: apmsql.DriverPrefix + driverName,
```

module/apmgormv2/driver/postgres/init.go
```diff
@@ -0,0 +1,42 @@
...
+// Package apmpostgres imports the gorm postgres dialect package,
+// and also registers the lib/pq driver with apmsql.
+package apmgpostgres
+
+import (
+	"go.elastic.co/apm/module/apmsql"
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+
+	_ "go.elastic.co/apm/module/apmsql/pq" // register lib/pq with apmsql
```
axw, line 27: Apparently gorm v2 uses pgx by default: https://github.com/go-gorm/postgres/blob/master/postgres.go

Maybe don't add the postgres dialect for now? WDYT?

axw (commented):
Thanks! I agree this looks much neater.

axw:
jenkins run the tests please

module/apmgormv2/driver/postgres/init.go
Deepak13245: Yes, Removing Postgres dialect for now

axw (approved):
Perfect, thanks!

axw:
jenkins run the tests please

axw:
> [2020-10-07T02:40:59.696Z] ../../gorm.io/gorm/utils/utils.go:46:30: reflect.ValueOf(val).IsZero undefined (type reflect.Value has no field or method IsZero)

It looks like gorm v2 requires Go 1.13+, so we'll need to update the build constraint.

axw:
Looks good, I think the remaining test failures are fixed on master. Can you please merge master so we can make sure?

module/apmgormv2/doc.go
```diff
@@ -0,0 +1,21 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+// +build go1.14
+
```
axw, line 19:
```suggestion
```

One last thing I hope. This is a bit of a hack to get `check-vet` to stop failing for older versions of Go.

axw:
CI looks happy. Thanks for another nice PR, and for bearing with my requests :)

Issue #809, opened 2020-08-27
Optimise StartTransaction when the tracer is disabled


jalvz:
As title.

Accidentally calling `StartTransaction` when the tracer is disabled incurs some marginal overhead that could be avoided by returning early. This should help in memory-constrained environments and align with user's expectations

newly12:
Thanks @jalvz for addressing this!

PR #827, opened 2020-10-05
Start cheap transactions when the tracer is not recording

jalvz:
Surely I must be missing something, not sure what :)

Closes https://github.com/elastic/apm-agent-go/issues/809

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -57,7 +57,7 @@ func (t *Tracer) StartTransactionOptions(name, transactionType string, opts Tran
 	// transaction.
 	instrumentationConfig := t.instrumentationConfig()
 	tx.recording = instrumentationConfig.recording
-	if !t.Recording() {
+	if !tx.recording || !t.Active() {
 		return tx
 	}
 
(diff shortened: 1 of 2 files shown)


transaction.go
```diff
@@ -57,7 +57,7 @@ func (t *Tracer) StartTransactionOptions(name, transactionType string, opts Tran
 	// transaction.
 	instrumentationConfig := t.instrumentationConfig()
 	tx.recording = instrumentationConfig.recording
-	if !tx.recording {
+	if !t.Recording() {
```
axw, line 60:
```suggestion
	if !tx.recording || !t.Active() {
```

`t.Recording()` internally calls `t.instrumentationConfig()` to fetch a config snapshot, which we've already done above. Using a single config snapshot for all transaction behaviour makes it easier to reason about.

axw (changes requested):
Good catch! Tests are failing for unrelated reason, looks like we need to update version pinning again (in another PR).

axw (approved):

PR #831, opened 2020-10-06
support pgx v4 driver

Deepak13245:


internal/sqlutil/drivername.go
```diff
@@ -41,6 +41,10 @@ func DriverName(d driver.Driver) string {
 		if strings.HasSuffix(t.PkgPath(), "github.com/lib/pq") {
 			return "postgresql"
 		}
+
+		if strings.HasSuffix(t.PkgPath(), "github.com/jackc/pgx/v4/stdlib") {
```
axw, line 45:
```suggestion
		if strings.HasSuffix(t.PkgPath(), "github.com/jackc/pgx/") {
```

I think that would be safe and future proof, WDYT?

axw (changes requested):
Very nice, thank you! We just need to maintain API compatibility (see comment about apmpq.ParseDSN), otherwise this is good to merge.

axw:
jenkins run the tests please

internal/pgutil/parser_test.go
```diff
@@ -15,16 +15,16 @@
...
-package apmpq_test
+package pgutil_test
 
 import (
 	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 
+	apmpq "go.elastic.co/apm/internal/pgutil"
```
axw, line 26:
```suggestion
	"go.elastic.co/apm/internal/pgutil"
```

(and update references)

module/apmsql/pgxv4/init.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"github.com/jackc/pgx/v4/stdlib"
+
+	"go.elastic.co/apm/internal/pgutil"
+
+	"go.elastic.co/apm/module/apmsql"
+)
+
+func init() {
+	apmsql.Register("pgxv4", &stdlib.Driver{}, apmsql.WithDSNParser(pgutil.ParseDSN))
```
axw, line 31:
```suggestion
	apmsql.Register("pgx", &stdlib.Driver{}, apmsql.WithDSNParser(pgutil.ParseDSN))
```

(needs to match the underlying driver name)

axw (approved):
A couple more small things, otherwise LGTM.

axw:
`internal/pgutil` will need to be moved to `module/apmsql/internal/pgutil`, otherwise `go.elastic.co/apm` has a dependency on `apmsql` which we do not want.

axw:
Nice work, thanks very much @Deepak13245!

PR #834, opened 2020-10-12
Update sample rate precision handling

axw:
We already had sampling rate rounding, but it
was using the older proposal of 3 decimal places,
and did not clamp the lower bound.

Updated to use the same logic in both the ratio
sampler implementation, and the sampler
implementation independent interpretation of the
sample rate.

Note that we intentionally round in two places:
once in the sampler implementation for the
initial sampling decision, and once in the code
that injects the rate into trace context to guarantee
an upper bound for tracestate.

Also, use the "round" function instead
of math.Round directly, to support older
versions of Go.

### DIFF
--- a/scripts/before_install.sh
+++ b/scripts/before_install.sh
@@ -31,6 +31,7 @@ if (! go run scripts/mingoversion.go 1.11 &>/dev/null); then
   pin github.com/elastic/go-sysinfo v1.3.0
   pin google.golang.org/grpc v1.30.0 https://github.com/grpc/grpc-go
   pin github.com/jinzhu/gorm v1.9.16
+  pin github.com/ugorji/go v1.1.10
 fi
 
 if (! go run scripts/mingoversion.go 1.10 &>/dev/null); then
--- a/transaction.go
+++ b/transaction.go
@@ -110,7 +110,7 @@ func (t *Tracer) StartTransactionOptions(name, transactionType string, opts Tran
 				// we will scale the sampled transactions.
 				result.SampleRate = 0
 			}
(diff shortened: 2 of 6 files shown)


scripts/before_install.sh
```diff
@@ -31,6 +31,7 @@ if (! go run scripts/mingoversion.go 1.11 &>/dev/null); then
   pin github.com/elastic/go-sysinfo v1.3.0
   pin google.golang.org/grpc v1.30.0 https://github.com/grpc/grpc-go
   pin github.com/jinzhu/gorm v1.9.16
+  pin github.com/ugorji/go v1.1.10
```
simitt, line 34: how is this related to the `round` changes?

transaction.go
```diff
@@ -110,7 +110,7 @@ func (t *Tracer) StartTransactionOptions(name, transactionType string, opts Tran
 				// we will scale the sampled transactions.
 				result.SampleRate = 0
 			}
-			sampleRate := round(1000*result.SampleRate) / 1000
+			sampleRate := roundSampleRate(result.SampleRate)
```
jalvz, line 113: ah, hadn't noticed this!

jalvz (approved):

simitt (approved):

scripts/before_install.sh
axw: It's not. I was lazy and fixed an unrelated CI failure in the same PR.

transaction.go
axw: No worries, I forgot about it when reviewing your PR :smile:

PR #836, opened 2020-10-20
Transaction.Discard sets TransactionData to nil

### LINKED ISSUE #835: tx.Discard() should set tx.TransactionData to nil as tx.End()
**Describe the bug**
Before [this change](https://github.com/elastic/apm-agent-go/pull/478/commits/d4ce90ae56156a6bcb527e3b97816c0ea9adc1b0#diff-63ab601287b8b9c040760fe0bdd288f55b73f37cd7e4f1e519bea2bd43a18bbaL211),  `tx.Discard()` used to clear `tx.TransactionData` just as `tx.End()`. This makes it safe to call `Discard()` multiple times and/or mix with `End()` calls. But now calling `End()` after `Discard()` leads to panic (under some workload), and since internal `tx.ended` would report false after `Discard()`, this may also lead to other unexpected behaviors.

**To Reproduce**

```go
package main

import (
	"fmt"
	"sync"

	"go.elastic.co/apm/transport/transporttest"
)

func main() {
	fmt.Println("Hello, playground")

	tracer, _ := transporttest.NewRecorderTracer()
	defer tracer.Close()

	wg := new(sync.WaitGroup)
	for i := 0; i < 100000; i++ { // increase this if your machine is fast
		wg.Add(1)
		go func() {
			tx := tracer.StartTransaction("name", "type")
			tx.Discard()
			tx.End()
			wg.Done()
		}()
	}
	wg.Wait()
}
```

**Expected behavior**

Call `Discard` should clear `tx.TransactionData`

axw:
Fix Transaction.Discard so that it sets TransactionData
to nil, like its documentation says. This is necessary
to avoid adding a transaction to the `sync.Pool` multiple
times (e.g. by calling End after after it has been discarded.)

Closes #835

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -35,6 +35,7 @@ https://github.com/elastic/apm-agent-go/compare/v1.8.0...master[View commits]
 - Add cloud metadata, configurable with ELASTIC_APM_CLOUD_PROVIDER {pull}823[#(823)]
 - Round ELASTIC_APM_SAMPLING_RATE with 4 digits precision {pull}828[#(828)]
 - module/apmhttp: implement io.ReaderFrom in wrapped http.ResponseWriter {pull}830[#(830)]
+- Fixed Transaction.Discard so that it sets TransactionData to nil {pull}836[#(836)]
 
 [[release-notes-1.x]]
 === Go Agent version 1.x
--- a/transaction.go
+++ b/transaction.go
@@ -246,6 +246,7 @@ func (tx *Transaction) Discard() {
 		return
 	}
 	tx.reset(tx.tracer)
+	tx.TransactionData = nil
 }
 
 // End enqueues tx for sending to the Elastic APM server.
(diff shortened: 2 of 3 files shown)


jalvz (approved):

CHANGELOG.asciidoc
```diff
@@ -35,6 +35,7 @@ https://github.com/elastic/apm-agent-go/compare/v1.8.0...master[View commits]
 - Add cloud metadata, configurable with ELASTIC_APM_CLOUD_PROVIDER {pull}823[#(823)]
 - Round ELASTIC_APM_SAMPLING_RATE with 4 digits precision {pull}828[#(828)]
 - module/apmhttp: implement io.ReaderFrom in wrapped http.ResponseWriter {pull}830[#(830)]
+- Fixed Transaction.Discard so that it sets TransactionData to nil {pull}836[#(836)]
```
jalvz, line 38: shouldn't this be under "bug fixes"?

axw: The Go agent changelog has not, to date, separated additions from fixes. There's a template comment above with a "bug fixes" section, but it's not used anywhere. Maybe it should be...

I'd like to defer that to when we release, and look at what the other agents are doing.

Issue #843, opened 2020-11-04
Let applications enable support for follows-from spans


randomswdev:
In the opentracing moduke, the file [context.go](https://github.com/elastic/apm-agent-go/blob/v1.9.0/module/apmot/context.go), contains logig to consider valid only the spans generated by a child-of ref.
A comment in the file explains that a unit test is able to override this behaviour.

Is there any plan to overcome this limitation, that creates troubles when using the tracing module with third party libraries?

Or at least do you think it would be possible to let applications do th same overriding the unit testing does?

In this latter case it is up to the application developer to evaluate if it is safe to override the logic that checks which spans are "valid".

axw:
Hi @randomswdev, thanks for opening the issue!

Right now we don't have any concrete plans to accept follows-from references. We started discussing it a while back (https://github.com/elastic/apm/issues/122), but that has stalled.

I would be open to adding a tracer option that overrides the default behaviour, but as you say it would be up to the developer to make sure that it's safe to do so, and any semantic issues that arise are acceptable to them. Are you interested in submitting a PR that adds an option?

axw:
Closed by #847

PR #847, opened 2020-11-09
Allow overriding the span ref validation

randomswdev:
This pull request is a candidate for solving issue https://github.com/elastic/apm-agent-go/issues/843

### DIFF
--- a/module/apmot/tracer.go
+++ b/module/apmot/tracer.go
@@ -35,7 +35,10 @@ import (
 // By default, the returned tracer will use apm.DefaultTracer.
 // This can be overridden by using a WithTracer option.
 func New(opts ...Option) opentracing.Tracer {
-	t := &otTracer{tracer: apm.DefaultTracer}
+	t := &otTracer{
+		tracer:         apm.DefaultTracer,
+		isValidSpanRef: isChildOfSpanRef,
+	}
 	for _, opt := range opts {
 		opt(t)
 	}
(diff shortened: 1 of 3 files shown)


module/apmot/tracer.go
```diff
@@ -34,8 +34,14 @@ import (
 //
 // By default, the returned tracer will use apm.DefaultTracer.
 // This can be overridden by using a WithTracer option.
+// The option WithSpanRefValidator allows to override the
+// set of spans that are traced. By default only child-of
+// spans are traced.
```
axw, line 39:
```suggestion
// The option WithSpanRefValidator allows one to override the
// set of spans that are recorded. By default only child-of
// spans are recorded.
```

module/apmot/tracer.go
```diff
@@ -197,6 +204,20 @@ func WithTracer(t *apm.Tracer) Option {
 	}
 }
 
+// SpanRefValidator verifies if a span is valid and can be processed.
```
axw, line 207:
```suggestion
// SpanRefValidator verifies if a span is valid and should be recorded.
```

axw (approved):
Thank you, this looks great! Just a couple of minor requested doc comment changes.

axw:
jenkins run the tests please

module/apmot/tracer.go
randomswdev: Changed in my last commit

module/apmot/tracer.go
randomswdev: Changed in my last commit

axw:
Thanks for your contribution @randomswdev, LGTM!

PR #852, opened 2020-11-18
docs: clean up conditional code

bmorelli25:
This PR cleans up the conditional code added in https://github.com/elastic/apm-agent-go/pull/839 for the `7.10` stack release.


axw (approved):
Does this change need to be on the 1.x branch too?

bmorelli25:
> Does this change need to be on the 1.x branch too?

Not immediately. I think it's fine to leave the `1.x` docs as-is until they're overwritten during the next release.

PR #858, opened 2020-12-10
Update tests and doc comments to use SERVER_URL

### LINKED ISSUE #850: [META 359] Consolidate server_url/server_urls option
Rename `server_urls` to `server_url` (see https://github.com/elastic/apm/issues/359) in a backwards compatible way.

Adapt documentation to only describe the `server_url` option.

axw:
The Go Agent will continue to support SERVER_URLS,
but we will de-epmphasise its use in favour of using
load balancer proxies with agents generally. We never
actually documented SERVER_URLS in the user docs,
so nothing to do there.

Closes #850

### DIFF
--- a/transport/http.go
+++ b/transport/http.go
@@ -91,11 +91,9 @@ type HTTPTransport struct {
 // streaming data to the APM Server. The returned HTTPTransport will be
 // initialized using the following environment variables:
 //
-// - ELASTIC_APM_SERVER_URLS: a comma-separated list of APM Server URLs.
-//   The transport will use this list of URLs for sending requests,
-//   switching to the next URL in the list upon error. The list will be
-//   shuffled first. If no URLs are specified, then the transport will
-//   use the default URL "http://localhost:8200".
+// - ELASTIC_APM_SERVER_URL: the APM Server URL used for sending
+//   requests. If no URL is specified, then the transport will use the
+//   default URL "http://localhost:8200".
 //
 // - ELASTIC_APM_SERVER_TIMEOUT: timeout for requests to the APM Server.
 //   If not specified, defaults to 30 seconds.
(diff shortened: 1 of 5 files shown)


transport/http.go
```diff
@@ -91,11 +91,9 @@ type HTTPTransport struct {
...
 // initialized using the following environment variables:
 //
-// - ELASTIC_APM_SERVER_URLS: a comma-separated list of APM Server URLs.
-//   The transport will use this list of URLs for sending requests,
-//   switching to the next URL in the list upon error. The list will be
-//   shuffled first. If no URLs are specified, then the transport will
-//   use the default URL "http://localhost:8200".
+// - ELASTIC_APM_SERVER_URL: the APM Server URL used for sending
+//   requests. If no URL is specified, then the transport will use the
+//   default URL "http://localhost:8200".
```
simitt, line 96: Just a minor thing, but since `ELASTIC_APM_SERVER_URLS` is deprecated, should we also change the order of how the ENV vars are parsed, so that `ELASTIC_APM_SERVER_URL` is always used if provided.  https://github.com/elastic/apm-agent-go/blob/8b1a0fe81cca0549e5b2013d209f2dcdd34a9aa0/transport/http.go#L528-L538.

simitt (approved):

transport/http.go
axw: I did think about doing that, and decided at the time to make it a purely non-functional change. I don't think it's likely to matter either way, but I'd like to avoid a breaking change unless necessary.

axw: I assume since you approved that you're good either way, so I'll merge. If you disagree with my decision, let me know and we can come back to it.

simitt: yes, I am fine either way - thanks for the explanation.

PR #859, opened 2020-12-10
Make `log_level` centrally configurable

### LINKED ISSUE #829: [META 321] Add log_level spec
Implementing elastic/apm#332

Note that the implementation of this spec also includes adding the option to [Kibana](https://github.com/elastic/kibana/blob/master/x-pack/plugins/apm/common/agent_configuration/setting_definitions/general_settings.ts).

axw:
`$ELASTIC_APM_LOG_LEVEL` now defines the local value, and this can be overridden via central config. If central config is updated and then removed, we'll revert back to the local value like with other config.

If a custom logger is defined, or no log file is specified, the `log_level` attribute is ignored regardless of whether central config is used. If a custom logger is defined we will log a warning to it about the central config change being ineffective.

Closes #829

### DIFF
--- a/config_test.go
+++ b/config_test.go
@@ -20,10 +20,12 @@ package apm_test
 import (
 	"context"
 	"encoding/json"
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"testing"
@@ -36,6 +38,7 @@ import (
 	"go.elastic.co/apm"
 	"go.elastic.co/apm/apmconfig"
 	"go.elastic.co/apm/apmtest"
+	"go.elastic.co/apm/internal/apmlog"
 	"go.elastic.co/apm/transport"
 	"go.elastic.co/apm/transport/transporttest"
 )
@@ -44,7 +47,7 @@ func TestTracerCentralConfigUpdate(t *testing.T) {
 	run := func(configKey, configValue string, isRemote func(*apmtest.RecordingTracer) bool) {
 		t.Run(configKey, func(t *testing.T) {
 			response, _ := json.Marshal(map[string]string{configKey: configValue})
-			testTracerCentralConfigUpdate(t, string(response), isRemote)
+			testTracerCentralConfigUpdate(t, apmtest.NewTestLogger(t), string(response), isRemote)
 		})
 	}
 	run("transaction_sample_rate", "0", func(tracer *apmtest.RecordingTracer) bool {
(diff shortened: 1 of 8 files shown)


internal/apmlog/logger.go
```diff
@@ -65,85 +80,98 @@ func initDefaultLogger() {
...
+	case DebugLevel:
 		return "debug"
-	case infoLevel:
+	case InfoLevel:
 		return "info"
-	case warnLevel:
+	case WarnLevel:
 		return "warn"
-	case errorLevel:
+	case ErrorLevel:
```
simitt, line 116: `CriticalLevel` is missing

tracer.go
```diff
@@ -426,6 +426,14 @@ func newTracer(opts TracerOptions) *Tracer {
 	t.setLocalInstrumentationConfig(envSanitizeFieldNames, func(cfg *instrumentationConfigValues) {
 		cfg.sanitizedFieldNames = opts.sanitizedFieldNames
 	})
+	if apmlog.DefaultLogger != nil {
+		defaultLogLevel := apmlog.DefaultLogger.Level()
```
simitt, line 430: Hasn't this already potentially changed from the local ENV to the remote config? I don't see how this falls back to the original value in case a remote config value was provided but isn't any longer.

config_test.go
```diff
@@ -97,9 +100,31 @@ func TestTracerCentralConfigUpdate(t *testing.T) {
 		assert.Len(t, payloads.Transactions[0].Context.Request.Cookies, 1)
 		return payloads.Transactions[0].Context.Request.Cookies[0].Value == "[REDACTED]"
 	})
+	t.Run("log_level", func(t *testing.T) {
```
simitt, line 103: I might be missing something - but are there any tests for actually applying the remote log level and also falling back to the local config if a remote config is no longer provided?

internal/apmlog/logger.go
axw: Nice catch, thanks!

tracer.go
axw: Not 100% sure I follow your question.
- `apmlog.DefaultLogger.Level()` is called when constructing the Tracer, so there's no remote config yet received. We're just capturing the initial (default/environment variable) value.
- the function literal passed into `t.setLocalInstrumentationConfig` below will be invoked immediately (having no effect, since it's just setting the level to the same value), and also when remote config is set and later removed. The latter bit is taken care of by https://github.com/elastic/apm-agent-go/blob/879c34a347e6486e7cd1767b9e6e3659e23283b6/config.go#L407-L417

config_test.go
axw: `testTracerCentralConfigUpdate` (which this test calls) does that in https://github.com/elastic/apm-agent-go/blob/879c34a347e6486e7cd1767b9e6e3659e23283b6/config_test.go#L142-L165.

Each of these sub-tests defines a config key, value to set remotely, and a predicate that indicates whether the remote config has been applied. The code block I pointed to first checks that with the initial tracer config the predicate returns false; then sets the remote config and checks the predicate returns true; then clears out all remote config and again checks that the predicate returns false.

tracer.go
simitt: I missed `defaultLogLevel := apmlog.DefaultLogger.Level()`
in https://github.com/elastic/apm-agent-go/pull/859/files#diff-4e84387cbc4fa01c5f9e572c02575e6c8536608159891072d2fbfa69af83a86fR430 - all good then, thanks.

simitt (approved):

internal/apmlog/logger.go
```diff
@@ -65,85 +80,98 @@ func initDefaultLogger() {
...
+// ParseLogLevel parses s as a log level.
+func ParseLogLevel(s string) (Level, error) {
 	switch strings.ToLower(s) {
 	case "debug":
-		return debugLevel, nil
+		return DebugLevel, nil
 	case "info":
-		return infoLevel, nil
+		return InfoLevel, nil
 	case "warn":
```
trentm, line 129: (This is my first time looking at the go agent, and really reviewing much Go code at all, so feel free to ignore if I'm missing something obvious.)

The [spec](https://github.com/elastic/apm/blob/master/specs/agents/logging.md) uses "warning" rather than "warn". Is this function parsing level strings from central config?

axw: @trentm Thanks, nice pick up! You're quite right, I missed that. I'll open another PR to fix it.

Issue #814, opened 2020-09-04
TRANSACTION_IGNORE_URLS should be dynamically configurable


axw:
`TRANSACTION_IGNORE_URLS` config should be dynamically configurable via central config. We currently only support configure this via an environment variable, in the apmhttp package. We'll need move the configuration over to the tracer.

See https://github.com/elastic/apm/pull/333

axw:
We need to add `go` to the `includeAgents` list in Kibana: https://github.com/elastic/kibana/blob/053628cae862103b8c27bc60656bf45d76dc3f5d/x-pack/plugins/apm/common/agent_configuration/setting_definitions/general_settings.ts#L241-L258

Reopening this issue so we don't forget.

PR #872, opened 2021-01-19
Add ignore_urls to central config

jalvz:
Closes https://github.com/elastic/apm-agent-go/issues/814


module/apmhttp/ignorer.go
```diff
@@ -53,6 +54,16 @@ func DefaultServerRequestIgnorer() RequestIgnorerFunc {
 	return defaultServerRequestIgnorer
 }
 
+// DynamicServerRequestIgnorer returns the RequestIgnorer to use in
```
axw, line 57: I think we should add a deprecation comment to DefaultServerRequestIgnorer above, seeing as it confusingly will no longer name have an apt name "Default..."

module/apmhttp/ignorer.go
```diff
@@ -53,6 +54,16 @@ func DefaultServerRequestIgnorer() RequestIgnorerFunc {
...
+// handlers. The list of wildcard patterns comes from central config
+func DynamicServerRequestIgnorer(t *apm.Tracer) RequestIgnorerFunc {
+	matchers := t.GetIgnoreURLs()
+	if len(matchers) != 0 {
+		defaultServerRequestIgnorer = NewWildcardPatternsRequestIgnorer(matchers)
+	}
+	return defaultServerRequestIgnorer
```
axw, line 64:
```suggestion
	return func(r *http.Request) bool {
		matchers := t.GetIgnoreURLs()
		return matchers.MatchAny(r.URL.String())
	}
```

For it to be dynamic we'll need to fetch the matchers on each call, not once at construction time.

module/apmhttp/handler.go
```diff
@@ -42,7 +42,7 @@ func Wrap(h http.Handler, o ...ServerOption) http.Handler {
 		handler:        h,
 		tracer:         apm.DefaultTracer,
 		requestName:    ServerRequestName,
-		requestIgnorer: DefaultServerRequestIgnorer(),
+		requestIgnorer: DynamicServerRequestIgnorer(apm.DefaultTracer),
```
axw, line 45: We'll need to set the default requestIgnorer after running the option functions (and only if a requestIgnorer is not supplied through WithRequestIgnorer), in case a non-default tracer is supplied with WithTracer.

axw (commented):
Looking good!

A couple of additional bits are needed:
 - In `newTracer`, another call to `t.setLocalInstrumentationConfig` is needed to set the initial local config. This is required so if remote config is set, and then later unset, the tracer will revert to the local config.
 - In `config_test.go`, add a subtest to TestTracerCentralConfigUpdate for "transaction_ignore_urls"

config.go
```diff
@@ -480,6 +495,11 @@ func (t *Tracer) updateInstrumentationConfig(f func(cfg *instrumentationConfig))
 	}
 }
 
+// IgnoredTransactionURLMatchers returns the ignored transaction URLs
+func (t *Tracer) IgnoredTransactionURLMatchers() wildcard.Matchers {
+	return t.instrumentationConfig().ignoreTransactionURLs
+}
```
axw, line 501: On second thoughts, can you please make this method accept a URL and return a boolean? Package "wildcard" is internal, so we shouldn't be exposing it here. e.g.

```suggestion
// IgnoredTransactionURL reports whether transactions for the given URL should be ignored.
func (t *Tracer) IgnoredTransactionURLMatchers(url *url.URL) bool {
	return t.instrumentationConfig().ignoreTransactionURLs.MatchAny(url.String())
}
```

module/apmhttp/ignorer.go
```diff
@@ -53,6 +56,15 @@ func DefaultServerRequestIgnorer() RequestIgnorerFunc {
 	return defaultServerRequestIgnorer
 }
 
+// DynamicServerRequestIgnorer returns the RequestIgnorer to use in
+// handlers. The list of wildcard patterns comes from central config
+func DynamicServerRequestIgnorer(t *apm.Tracer) RequestIgnorerFunc {
```
axw, line 61:
```suggestion
func NewDynamicServerRequestIgnorer(t *apm.Tracer) RequestIgnorerFunc {
```

Let's align the naming with the other functions below, since DefaultServerRequestIgnorer will be going away eventually.

axw: In fact, if you implement the suggestion above about changing Tracer.IgnoredTransactionURLMatchers to IgnoredTransactionURL, then we can just pass `tracer.IgnoredTransactionURL` directly in as it will satisfy the function signature -- no need for (New)DynamicServerRequestIgnorer then.

CHANGELOG.asciidoc
```diff
@@ -12,7 +12,7 @@ endif::[]
 
 [float]
 ===== Features
-* Cool new feature: {pull}2526[#2526]
+- Make TRANSACTION_IGNORE_URLS dynamically configurable: {pull}872[#872]
```
axw, line 15: The changelog entry should be lower down, under "View commits" on line 24. We don't currently split up between fixes, features, etc.

module/apmhttp/ignorer.go
axw: I realise now that RequestIgnorerFunc takes an *http.Request, whereas IgnoredTransactionURL takes a URL, so my second comment doesn't work. But let's please align the naming, and add `New` to the front.

module/apmhttp/ignorer_test.go
```diff
@@ -65,11 +65,29 @@ func testDefaultServerRequestIgnorer(t *testing.T, ignoreURLs string, r *http.Re
...
+	r := &http.Request{URL: &url.URL{Path: "/foo"}}
+	tracer := newTracer()
+	defer tracer.Close()
+
+	dynamicIgnorer := apmhttp.DynamicServerRequestIgnorer(tracer)
+	assert.Equal(t, false, dynamicIgnorer(r))
+
+	tracer.SetIgnoreTransactionURLs("/fo*")
+	assert.Equal(t, true, dynamicIgnorer(r))
+}
```
axw, line 89: Apart from the call to `tracer.SetIgnoreTransactionURLs`, is this testing anything that TestServerRequestIgnorer doesn't already? Doesn't look like it. If so let's drop this and add a test for Tracer.SetIgnoreTransactionURLs in the top level package if needed.

axw (changes requested):

module/apmhttp/ignorer_test.go
```diff
@@ -65,11 +65,29 @@ func testDefaultServerRequestIgnorer(t *testing.T, ignoreURLs string, r *http.Re
...
+	r := &http.Request{URL: &url.URL{Path: "/foo"}}
+	tracer := newTracer()
+	defer tracer.Close()
+
+	dynamicIgnorer := apmhttp.NewDynamicServerRequestIgnorer(tracer)
+	assert.Equal(t, false, dynamicIgnorer(r))
+
+	tracer.SetIgnoreTransactionURLs("/fo*")
+	assert.Equal(t, true, dynamicIgnorer(r))
+}
```
axw, line 89:
```suggestion
```

axw (approved):
LGTM once the extraneous test is removed. Thank you!

PR #875, opened 2021-01-23
Change go redis version to v8.4.9 and update the test. In go-redis v8.4.9

sangianpatrick:
ClusterClient wraps the command with MULTI/EXEC in go redis v8.4.9


sangianpatrick:
I just change the version of go redis in apm agent to v8.4.9 and update the test
It's because I can't use the hook since the go redis v8 update the tag

Issue #654, opened 2019-10-16
Agent v2 meta issue


axw:
Meta issue for tracking breaking changes that should go into the next major version of the agent, v2.0.0.

- [x] https://github.com/elastic/apm-agent-go/issues/655
- [x] https://github.com/elastic/apm-agent-go/issues/775
- [x] https://github.com/elastic/apm-agent-go/issues/646
- [x] https://github.com/elastic/apm-agent-go/issues/465
- [x] https://github.com/elastic/apm-agent-go/issues/1184
- [x] https://github.com/elastic/apm-agent-go/issues/1136
- [x] https://github.com/elastic/apm-agent-go/issues/1194
- [x] confirm minimum supported Go version for v2

axw:
@elastic/apm-agent-go this is done according to the check boxes, but I'd like it if everyone could have a look through the code and have a think about whether anything else can be removed or improved (requiring breaking changes) before we commit to v2. Like v1, I expect it to last for quite some time.

simitt:
I went through the open issues, have you considered whether https://github.com/elastic/apm-agent-go/issues/773 should be implemented? If not, then we should probably close it, as this would be considered a breaking change.

PR #876, opened 2021-01-24
Change go redis version to v8.4.9 in apmgoredisv8 and update the test

sangianpatrick:
ClusterClient wraps the command with MULTI/EXEC in go redis v8.4.9


module/apmgoredisv8/hook_test.go
```diff
@@ -120,14 +120,14 @@ func TestHookTxPipeline(t *testing.T) {
...
-			if _, ok := client.(*redis.Client); ok {
-				// *redis.Client wraps queued commands with MULTI/EXEC
-				// in (*redis.Client) processTxPipeline
-				assert.Equal(t, "MULTI, GET, SET, GET, (empty command), EXEC", spans[0].Name)
-			} else {
-				// *redis.ClusterClient and *redis.Ring don't wrap queued commands
-				// with MULTI/EXEC in (*redis.Client) processTxPipeline
+			if _, ok := client.(*redis.Ring); ok {
+				// *redis.Client wraps queued commands without MULTI/EXEC
+				// in (*redis.Ring) processTxPipeline
```
axw, line 125:
```suggestion
				// *redis.Ring does not wrap queued commands with MULTI/EXEC
				// in (*redis.Ring) processTxPipeline
```

I think? IIUC the only change is that ClusterClient now *does* wrap them.

axw (approved):
LGTM, thanks @sangianpatrick! I think there's a minor adjustment needed in a comment as noted. Please take a look and update, and then I'll merge.

axw:
jenkins run the tests please

module/apmgoredisv8/hook_test.go
sangianpatrick: @axw thanks for reminding me about the comment, already fixed it.

axw:
@sangianpatrick the tests are failing with Go 1.11 and 1.12:

> [2021-01-25T04:13:39.534Z] ../../../../../pkg/mod/github.com/go-redis/redis/v8@v8.4.9/redis.go:222:13: undefined: errors.Unwrap

go-redis has bumped the minimum required version of Go to 1.13 in a minor release. If we were to increase the version in the agent, then it would break users who are still on older versions of Go. We consider that a breaking change, and not something we will do in a minor version.

I think this change would need to go into https://github.com/elastic/apm-agent-go/issues/654, or otherwise changes are required in go-redis to support older versions of Go.

You can still bump the version of go-redis in your application's `go.mod` and the instrumentation should work as expected.

axw:
Closing - can't update because of the non-semver change upstream. Thanks all the same.

PR #902, opened 2021-02-17
Fix gRPC status code outcome mapping, implement Gherkin outcome spec

axw:
Fix the mapping of gRPC status codes to outcomes. In particular, the "Unknown" status should be considered a failure as it describes an unknown error -- not that the status is unknown. All status codes other than OK are failures from the client's perspective.

Also, implement the Gherkin outcome spec.

Supersedes #898


simitt (approved):
LGTM

PR #904, opened 2021-02-20
module/apmgrpc: record underlying HTTP/2 context

### LINKED ISSUE #900: Add Support for storing HTTP Headers (Req & Resp) for gRPC calls
There does not seem to be any documentation on how view headers from a gRPC request. 

For example, the `Host` gets set for `apmhttp` (See https://github.com/elastic/apm-agent-go/pull/280) but not for `apmgrpc`. Why not, when grpc is over an HTTP2 connection?

I would love if it could store the [same request headers](https://github.com/elastic/apm-agent-go/blob/6c3b46004fd1b83abc2089fc399739affc48bde2/context.go#L162) and the [same response headers](https://github.com/elastic/apm-agent-go/blob/6c3b46004fd1b83abc2089fc399739affc48bde2/context.go#L225) as `apmhttp` modules does. Thus, its easy to see standard headers from the `apmgrpc` interceptors. Ideality, it should happen in `getIncomingMetadataTraceContext` since it already pulls the Request Metadata.

Alternatives include creating our own interceptor that runs after the `apmgrpc` interceptor. It adds the headers to the parent span.

**Additional context**

axw:
Record underlying HTTP/2 context:
 - URL using :authority for host, presence of TLS auth info for scheme, and full method name for path.
 - gRPC metadata as HTTP request headers
 - peer IP for remote address in transactions

Closes #900


stuartnelson3 (approved):
looks good!

PR #906, opened 2021-03-04
Capture request body in CaptureError calls

### LINKED ISSUE #877: apm.CaptureError should record request body when passed an apmhttp context
**Describe the bug**

When using `apmhttp` (or some other HTTP server instrumentation that supports request body capture), the request body will be captured either for all transactions, or for panics, depending on the capture mode (transactions, errors, all).

If an error is explicitly captured, using `apm.CaptureError`, then the request body is not automatically captured.

**Expected behavior**

Calls to `apm.CaptureError`, when given a context from an instrumented HTTP server, should automatically capture the request body if the tracer is configured appropriately.

jalvz:
~~This is very chicken-eggy, but didn't find another way to do it.~~

Should fix #877


axw (commented):
Looks good, but I'm wondering if we can simplify this a bit and avoid changing or adding to the public API.

Also I just realised that `BodyCapturer` was never made safe for concurrent access, since there has been an assumption that it will only be used before and after the handler is called. We'll need to add a `sync.RWMutex` in there to avoid races.

A couple of thoughts on how we might be able to simplify:
 - If we call `Context.SetHTTPRequest` when the transaction is started (i.e. in `apmhttp.StartTransaction`), then we can copy that context to the error in `apm.Error.SetTransaction`.
 - If we store the `BodyCapturer` in `TransactionData` when the transaction is started (again in `apmhttp.StartTransaction`), then we can (a) copy it across to the transaction, and (b) call `Context.SetHTTPRequestBody` in `Transaction.End` and `Error.Send`.

Does that sound viable?

jalvz:
> If we call Context.SetHTTPRequest when the transaction is started (i.e. in apmhttp.StartTransaction), then we can copy that context to the error in apm.Error.SetTransaction.

Yep, missed that - thanks.

> If we store the BodyCapturer in TransactionData when the transaction is started (again in apmhttp.StartTransaction), then we can (a) copy it across to the transaction, and (b) call Context.SetHTTPRequestBody in Transaction.End and Error.Send

Not so sure... First, I think `tx.End()` would be doing 1 thing too many. Right now it records timing information and buffers the transaction for delivery which what the name suggest to me without looking at the implementation. Then, I think the bodycapturer is still needed in `ServeHTTP`, so how that would look like?

I took the liberty to try something else, let me know if that works

module/apmhttp/handler_test.go
```diff
@@ -107,6 +108,36 @@ func TestHandler(t *testing.T) {
...
+	req, _ := http.NewRequest("POST", "http://server.testing/foo", strings.NewReader("foo-body"))
+	req.Header.Set("User-Agent", "apmhttp_test")
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	req.RemoteAddr = "client.testing:1234"
+	h.ServeHTTP(w, req)
+	tracer.Flush(nil)
+
+	payloads := transport.Payloads()
+	e := payloads.Errors[0]
+	assert.Equal(t, &model.RequestBody{Raw: "foo-body"}, e.Context.Request.Body)
```
axw, line 138: We should also check that the transaction still has the request body.

module/apmecho/middleware.go
```diff
@@ -72,8 +72,9 @@ func (m *middleware) handle(c echo.Context) error {
 	name := req.Method + " " + c.Path()
 	tx, req := apmhttp.StartTransaction(m.tracer, name, req)
 	defer tx.End()
-	c.SetRequest(req)
 	body := m.tracer.CaptureHTTPRequestBody(req)
+	req.WithContext(apm.ContextWithBodyCapturer(req.Context(), body))
```
axw, line 76:
```suggestion
	req = apmhttp.RequestWithContext(req, apm.ContextWithBodyCapturer(req.Context(), body))
```

`req.WithContext` returns a copy with the context set. Without reassigning req, that WithContext is a no-op.
apmhttp.RequestWithContext is a slightly more optimised version of req.WithContext.

module/apmhttp/handler_test.go
axw: Also, can you please add this new test to `apmtest.HTTPTestSuite` instead? This functionality should be common across each of the HTTP integrations. If the implementation was all inside `apmhttp.StartTransaction` then it would be fine to test in one place.

axw (changes requested):
> I took the liberty to try something else, let me know if that works

Your approach is fine. I don't really love increasing the API surface area, but I suppose it keeps things a bit simpler conceptually.

One thing I don't like about it is how it has led to the same bug in each of the HTTP instrumentation modules (not reassigning the result of `req.WithContext`). Perhaps we should introduce a new extended version of the `apmhttp.StartTransaction`, which also takes care of creating the body capturer and adding it to the request context:

```go
package apmhttp

// DEPRECATED, use StartTransactionWithBody instead.
func StartTransaction(tracer *apm.Tracer, name string, req *http.Request) (*apm.Transaction, *http.Request) {
    tx, _ , req := StartTransactionWithBody(tracer, name, req)
    return tx, req
}

func StartTransactionWithBody(tracer *apm.Tracer, name string, req *http.Request) (*apm.Transaction, *apm.BodyCapturer, *http.Request) {
    ...
}
```

BodyCapturer still needs to be updated to make it safe for concurrent access. If you're just asking for high level feedback, please move this back into draft.

jalvz:
I got a bit lost implementing the suggestions and now I can't make the tests pass, moving this back to draft for the moment

apmtest/httpsuite.go
```diff
@@ -135,3 +138,25 @@ func (s *HTTPTestSuite) TestPanicAfterWrite() {
 	s.Equal(tx.ID, e.ParentID)
 	s.Equal(resp.StatusCode, e.Context.Response.StatusCode)
 }
+
+// TestExplicitErrorCapture tests that a CaptureError explicit call
```
axw, line 142: thanks, this looks great :)

context.go
```diff
@@ -178,6 +178,9 @@ func (c *Context) SetHTTPRequest(req *http.Request) {
 		HTTPVersion: httpVersion,
 		Cookies:     req.Cookies(),
 	}
+	if c.requestBody.Form == nil && req.Form != nil {
+		c.requestBody.Form = req.Form
+	}
```
axw, line 183: Nice idea!

A few small issues (see SetHTTPRequestBody and BodyCapturer.setContext):
 - we should use req.PostForm instead of req.Form
 - we should be copying and truncating the values
 - we should honour c.captureBody (capturing body can be disabled through config)

My recommendation is that we record the net/http.Request on the context object, and then update SetHTTPRequestBody to pass it into BodyCapturer.setContext. Something like:

Here:
```go
    c.httpRequest = req // store for SetHTTPRequestBody
```

SetHTTPRequestBody
```go
    if bc.setContext(&c.requestBody, c.httpRequest) {
```

BodyCapturer.setContext:
```go
func (bc *BodyCapturer) setContext(out *model.RequestBody, req *http.Request) bool {
    // bc.request may have been copied, hence its PostForm will not be modified.
    // req is the request passed to the instrumented handler.
    if req.PostForm != nil {
        ...
```

gocontext.go
```diff
@@ -130,6 +143,11 @@ func CaptureError(ctx context.Context, err error) *Error {
 		}
 		e := tx.tracer.NewError(err)
 		e.Handled = true
+		bc := bodyCapturerFromContext(ctx)
+		if bc != nil {
+			e.Context.SetHTTPRequest(bc.request)
```
axw, line 148: this will be the pre-copy version of the request, so PostForm won't be set on this one...
I think this fine for now

module/apmhttp/handler.go
```diff
@@ -116,11 +118,24 @@ func StartTransaction(tracer *apm.Tracer, name string, req *http.Request) (*apm.
...
 
+// StartTransactionWithBody returns a new Transaction with name,
+// created with tracer, and taking trace context from req.
+//
+// If the transaction is not ignored, the request and the request body
+// capturer will be returned with the transaction added to its context.
+func StartTransactionWithBody(tracer *apm.Tracer, name string, req *http.Request) (*apm.Transaction, *apm.BodyCapturer, *http.Request) {
+	tx, req := StartTransaction(tracer, name, req)
+	bc := tracer.CaptureHTTPRequestBody(req)
+	req = RequestWithContext(apm.ContextWithBodyCapturer(req.Context(), bc), req)
```
axw, line 135:
```suggestion
	if bc != nil {
		req = RequestWithContext(apm.ContextWithBodyCapturer(req.Context(), bc), req)
	}
```

bc is nil if body capturing is disabled

apmtest/httpsuite.go
```diff
@@ -135,3 +138,25 @@ func (s *HTTPTestSuite) TestPanicAfterWrite() {
...
+// inside an HTTP request transaction captures the request body
+func (s *HTTPTestSuite) TestExplicitErrorCapture() {
+	resp, err := http.Post(
```
axw, line 145:
```suggestion
	s.Tracer.SetCaptureBody(apm.CaptureBodyAll)
	resp, err := http.Post(
```

axw: can we do it once here rather than in each package?

module/apmhttp/client_test.go
```diff
@@ -383,3 +383,37 @@ type RoundTripFunc func(*http.Request) (*http.Response, error)
 func (f RoundTripFunc) RoundTrip(req *http.Request) (*http.Response, error) {
 	return f(req)
 }
+
+// TestCaptureErrorWithRequestBody tests that a CaptureError explicit call
```
axw, line 387: can we delete this one now?

module/apmhttp/handler.go
```diff
@@ -116,11 +118,24 @@ func StartTransaction(tracer *apm.Tracer, name string, req *http.Request) (*apm.
 		traceContext.State, _ = ParseTracestateHeader(req.Header[TracestateHeader]...)
 	}
 	tx := tracer.StartTransactionOptions(name, "request", apm.TransactionOptions{TraceContext: traceContext})
+	tx.Context.SetHTTPRequest(req)
```
axw, line 121:
```suggestion
```

we shouldn't be calling this here

axw (changes requested):
Thanks, I'm happy with this approach. I'm pretty sure there's an issue with capturing PostForm in errors, but we can iterate on that later.

axw (approved):
thanks for sticking with it, looks good! please update the changelog before merging

PR #907, opened 2021-03-05
scripts/jenkins: fix CI builds on 1.16

axw:
In 1.16, by default go will complain if go.mod
is missing dependencies. In CI this is occurring
due to the installation of tools that are not
present in the top-level go.mod.

To avoid adding dependencies for agent consumers,
introduce a go.mod specifically for installing
tools used only in CI.

### DIFF
--- a/scripts/docker-test.sh
+++ b/scripts/docker-test.sh
@@ -1,3 +1,0 @@
-#!/usr/bin/env bash
-set -euxo pipefail
-
--- a/scripts/docker-test.sh
+++ b/scripts/docker-test.sh
@@ -1,19 +1,3 @@
 #!/usr/bin/env bash
 set -euxo pipefail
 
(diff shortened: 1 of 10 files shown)


scripts/docker-test.sh
```diff
@@ -1,19 +1,3 @@
 #!/usr/bin/env bash
 set -euxo pipefail
```
v1v, line 3: Can it be deleted this file since it's empty?

v1v (approved):

scripts/docker-test.sh
axw: :man_facepalming: yes, thanks

axw:
Not sure why MacOS failed, but it's not related to this PR AFAICT. Maybe because my branch doesn't include the ARM64 changes. I'll merge and follow up if needed.

PR #918, opened 2021-03-24
module/apmgrpc: add stream interceptors

### LINKED ISSUE #901: gRPC Stream Interceptors
There currently exists within `apmgrpc` [`UnaryClientInterceptor`](https://github.com/elastic/apm-agent-go/blob/master/module/apmgrpc/client.go#L41) and [`UnaryServerInterceptor`](https://github.com/elastic/apm-agent-go/blob/master/module/apmgrpc/server.go#L51). 

I see within the [documentation](https://www.elastic.co/guide/en/apm/agent/go/master/builtin-modules.html#builtin-modules-apmgrpc) it says no Stream support exists for gRPC.
> There is currently no support for intercepting at the stream level. Please file an issue and/or send a pull request if this is something you need.

Well, here I am asking for a Client and Server Stream Interceptor for gRPC use. 🙃 
We have many services that use gRPC Stream APIs, and this is sadly one of our blockers for a quick migration.

A similar interceptor (just as easy) as the `UnaryClientInterceptor` and `UnaryServerInterceptor`.

**Additional context**
Unrelated, but gRPC specific: Would be wonderful if we could also get #900 included.

axw:
Add server and client stream interceptors. Very similar to unary interceptors -- the transaction and span represents the entire gRPC stream.

Client spans are ended when the stream is closed, which is observed when stream methods return with an error, or for unary servers, RecvMsg returns (not necessarily with an error). This is very similar to the client wrapper in apmhttp, where the span is ended only after the response body is consumed.

Closes #901


module/apmgrpc/server.go
```diff
@@ -94,6 +95,62 @@ func NewUnaryServerInterceptor(o ...ServerOption) grpc.UnaryServerInterceptor {
...
+	) (err error) {
+		if !opts.tracer.Recording() || opts.streamIgnorer(info) {
+			return handler(srv, stream)
+		}
+		ctx := stream.Context()
+		tx, ctx := startTransaction(ctx, opts.tracer, info.FullMethod)
+		defer tx.End()
+
+		// TODO(axw) define context schema for RPC,
+		// including at least the peer address.
```
stuartnelson3, line 132: is context `span.Context` here?

stuartnelson3 (approved):
comment on a TODO but other than that looks good

module/apmgrpc/server.go
axw: yes, I'll reword to clarify

PR #928, opened 2021-04-08
Instrument DynamoDB calls

stuartnelson3:
Instrument DynamoDB calls according to the [aws](https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-aws.md#dynamodb) and [database](https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-db.md#database-spans) specs.

Happily, the `session.Session` struct only needs to be wrapped by users; the changes for instrumenting DynamoDB calls are all internal.

Note on the code:

There is a moderately sized refactoring that pulls service-specific instrumentation into structs implementing an interface. The code from the S3-specific instrumentation from the [previous pr](https://github.com/elastic/apm-agent-go/pull/927) has been moved into `s3.go`, while the DynamoDB-specific instrumentation code is in `dynamodb.go`. Some code could potentially be streamlined more, but I'm thinking it might make sense to wait for more patterns / potential abstractions to surface when implementing support for additional services.

closes https://github.com/elastic/apm-agent-go/issues/880


module/apmawssdkgo/session_test.go
```diff
@@ -91,3 +92,59 @@ func TestS3(t *testing.T) {
...
+
+func TestDynamoDB(t *testing.T) {
+	region := "us-west-2"
+	cfg := aws.NewConfig().
+		WithRegion(region).
+		WithDisableSSL(true).
+		WithCredentials(credentials.AnonymousCredentials)
+
+	spanSubtype := "dynamodb"
+	spanType := serviceTypeMap[spanSubtype]
```
axw, line 104: According to the [database spec](https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-db.md#database-span-typesubtype):
 - `type` should be "db"
 - `subtype` should be the database name (e.g. "dynamodb")

IIANM, `spanType` here will be "dynamodb". In general I think we should avoid referring to internals of the implementation like this, as I find it makes issues like this less obvious. Instead, use literals for the expected values.

module/apmawssdkgo/dynamodb.go
```diff
@@ -0,0 +1,86 @@
...
+
+	body, err := ioutil.ReadAll(r)
+	if err != nil {
+		return nil, nil, err
+	}
+	b := dynamoDB{}
+	json.Unmarshal(body, &b)
+
+	return ioutil.NopCloser(bytes.NewBuffer(body)), &b, nil
+}
```
axw, line 86: I'd like to avoid touching the request body if possible. Can we use reflection for DynamoDB? It's expensive, but I'm fairly sure it's cheaper than duplicating and decoding the request body.

axw (changes requested):
Looks good overall, just a couple of things

module/apmawssdkgo/session_test.go
stuartnelson3: ah good catch, I got it mixed up with `db.type`.

>In general I think we should avoid referring to internals of the implementation like this, as I find it makes issues like this less obvious. Instead, use literals for the expected values.

I'm not sure I'm understanding. You mean to not use an incoming value like `req.ClientInfo.ServiceName` as the subtype, but match on it/some indicator in the request, and then return specific literals of what we want?

axw: no sorry, I just meant in the tests to stop referring to `serviceTypeMap`, and instead do:

```go
assert.Equal(t, "db", span.Type)
assert.Equal(t, "dynamodb", span.Subtype)
```

where possible I prefer the tests to be external test packages (apmawssdkgo_test) to force the separation

stuartnelson3: ahhh ok, got it

module/apmawssdkgo/dynamodb.go
stuartnelson3: Yeah, works fine. Reading the body and unmarshaling using reflection as well, so it probably has to be less expensive

module/apmawssdkgo/dynamodb.go
```diff
@@ -0,0 +1,92 @@
...
+
+// parseDynamoDBParams reads the request Params to parse out the TableName and
+// KeyConditionExpression, if present.
+func parseDynamoDBParams(req *request.Request) (*dynamoDB, error) {
+	b := new(dynamoDB)
+
+	params := reflect.ValueOf(req.Params).Elem()
+	if v := params.FieldByName("TableName"); v.IsValid() {
+		if n, ok := v.Interface().(*string); ok {
```
axw, line 76:
```suggestion
		if n, ok := v.Interface().(*string); ok && n != nil {
```

module/apmawssdkgo/dynamodb.go
```diff
@@ -0,0 +1,92 @@
...
+			b.TableName = *n
+		} else {
+			return nil, errors.New("could not parse TableName")
+		}
+	} else {
+		return nil, errors.New("required field TableName not present")
+	}
+
+	if v := params.FieldByName("KeyConditionExpression"); v.IsValid() {
+		if n, ok := v.Interface().(*string); ok {
```
axw, line 86:
```suggestion
		if n, ok := v.Interface().(*string); ok && n != nil {
```

module/apmawssdkgo/session.go
```diff
@@ -67,17 +77,35 @@ func send(req *request.Request) {
...
+	switch spanSubtype {
+	case serviceS3:
+		svc = newS3(req)
+	case serviceDynamoDB:
+		svc, err = newDynamoDB(req)
+		if err != nil {
+			// The only error case is if TableName is not present
+			// on the struct, but it's a required value.
+			// TODO: How do we handle an error? Or is this being
+			// overly defensive?
```
axw, line 100: I'd be inclined to create a span with the name `DynamoDB $Operation` if there's no table name. Not sure what other agents are doing. It's probably also fine to not create a span for now.

axw (approved):

axw:
Can you please add an entry to the changelog too?

module/apmawssdkgo/session.go
stuartnelson3: As far as I can see it's always required, I figure we create the span and if there's an edge case somewhere, an enterprising user will report it

axw: SGTM

stuartnelson3:
thanks for all the reviews! I included the s3 update to the changelog as well, since I didn't do that in the previous PR

PR #949, opened 2021-04-19
Add context message queue name

stuartnelson3:
Looping back to this, since it was a TODO from when first implementing `sqs` instrumentation.

### DIFF
--- a/spancontext.go
+++ b/spancontext.go
@@ -33,6 +33,7 @@ type SpanContext struct {
 	destination          model.DestinationSpanContext
 	destinationService   model.DestinationServiceSpanContext
 	destinationCloud     model.DestinationCloudSpanContext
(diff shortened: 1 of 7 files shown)


spancontext.go
```diff
@@ -68,10 +69,16 @@ type DestinationServiceSpanContext struct {
...
 
+// MessageSpanContext holds contextual information about a message.
+type MessageSpanContext struct {
+	// QueueName holds the message queue name.
+	QueueName string
+}
+
 func (c *SpanContext) build() *model.SpanContext {
 	switch {
 	case len(c.model.Tags) != 0:
```
axw, line 84:
```suggestion
	case len(c.model.Tags) != 0:
	case c.model.Message != nil:
```

axw (approved):

PR #950, opened 2021-04-20
[apmawssdkgo] instrument RPCs

stuartnelson3:
These are a few metrics I would want reported if I were to use this package myself. 100% open to discussion regarding their inclusion, in addition to implementation, naming, etc etc.

For supported services + RPCs, instrument:

- inflight requests (service, rpc)
- bytes sent (service, rpc, status code)
- bytes received (service, rpc, status code)


module/apmawssdkgo/session.go
```diff
@@ -32,26 +36,107 @@ func init() {
 	)
 }
 
+// This is a weird way to grab the *awsWrapper without exporting it. We don't
+// want to export it, but we also can't return it as an argument since the
+// linter doesn't like that.
+var wrapper *awsWrapper
```
stuartnelson3, line 42: Is there a preferred way to make this internal type accessible for testing?

module/apmawssdkgo/session.go
```diff
@@ -186,19 +271,84 @@ func send(req *request.Request) {
...
+	).Add(float64(responseSize(req)))
+
 	span.Context.SetHTTPStatusCode(req.HTTPResponse.StatusCode)
 
 	if err := req.Error; err != nil {
 		apm.CaptureError(ctx, err).Send()
 	}
 }
+
+func requestSize(req *request.Request) int {
```
stuartnelson3, line 310: this and `responseSize` were yanked from code in `client_golang`

module/apmawssdkgo/session.go
```diff
@@ -32,10 +38,50 @@ func init() {
...
+		},
+		[]string{"service", "rpc", "code"},
+	)
+)
+
 // WrapSession wraps the provided AWS session with handlers that hook into the
 // AWS SDK's request lifecycle. Supported services are listed in serviceTypeMap
 // variable below.
-func WrapSession(s *session.Session) *session.Session {
+func WrapSession(s *session.Session, opts ...Option) *session.Session {
```
axw, line 70: I'm pretty strict on backwards compatibility: I don't think we'll be able to make this signature change, or the "go1.13" build tag change, once we've released the apmawssdkgo module. (Until we bump the major version that is.)

module/apmawssdkgo/session.go
axw: maybe move to `internal/apmhttputil`? also probably needs copyright attribution

module/apmawssdkgo/session.go
```diff
@@ -52,6 +98,38 @@ func WrapSession(s *session.Session) *session.Session {
...
+	return func(c *config) {
+		c.tracer = t
+	}
+}
+
+// WithRegistry returns an Option which sets r as the prometheus.Registry to
+// use when recording AWS SDK metrics. If no registry is supplied, a new one
+// will be instantiated.Metrics will be reported to the APM server regardless
+// of the prometheus.Registry supplied.
+func WithRegistry(r *prometheus.Registry) Option {
```
axw, line 124: What's the reason for making the registry configurable? Seems to me `prometheus/client_golang` is an implementation detail. In theory we might later want to specify units for metrics, and that won't be possible as long as we use Prometheus. I also tend to think we should keep the scope minimal, and only support sending to APM Server.

axw (commented):
Looks good and I know it would be useful, but it still doesn't sit quite right with me. I wonder if we shouldn't instead be building a (much bigger, more comprehensive) feature which would allow extracting metrics from spans. @felixbarny @Mpdreamz WDYT?

module/apmawssdkgo/session.go
stuartnelson3: hm, I guess I didn't really have any reasoning other than "maybe people want to pass in their registry". But it's simpler from an API perspective, and if folks for some reason double register, prevents a panic when trying to register the same metrics twice.

module/apmawssdkgo/session.go
stuartnelson3: Yeah makes sense. Let's keep it simple and remove this.

stuartnelson3:
@axw makes perfect sense, no reason to rush anything in. I probably should have left this as a draft since I'm not married to its inclusion. But it's maybe a good place to start a conversation, just to see what something might look like, even if it's only to know this is what we don't want.

axw:
@stuartnelson3 sounds good. you can also move it to draft now (top right)

stuartnelson3:
Closing this as stale

PR #953, opened 2021-04-22
BodyCapturer: Fix NPE on benchmark when nil Context.request

### LINKED ISSUE #952: BenchmarkBodyCapturer panics with NPE
**Describe the bug**
I couldn't help notice that the benchmark `BenchmarkBodyCapturer` [panics](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-go%2Fapm-agent-go-mbp/detail/master/236/pipeline/) and wanted to see exactly why.

By looking at the code it looks to me that the `tracer.Context.request` field is never set, thus when `context.SetHTTPRequestBody` and `bc.setContext(&c.requestBody, c.httpRequest)` is called, the passed `*http.Request` is nil.

The panic can be solved in a couple of ways, either by changing benchmark to call `tx.Context.SetHTTPRequest(req)` in line 36 or by adding a `req != nil` check in [capturebody.go:140](https://github.com/elastic/apm-agent-go/blob/d9758ef8a6b48310fd596f67aee9da3f062ea48f/capturebody.go#L140).

**To Reproduce**
* Check out master.
* Run the benchmark with `go test -v -run none -benchtime 3s -benchmem -bench BenchmarkBody .`
* You'll see the panic.

**Expected behavior**

Benchmarks shouldn't panic.

marclop:
## Description
Fixes a nil pointer exception on the benchmark `BenchmarkBodyCapturer`.
The `Context.request` field wasn't set in the benchmark and is being
passed in the `BodyCapturer.setContext()` call as nil, causing an NPE
when `req.PostForm` is accessed.

The bug seems to have been introduced in #906.

This might not be the right fix since I'm not very familiar with the
codebase.

## Related issues

Closes #952

### DIFF
--- a/capturebody.go
+++ b/capturebody.go
@@ -137,7 +137,7 @@ func (bc *BodyCapturer) Discard() {
 }
 
 func (bc *BodyCapturer) setContext(out *model.RequestBody, req *http.Request) bool {
-	if req.PostForm != nil {
+	if req != nil && req.PostForm != nil {
 		// We must copy the map in case we need to
 		// sanitize the values. Ideally we should only
 		// copy if sanitization is necessary, but body


marclop:
Verified with:

```console
$ go test -v -run none -benchtime 3s -benchmem -bench BenchmarkBody .
goos: darwin
goarch: amd64
pkg: go.elastic.co/apm
cpu: Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
BenchmarkBodyCapturer
BenchmarkBodyCapturer-8   	35825356	        93.33 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	go.elastic.co/apm	3.916s
```

axw (approved):
thank you :)

jalvz (commented):
thanks Marc!

PR #956, opened 2021-04-23
apm: add ParentID methods to Span and Transaction

### LINKED ISSUE #909: Introduce a way to extract the parent ID from Transactions and Spans

It is currently not possible to extract the parent ID from a Transaction or Span. This makes it impossible to, for example, inject the parent ID into logs. See https://discuss.elastic.co/t/how-to-get-parent-span-id-from-span/266430

Add a method to Transaction and Span to obtain the parent ID.

Alternatively, simply expose the parent ID as a field on Transaction and Span.

marclop:
## Description
Adds a new `parentID` field to both `Transaction` and `Span` structures.

The `parentID` is only populated at start time and remains immutable
throughout the lifecycle of the Transaction or Span, making it possible
to retrieve the `parentID` for either after they have ended.

However, the `ParentID()` call might return a zero value `SpanID` when
a trace has no parent.

Removes the `TransactionData.parentSpan` field in favor of the new
`Transaction.parentID` and updates `EnsureParent()` to update that field
and `Transaction.ParentID()` to hold a read lock to read `tx.parentID`.

Last, removes the `SpanData.parentID` field and instead relies on the
new top level `Span.parentID`.

## Related issues
Closes #909.


span.go
```diff
@@ -316,6 +316,15 @@ func (s *Span) End() {
 	s.SpanData = nil
 }
 
+// ParentID returns the span's Parent ID as SpanID.
+// If Span is nil or has been ended, a zero (invalid) SpanID is returned.
+func (s *Span) ParentID() SpanID {
+	if s.SpanData == nil || s.parentID.isZero() {
```
axw, line 322:
```suggestion
	s.mu.RLock()
	defer s.mu.RUnlock()
	if s.ended() || s.parentID.isZero() {
```

`s.SpanData` must be accessed with `s.mu` held

span.go
```diff
@@ -316,6 +316,15 @@ func (s *Span) End() {
 	s.SpanData = nil
 }
 
+// ParentID returns the span's Parent ID as SpanID.
```
axw, line 319:
```suggestion
// ParentID returns the ID of the span's parent span or transaction.
```

span.go
axw: Having said that, I think we should create a new `Span.parentID` field, so it can be accessed even after the span has been ended, like TraceContext. If we do that, no mutex will be required, as parent ID will be immutable.

transaction.go
```diff
@@ -236,6 +236,15 @@ func (tx *Transaction) EnsureParent() SpanID {
...
 }
 
+// ParentID returns the transaction's Parent ID generating a parent
+// span ID if one has not already been set and tx has not been ended.
+// If tx is nil or has been ended, a zero (invalid) SpanID is returned.
+//
+// Internally calls `Transaction.EnsureParent()`.
+func (tx *Transaction) ParentID() SpanID {
+	return tx.EnsureParent()
+}
```
axw, line 246: I'm not too keen on this -- EnsureParent is for the fairly specific, and uncommon use case of generating IDs for an upstream, such as RUM.

I think `ParentID` should behave the same both for spans and transactions: return a zero SpanID if there is no parent, otherwise return the parent ID.

docs/api.asciidoc
```diff
@@ -152,6 +152,38 @@ func initialPageHandler(w http.ResponseWriter, req *http.Request) {
...
+		"trace_id":              traceContext.Trace,
+		"transaction_id":        traceContext.Span,
+		"transaction_parent_id": tx.ParentID(),
+	}
+	if span := apm.SpanFromContext(ctx); span != nil {
+		fields["span_id"] = span.ParentID()
+	}
+	return fields
+}
+----
```
axw, line 185: I feel like all the boilerplate around the use of `span.ParentID` is a bit obscuring. Do we need an example for this? I'm not sure it's going to be commonly used (it's only come up once AFAICR), so might not be worthwhile.

docs/api.asciidoc
```diff
@@ -313,6 +345,35 @@ SpanFromContext returns a span previously stored in the context using
...
+[[span-parentid]]
+==== `func (*Span) ParentID() SpanID`
+
+ParentID returns the span's Parent ID generating a parent span ID if one has
+not already been set and tx has not been ended.
+If tx is nil or has been ended, a zero (invalid) SpanID is returned.
+
+[source,go]
+----
+// TraceContext returns a map[string]interface{} containing the trace
```
axw, line 358: ditto to previous comment

axw (changes requested):
Thanks @marclop! Looks good overall, but I'd like to request a couple of main changes:
 - Don't generate a parent ID in Transaction.ParentID (name doesn't suggest that, and I don't think it's generally desirable behaviour anyway)
 - Add parentID as an immutable field to the top-level Transaction and Span types. When creating a transaction or span, set these fields. Then we can access them even after they're ended, like we can for TraceContext.

docs/api.asciidoc
marclop: yeah, It might be too much to have an example here, I think it's fair to remove the example and see if it needs an example after it's been released.

transaction.go
marclop: I think that makes more sense than what I initially did, by adding a new top level field `parentID` should reduce the amount of work this method does.

docs/api.asciidoc
```diff
@@ -156,33 +156,7 @@ See the {apm-rum-ref}/index.html[JavaScript RUM agent documentation] for more in
...
-		"transaction_id":        traceContext.Span,
-		"transaction_parent_id": tx.ParentID(),
-	}
-	if span := apm.SpanFromContext(ctx); span != nil {
-		fields["span_id"] = span.ParentID()
-	}
-	return fields
-}
-----
+ParentID returns the ID of the transaction's Parent or a zero (invalid) SpanID.
```
axw, line 159:
```suggestion
ParentID returns the ID of the transaction's parent, or a zero (invalid) SpanID if the transaction has no parent.
```

docs/api.asciidoc
```diff
@@ -349,30 +323,8 @@ does not contain a span.
...
-		"transaction_id": traceContext.Span,
-	}
-	if span := apm.SpanFromContext(ctx); span != nil {
-		fields["span_id"] = span.ParentID()
-	}
-	return fields
-}
-----
+ParentID returns the ID of the span's parent span or transaction.
+If Span is nil or has been ended, a zero (invalid) SpanID is returned.
```
axw, line 327:
```suggestion
ParentID returns the ID of the span's parent.
```

Spans can't exist without a parent. It's true that a zero SpanID is returned for a nil Span, but I'm not sure we need to explain that in these docs?

span.go
```diff
@@ -316,12 +318,9 @@ func (s *Span) End() {
...
 }
 
-// ParentID returns the span's Parent ID as SpanID.
+// ParentID returns the ID of the span's parent span or transaction.
 // If Span is nil or has been ended, a zero (invalid) SpanID is returned.
 func (s *Span) ParentID() SpanID {
-	if s.SpanData == nil || s.parentID.isZero() {
-		return SpanID{}
-	}
 	return s.parentID
```
axw, line 324: we probably should keep the nil check, like in Span.TraceContext()

span.go
```diff
@@ -316,12 +318,9 @@ func (s *Span) End() {
 	s.SpanData = nil
 }
 
-// ParentID returns the span's Parent ID as SpanID.
+// ParentID returns the ID of the span's parent span or transaction.
 // If Span is nil or has been ended, a zero (invalid) SpanID is returned.
```
axw, line 322:
```suggestion
```

transaction.go
```diff
@@ -236,13 +238,9 @@ func (tx *Transaction) EnsureParent() SpanID {
...
 
-// ParentID returns the transaction's Parent ID generating a parent
-// span ID if one has not already been set and tx has not been ended.
-// If tx is nil or has been ended, a zero (invalid) SpanID is returned.
-//
-// Internally calls `Transaction.EnsureParent()`.
+// ParentID returns the ID of the transaction's Parent or a zero (invalid) SpanID.
 func (tx *Transaction) ParentID() SpanID {
-	return tx.EnsureParent()
+	return tx.parentID
```
axw, line 243: we probably should keep the nil check, like in Transaction.TraceContext()

axw (commented):
Looks good, thanks! Just a few minor things.

I think we'll hold off on merging this until the next release. We probably need a new branching strategy now that we're doing feature freezes...

axw (approved):
lovely, thanks for the updates :)

marclop:
Still TODO is entry in the change log, once the upcoming version is released, we should update this PR with the change log entry.

axw:
> I think we'll hold off on merging this until the next release. We probably need a new branching strategy now that we're doing feature freezes...

Turns out I misunderstood the agreed upon definition of feature freeze for agent development. @marclop it's OK to merge this now (with changelog updated).

axw:
Windows failed due to CI platform issues. Merging, nothing platform-specific here.

PR #960, opened 2021-05-21
chore: fixes vanity import.

### LINKED ISSUE #844: ci: add check that vanity import comments exist
See https://github.com/elastic/apm-agent-go/pull/841#pullrequestreview-523192012. Like https://github.com/elastic/apm-agent-go/blob/master/scripts/check_goimports.sh, we could call https://github.com/jcchavezs/porto to see if changes are required and exit with an error if so.

jcchavezs:
This PR adds a check for vanity imports to be updated on CI.

Ping @axw

Closes #844

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ precheck: check-goimports check-lint check-vanity-import check-vet check-dockerf
 .PHONY: check-lint
 .PHONY: check-licenses
 .PHONY: check-modules
+.PHONY: check-vanity-import
 ifeq ($(shell go run ./scripts/mingoversion.go -print 1.12),true)
 check-goimports:
 	sh scripts/check_goimports.sh
@@ -27,12 +28,16 @@ check-licenses:
 
 check-modules:
 	go run scripts/genmod/main.go -check .
+
+check-vanity-import:
+	@if [[ $(porto --skip-files ".*\\.pb\\.go$" -l . | wc -c) -ne 0 ]]; then echo "Vanity imports are not up to date" ; exit 1 ; fi
 else
 check-goimports:
 check-dockerfile-testing:
 check-lint:
 check-licenses:
 check-modules:
+check-vanity-import:
 endif
 
 .PHONY: check-vet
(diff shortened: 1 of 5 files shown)


scripts/before_install.sh
```diff
@@ -5,6 +5,7 @@ if (go run scripts/mingoversion.go 1.12 &>/dev/null); then
   go install -v golang.org/x/lint/golint
   go install -v golang.org/x/tools/cmd/goimports
   go install -v github.com/elastic/go-licenser
+  go get -v github.com/jcchavezs/porto
```
jcchavezs, line 8:
```suggestion
  go install -v github.com/jcchavezs/porto
```

Makefile
```diff
@@ -85,3 +85,7 @@ model/marshal_fastjson.go: model/model.go
 
 scripts/Dockerfile-testing: $(wildcard module/*)
 	go generate ./scripts
+
+.PHONY: check-vanity-import
+check-vanity-import:
```
axw, line 90: can you please move this up with the other conditional targets at line 15? and define an empty target for Go < 1.12

axw (changes requested):
thanks! I'd just like to shuffle the Makefile change up to be with the other conditional "check-" targets please

jcchavezs:
@axw I addressed your comments.

scripts/before_install.sh
axw: Any reason not to make that change? I would prefer to have consistency

axw (approved):
If we can change `go get` to `go install` for consistency that'd be great. Otherwise LGTM, thanks!

scripts/before_install.sh
jcchavezs: I think I did not commit it by mistake.

axw:
Tests are failing but pretty certain that's unrelated, so merging. Thanks!

axw:
> Tests are failing but pretty certain that's unrelated, so merging. Thanks!

Oops, looks like it's related to the testify upgrade actually. I'll sort it out.

PR #964, opened 2021-05-28
Fix tests, update go.mod/go.sum

axw:
`module/apmot` tests are failing due to a change in testify. We fix that by passing a pointer type to `testify/suite.Run`. While we're at it, clean up the go.mod and go.sum files.


axw:
MacOS is failing with "Vanity imports are not up to date"

PR #974, opened 2021-06-08
[fasthttp] Atreugo support!

savsgio:


axw:
@savsgio thanks for the PR. Sorry, but I don't think Atreugo is sufficiently popular to justify including in the apm-agent-go repo. There's a maintenance cost associated with each module we accept, and `fasthttp` was already borderline.

I think it should be possible for you to continue maintaining this externally in https://github.com/atreugo/apm. Please let us know if there's anything missing in the agent's API to make that easier.

Issue #752, opened 2020-04-30
Support configuring a server CA certificate


axw:
**Is your feature request related to a problem? Please describe.**

Configuring the agent to securely communicate with multiple APM Servers is impractical. You can currently configure the agent to pin a specific server certificate, but this is not practical for multiple servers each with their own certificate.

**Describe the solution you'd like**

Add support for specifying CA certificate(s), like in the Ruby agent, via [ELASTIC_APM_SERVER_CA_CERT](https://www.elastic.co/guide/en/apm/agent/ruby/current/configuration.html#config-ssl-ca-cert).

PR #982, opened 2021-07-02
Adding CA cert functionality to elastic

bmacnee:
https://github.com/elastic/apm-agent-go/issues/752

Attempting to add functionality similar to other elastic-apm agents where a ca cert can be provided instead of the literal server cert.

I am a bit confused by the comment `TestHTTPTransportServerCert` :
```	
// Reconfigure the transport so that it knows about the
// CA certificate. We avoid using server.Client here, as
// it is not available in older versions of Go.
```
	
This refers to a CA certificate, but my understanding is this functionality would only work for the servers cert.


bmacnee:
signed the Agreement after the fact, dunno if there's a way to kick off the githooks again...

transport/http.go
```diff
@@ -144,6 +145,22 @@ func NewHTTPTransport() (*HTTPTransport, error) {
 		}
 	}
 
+	caCertPath := os.Getenv(envCACert)
+	if caCertPath != "" {
+		rootCAs, err := x509.SystemCertPool()
+		if err != nil || rootCAs == nil {
+			rootCAs = x509.NewCertPool()
+		}
```
axw, line 153: I think if you configure a custom CA cert, we should not be using the system cert pool at all.

```suggestion
		rootCAs := x509.NewCertPool()
```

transport/http.go
```diff
@@ -144,6 +145,22 @@ func NewHTTPTransport() (*HTTPTransport, error) {
...
+		additionalCerts, err := ioutil.ReadFile(caCertPath)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to load root CA file from %s", caCertPath)
+		}
+		// TODO: could return an error or a log if no additional certs are loaded
+		rootCAs.AppendCertsFromPEM(additionalCerts)
```
axw, line 159: yes please :)

```suggestion
		if !rootCAs.AppendCertsFromPEM(additionalCerts) {
			return nil, fmt.Errorf("failed to load CA certs from %s", caCertPath)
		}
```

should be good enough I think?

transport/http_test.go
```diff
@@ -314,6 +314,51 @@ func TestHTTPTransportV2NotFound(t *testing.T) {
 	assert.EqualError(t, err, fmt.Sprintf("request failed with 404 Not Found: %s/intake/v2/events not found (requires APM Server 6.5.0 or newer)", server.URL))
 }
 
+func TestHTTPCACert(t *testing.T) {
```
axw, line 317:
```suggestion
func TestHTTPTransportCACert(t *testing.T) {
```

for consistency

axw (changes requested):
This is great, thanks for your contribution! Just a few small comments, otherwise LGTM.

> This refers to a CA certificate, but my understanding is this functionality would only work for the servers cert.

Yes indeed -- sorry, the comment is wrong :sweat_smile:
While you're in the vicinity, would you mind replacing "CA" with "server" in that comment?

transport/http.go
bmacnee: cool by me! will change.

bmacnee:
@axw curious on a ballpark estimate for how long it will take you folks to release the agent with these changes, as the work I'm doing is dependent on them. Thank you!! Worst case I can use my branch if it will take some time.

axw (approved):
> @axw curious on a ballpark estimate for how long it will take you folks to release the agent with these changes, as the work I'm doing is dependent on them. Thank you!! Worst case I can use my branch if it will take some time.

@bmacnee there's nothing specifically scheduled, but I anticipate we'll cut a new release in 3-4 weeks.

axw:
jenkins run the tests please

axw:
Go 1.8/1.9 failures are unrelated, will be fixed by https://github.com/elastic/apm-agent-go/pull/984 -- merging.

Thanks again!

PR #985, opened 2021-07-06
CI: use withGoEnv and remove gimme for the CI

v1v:
### What

Simplify the CI to use the `withGoEnv` step and remove the `gimme` references to configure the CI builds.

### Reason

Remove the go configuration in the CI


v1v:
Unfortunately this is not a trivial change, since `gimme` uses aliases while `gvm` uses fixed versions.

PR #991, opened 2021-07-26
[apmfasthttp] Upgrade fasthttp to v1.28.0

savsgio:


axw:
@savsgio what's the motivation behind this change? If users of apmfasthttp have a dependency on a newer version of fasthttp, then `go` will select that.

savsgio:
Keep fasthttp updated here, only!

axw:
I don't see why it's necessary. This module is only used by applications that already use fasthttp. Applications should be managing their own fasthttp dependency rather than relying on apmfasthttp to bump it.

axw:
I don't want to set a precedent for repeatedly bumping dependency versions, so I'll close this. If you think there's a need for it, I'm happy to hear an argument and reconsider.

PR #995, opened 2021-07-29
update contributing doc

stuartnelson3:
- Add a clarifying note for generating release tagging output
- Update `hub` command to `gh` (I don't think `hub` is supported/available anymore)


axw (approved):
thanks!

Issue #987, opened 2021-07-16
apmelasticsearch: send traceparent header to Elasticsearch


axw:
In https://github.com/elastic/elasticsearch/pull/74210, Elasticsearch grew support for parsing the `traceparent` header, and including `trace.id` and `transaction.id` in logs.

We should propagate the `traceparent` header in Elasticsearch requests when using `apmelasticsearch`. We currently do not add any headers.

PR #1002, opened 2021-08-10
send traceparent header to ES

stuartnelson3:
Send the traceparent header to elasticsearch, now that ES parses it from incoming requests.

Closes https://github.com/elastic/apm-agent-go/issues/987

### DIFF
--- a/module/apmelasticsearch/client.go
+++ b/module/apmelasticsearch/client.go
@@ -74,6 +74,12 @@ func (r *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 		return r.r.RoundTrip(req)
 	}
 
+	headerValue := apmhttp.FormatTraceparentHeader(tx.TraceContext())
+	req.Header.Set(apmhttp.W3CTraceparentHeader, headerValue)
+	if tx.ShouldPropagateLegacyHeader() {
+		req.Header.Set(apmhttp.ElasticTraceparentHeader, headerValue)
+	}
+
 	statement, req := captureSearchStatement(req)
 	username, _, _ := req.BasicAuth()
 	ctx = apm.ContextWithSpan(ctx, span)
--- a/module/apmelasticsearch/client.go
+++ b/module/apmelasticsearch/client.go
@@ -74,10 +74,14 @@ func (r *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 		return r.r.RoundTrip(req)
 	}
 
-	headerValue := apmhttp.FormatTraceparentHeader(tx.TraceContext())
+	traceContext := span.TraceContext()
+	headerValue := apmhttp.FormatTraceparentHeader(traceContext)
 	req.Header.Set(apmhttp.W3CTraceparentHeader, headerValue)
 	if tx.ShouldPropagateLegacyHeader() {
 		req.Header.Set(apmhttp.ElasticTraceparentHeader, headerValue)
+	}
+	if tracestate := traceContext.State.String(); tracestate != "" {
+		req.Header.Set(apmhttp.TracestateHeader, tracestate)
 	}
 
 	statement, req := captureSearchStatement(req)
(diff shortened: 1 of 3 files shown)


module/apmelasticsearch/client.go
```diff
@@ -74,6 +74,12 @@ func (r *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 		return r.r.RoundTrip(req)
 	}
 
+	headerValue := apmhttp.FormatTraceparentHeader(tx.TraceContext())
```
axw, line 77:
```suggestion
	headerValue := apmhttp.FormatTraceparentHeader(span.TraceContext())
```

axw (changes requested):
Looks good!

A couple of things:
- should we be setting the `tracestate` header too? Please check with the other agent devs if unsure
- we also need to propagate the parent transaction's traceparent when there's a non-sampled transaction in the context, or if the span is dropped (max spans per transaction reached). That means we'll need to change the `!tx.Sampled()` and `span.Dropped()` cases above. Take a look at apmhttp's client wrapper for reference: https://github.com/elastic/apm-agent-go/blob/2045b7b1014e9d7a6c4671d49323a818c488e067/module/apmhttp/client.go#L78

module/apmelasticsearch/client_test.go
```diff
@@ -303,6 +306,116 @@ func TestDestination(t *testing.T) {
...
+		_, err := client.Do(req.WithContext(ctx))
+		assert.NoError(t, err)
+	})
+
+	assert.Contains(t, headers, apmhttp.ElasticTraceparentHeader)
+	assert.Contains(t, headers, apmhttp.W3CTraceparentHeader)
+	assert.Contains(t, headers, apmhttp.TracestateHeader)
+}
+
+func TestClientSpanDropped(t *testing.T) {
```
stuartnelson3, line 332: @axw this and `TestClientTransactionUnsampled` are copied from `apmhttp/client_test.go`. They do verify that we're setting the headers when a span is dropped and when a transaction is not sampled, but what do you think about paring them back to just verifying we have the headers? I'm happy either way since this is essentially doubling up on checking the calculation of the headers, vs. just making sure we're setting them.

stuartnelson3:
jenkins run the tests

module/apmelasticsearch/client.go
```diff
@@ -63,17 +63,23 @@ type roundTripper struct {
 func (r *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	ctx := req.Context()
 	tx := apm.TransactionFromContext(ctx)
+	traceContext := tx.TraceContext()
 	if tx == nil || !tx.Sampled() {
+		apmhttp.SetHeaders(req, traceContext, false)
```
stuartnelson3, line 68: Should we default to propagating the legacy header?

axw: I think we can skip sending the legacy header. It's only for compatibility with old Elastic APM agents, so not relevant for the Elasticsearch use case where the plan is to extract the traceparent header and use the trace ID in logs.

module/apmhttp/client.go
```diff
@@ -133,7 +133,8 @@ func (r *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	return resp, err
 }
 
-func (r *roundTripper) setHeaders(req *http.Request, traceContext apm.TraceContext, propagateLegacyHeader bool) {
+// SetHeaders sets traceparent and tracestate headers on an http request.
```
axw, line 136: :+1:

module/apmelasticsearch/client_test.go
axw: I think what you have here is good. I do think it's important to check the trace context of the transaction vs. span is used like you have in this test.

module/apmelasticsearch/client.go
```diff
@@ -63,17 +63,23 @@ type roundTripper struct {
 func (r *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	ctx := req.Context()
 	tx := apm.TransactionFromContext(ctx)
+	traceContext := tx.TraceContext()
 	if tx == nil || !tx.Sampled() {
```
axw, line 67: I think we need to split the `if tx == nil` case out, and not call `apmhttp.SetHeaders` in that case?

axw (changes requested):
IIANM we should not call SetHeaders when there's no transaction in the context. Otherwise this LGTM

PR #1003, opened 2021-08-11
Infer destination service resource adapt public api

stuartnelson3:
- Add support for exit spans, auto-infer their `destination.service.resource`
- Remove `destination.service.{name,type}` from the public api and mark them as deprecated

closes https://github.com/elastic/apm-agent-go/issues/976

### DIFF
--- a/gocontext.go
+++ b/gocontext.go
@@ -90,17 +90,6 @@ func StartSpan(ctx context.Context, name, spanType string) (*Span, context.Conte
 	return StartSpanOptions(ctx, name, spanType, SpanOptions{})
 }
 
-// StartExitSpan starts an exit span. If the provided context already contains
-// an exit span, this function returns a dropped span.
-func StartExitSpan(ctx context.Context, name, spanType string) (*Span, context.Context) {
-	span, ctx := StartSpanOptions(ctx, name, spanType, SpanOptions{})
-	if span.Dropped() {
-		return span, ctx
-	}
-	span.setExitSpan(name, spanType)
-	return span, ctx
-}
-
 // StartSpanOptions starts and returns a new Span within the sampled transaction
 // and parent span in the context, if any. If the span isn't dropped, it will be
 // stored in the resulting context.
--- a/gocontext.go
+++ b/gocontext.go
@@ -97,14 +97,7 @@ func StartExitSpan(ctx context.Context, name, spanType string) (*Span, context.C
 	if span.Dropped() {
 		return span, ctx
 	}
-	span.exit = true
-	resource := spanType
-	if resource == "" {
-		resource = name
-	}
-	span.Context.SetDestinationService(DestinationServiceSpanContext{
-		Resource: resource,
-	})
+	span.setExitSpan(name, spanType)
 	return span, ctx
 }
 
--- a/gocontext.go
+++ b/gocontext.go
@@ -90,6 +90,24 @@ func StartSpan(ctx context.Context, name, spanType string) (*Span, context.Conte
 	return StartSpanOptions(ctx, name, spanType, SpanOptions{})
 }
 
+// StartExitSpan starts an exit span. If the provided context already contains
+// an exit span, this function returns a dropped span.
+func StartExitSpan(ctx context.Context, name, spanType string) (*Span, context.Context) {
+	span, ctx := StartSpanOptions(ctx, name, spanType, SpanOptions{})
+	if span.Dropped() {
+		return span, ctx
+	}
+	span.exit = true
+	resource := spanType
+	if resource == "" {
+		resource = name
+	}
+	span.Context.SetDestinationService(DestinationServiceSpanContext{
+		Resource: resource,
+	})
+	return span, ctx
+}
+
 // StartSpanOptions starts and returns a new Span within the sampled transaction
 // and parent span in the context, if any. If the span isn't dropped, it will be
 // stored in the resulting context.
--- a/model/model.go
+++ b/model/model.go
@@ -371,10 +371,10 @@ type DestinationSpanContext struct {
 // DestinationServiceSpanContext holds contextual information about a
 // destination service.
 type DestinationServiceSpanContext struct {
-	// Type holds the destination service type.
+	// Type holds the destination service type. Deprecated.
 	Type string `json:"type,omitempty"`
 
-	// Name holds the destination service name.
+	// Name holds the destination service name. Deprecated.
 	Name string `json:"name,omitempty"`
 
 	// Resource identifies the destination service
--- a/model/model.go
+++ b/model/model.go
@@ -375,7 +375,7 @@ type DestinationServiceSpanContext struct {
 	Type string `json:"type,omitempty"`
 
 	// Name holds the destination service name. Deprecated.
-	Name string `json:"name,omitempty"`
+	Name string `json:"name"`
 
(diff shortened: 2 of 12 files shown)


spancontext.go
```diff
@@ -57,10 +57,6 @@ type DatabaseSpanContext struct {
 
 // DestinationServiceSpanContext holds destination service span span.
 type DestinationServiceSpanContext struct {
-	// Name holds a name for the destination service, which may be used
-	// for grouping and labeling in service maps.
-	Name string
```
axw, line 62: We can't remove this, that would be a breaking API change. Just mark it as deprecated and don't require it to be set.

spancontext.go
```diff
@@ -222,10 +217,9 @@ func (c *SpanContext) SetMessage(message MessageSpanContext) {
 // Both service.Name and service.Resource are required. If either is empty,
 // then SetDestinationService is a no-op.
 func (c *SpanContext) SetDestinationService(service DestinationServiceSpanContext) {
-	if service.Name == "" || service.Resource == "" {
+	if service.Resource == "" {
 		return
 	}
-	c.destinationService.Name = truncateString(service.Name)
```
axw, line 228: We still need to send this field to APM Server, as older versions (<7.14) require it. Seeing as we don't actually make use of the value, I think if it's unset we could just set it to `service.Resource`. Alternatively, see my comment about omitempty on the model type

validation_test.go
```diff
@@ -119,12 +119,11 @@ func TestValidateDatabaseSpanContext(t *testing.T) {
 
 func TestValidateDestinationSpanContext(t *testing.T) {
 	overlong := strings.Repeat("x", 1025)
-	for _, name := range []string{"", overlong} {
+	for range []string{"", overlong} {
```
axw, line 122: I think we should revert the changes to this file, given that we'll need to keep the API stable

gocontext.go
```diff
@@ -90,6 +90,17 @@ func StartSpan(ctx context.Context, name, spanType string) (*Span, context.Conte
 	return StartSpanOptions(ctx, name, spanType, SpanOptions{})
 }
 
+// StartExitSpan starts an exit span. If the provided context already contains
+// an exit span, this function returns a dropped span.
```
axw, line 94: instead of creating a new method, how about we add a `ExitSpan bool` field to `apm.SpanOptions`?

model/model.go
```diff
@@ -371,10 +371,10 @@ type DestinationSpanContext struct {
 // DestinationServiceSpanContext holds contextual information about a
 // destination service.
 type DestinationServiceSpanContext struct {
-	// Type holds the destination service type.
+	// Type holds the destination service type. Deprecated.
 	Type string `json:"type,omitempty"`
 
-	// Name holds the destination service name.
+	// Name holds the destination service name. Deprecated.
 	Name string `json:"name,omitempty"`
```
axw, line 378: I forget whether we treat "" the same as omitted in apm-server, but maybe we can just remove "omitempty" from these two fields so we always send them as empty strings for backwards compatibility with older server versions that require the fields

span.go
```diff
@@ -64,7 +76,8 @@ func (tx *Transaction) StartSpan(name, spanType string, parent *Span) *Span {
 // span type, subtype, and action; a single dot separates span type and
 // subtype, and the action will not be set.
 func (tx *Transaction) StartSpanOptions(name, spanType string, opts SpanOptions) *Span {
-	if tx == nil {
+	if tx == nil || opts.parent.IsExitSpan() {
+		// TODO: Is this the equivalent of a noop span?
```
axw, line 80: Yes, "dropped spans" are no-op spans. They exist to remove the need to perform nil-pointer checks all over the place when creating spans, mostly because spans may be dropped due to the max span limit being reached

span.go
```diff
@@ -384,6 +400,30 @@ func (s *Span) ended() bool {
...
+func (s *Span) IsExitSpan() bool {
+	if s == nil || s.SpanData == nil {
+		return false
+	}
+	ctx := s.Context
+	return s.exit ||
+		ctx.destination != model.DestinationSpanContext{} ||
+		ctx.database != model.DatabaseSpanContext{} ||
+		ctx.message != model.MessageSpanContext{} ||
+		ctx.http != model.HTTPSpanContext{}
```
axw, line 424: Is this in the spec? I'd prefer not to second guess the "exit" field. If we do need to do this, then we'll need to lock the span -- s.SpanData can be modified concurrently or released back into a memory pool at any time.

span.go
```diff
@@ -392,6 +432,7 @@ type SpanData struct {
 	stackTraceLimit        int
 	timestamp              time.Time
 	childrenTimer          childrenTimer
+	exit                   bool
```
axw, line 435: SpanData is removed when the span is ended, so I think we'll need to move `exit` to the Span type as a sibling to the Span.TraceContext field. That way we can still tell if a span is/was an exit span even after it's ended.

axw (changes requested):

span.go
stuartnelson3: Yeah, spec is available here: https://github.com/elastic/apm/blob/master/specs/agents/tracing-spans.md#exit-spans. Will add the locking.

axw: Okay, thanks. That's kind of a pain, because it means that when you end a span it'll lose it's inferred exit span status. I think we'd better set `span.exit=true` when ending the span if any of those properties are set, so we can persist the fact after it's ended.

span.go
```diff
@@ -93,15 +81,20 @@ func (tx *Transaction) StartSpanOptions(name, spanType string, opts SpanOptions)
...
-		defer opts.parent.mu.Unlock()
 	}
 
 	// Prevent tx from being ended while we're starting a span.
 	tx.mu.RLock()
 	defer tx.mu.RUnlock()
 	if tx.ended() {
+		if opts.parent != nil {
+			opts.parent.mu.Unlock()
+		}
```
axw, line 92: Moving the unlock away from the lock scares me a bit. TBH I'm really not a fan of this implicit exit span status, so I've raised the question with agent devs whether it is necessary to implement that part of the spec.

stuartnelson3: :+1: makes sense. I'm not a fan of this locking either, so let's see what happens.

span.go
```diff
@@ -384,6 +398,25 @@ func (s *Span) ended() bool {
 	return s.SpanData == nil
 }
 
+func (s *Span) setExitSpan(name, spanType string) {
```
axw, line 401: Looking at the spec, I believe the resource is meant to be set to `subtype || type`.

Subtype and Type can be set after starting a span, so I think we need to do that when ending the span. What we can do is keep track of whether SpanContext.SetDestinationService has been called (even with empty strings), and if it is never called for an exit span, then do this implicit call when ending the exit span.

axw (changes requested):
Thanks, this is looking much better now. Just one issue now about timing of the implicit SetDestinationService call, and the values used.

span.go
```diff
@@ -384,6 +403,24 @@ func (s *Span) ended() bool {
 	return s.SpanData == nil
 }
 
+func (s *Span) setExitSpan() {
```
axw, line 406: `setExitSpanDestinationService` might be a bit clearer on the intent?

axw (approved):
thanks, looks great :)

Issue #885, opened 2021-01-29
[META 410] Instrumentation for Azure Storage


AlexanderWert:
Provide instrumentation for Azure Storage.

Meta issue: elastic/apm#410
SPEC: elastic/apm#411

stuartnelson3:
Based on this issue https://github.com/Azure/azure-sdk-for-go/issues/7278 I think we should handle both Tables Service and CosmosDB using just the cosmosdb sdk

stuartnelson3:
@AlexanderWert can you clarify which services this issue refers to? storage has several "sub" services, but queue received its own separate issue.

AlexanderWert:
@stuartnelson3 The relevant services are listed in the [spec here](https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-azure.md#azure-storage).

> but queue received its own separate issue
That's true. The issues were just created before the structure of the services was clear.

PR #1100, opened 2021-08-16
[azure] blob storage instrumentation

stuartnelson3:
Add instrumentation for azure's blob storage. This will probably change as additional services are instrumented and patterns/abstractions start to appear.

There is also queue, table, and file share storage. These will be added in separate PRs.

requested in #885


scripts/check_vet.sh
```diff
@@ -0,0 +1,15 @@
...
+
+for dir in $(scripts/moduledirs.sh); do
+    (
+      cd $dir
+      # Before go1.13, there was a bug that prevented the toolchain from
+      # finding a module when no buildable files were present.
+      # In go1.14 this is detected and the message states that build
+      # constraints exclude all Go files.
+      go list -json 2>&1 | grep -E 'build constraints exclude|cannot find module' > /dev/null
+      [ "$?" = "0" ] && echo "skipping $dir: no buildable Go files" && exit 0
```
stuartnelson3, line 12: @axw Unfortunately, the go vet documentation states:

>Vet's exit code is non-zero for erroneous invocation of the tool

which I interpret as "if you're running it in a dir with no go files, that's erroneous".

This is checking for the list of files from `go list`, and checking to see if it meets the "no buildable files" condition; if it does we just continue along.

Note: Technically, you can have buildable test files and no buildable source files, but that doesn't seem like a sane thing to ever happen.

stuartnelson3: the same failures start showing up in `make test`, sticking with adding a `doc.go` file for now

stuartnelson3:
this PR is using the incorrect sdk library; it should be using github.com/Azure/azure-storage-blob-go, and not github.com/Azure/azure-sdk-for-go/sdk/storage/armstorage.

A lot of the code here will transfer over, but have to be hooked in slightly differently, owing to the different API. I'll close this PR and open a new one.

PR #1104, opened 2021-08-18
[apmfasthttp] set correct http status

### LINKED ISSUE #1102: All transactions in fasthttp send 200 OK
All transactions are sent to APM with status code 200 and no headers are captured. 

Example code

```
package main

import (
	"github.com/valyala/fasthttp"
	"go.elastic.co/apm/module/apmfasthttp"
)

func handler(ctx *fasthttp.RequestCtx) {
	ctx.Error(fasthttp.StatusMessage(fasthttp.StatusUnauthorized), fasthttp.StatusUnauthorized)
}

func main() {
	fasthttp.ListenAndServe("127.0.0.1:8001", apmfasthttp.Wrap(handler))
}

```
Instead of sending 401, it is sending 200 for this transaction. 

In context.go setResponseContext always get 200 from statusCode := ctx.Response.Header.StatusCode()

stuartnelson3:
fixes #1102


module/apmfasthttp/server.go
```diff
@@ -93,6 +93,12 @@ func (h *apmHandler) handler(ctx *fasthttp.RequestCtx) {
 	}()
 
 	h.requestHandler(ctx)
+
+	if err := setResponseContext(ctx, tx, bc); err != nil {
+		return
+	}
+
+	tx.End()
```
simitt, line 101: Isn't it a problem that in the error case the transaction is not ended?

simitt (commented):
Can you add a comment to this PR about the effect of removing the logic you removed please.

module/apmfasthttp/server.go
axw: Also my understanding from the original PR is that there are cases (streaming) where the handler may return before the request is done. I was led to believe the closer would be invoked when the request completed. Is that not happening?

stuartnelson3: As far as I could see in the original code, the closer was not being invoked. When I inlined `setResponseContext`, I started getting the correct status code in my test case. When I saw that inlining the code "worked" (without knowing about early returns from streaming cases), `txCloser` looked unnecessary to me since it was never being used. But I guess that's what the code in `init` was supposed to accomplish?

stuartnelson3:
As far as I could see, the `txCloser` was not being invoked to set the response context, and I didn't realize it had another purpose (namely, that it was supposed to handle early returns from the handler), so when my test case reproducing the reported error started passing, I figured it was working.

module/apmfasthttp/server.go
axw: Not sure what `init` you're referring to, but my understanding is that the closer added by StartTransactionWithBody is meant to be invoked when the request ends:

https://pkg.go.dev/github.com/valyala/fasthttp#RequestCtx.SetUserValue

> All the values are removed from ctx after returning from the top RequestHandler. Additionally, Close method is called on each value implementing io.Closer before removing the value from ctx.

I suppose that's not happening as expected

stuartnelson3: The `init()` func: https://github.com/elastic/apm-agent-go/pull/1104/files#diff-68d60b1d06492503cf46894612cd31efc1d0e3ef6456a202af9a162e52f6e9b8L36-L54

>I suppose that's not happening as expected

According to my test case, it wasn't :shrug:

axw: I did a bit of testing and digging, and found that the closer is called but it's called after the response is reset - at which point the status code, headers, etc. are lost.

The reason for the closer is due to https://pkg.go.dev/github.com/valyala/fasthttp#RequestCtx.SetBodyStreamWriter. The handler may return but have set a streaming body writer; the request does not actually complete until after the body writer returns.

The docs for that method say:

> Access to RequestCtx and/or its' members is forbidden from sw.

So I think what we should do is:
 - set response context at the end of the handler wrapper
 - keep the closer, but use it only for ending the transaction and discarding the body capturer

I think we still need the `init` code. `fasthttp.RequestCtx` implements `context.Context` directly, with `Value` returning things set with `SetUserValue`: https://github.com/valyala/fasthttp/blob/v1.29.0/server.go#L2640

module/apmfasthttp/closer.go
```diff
@@ -21,28 +21,21 @@
...
-		ctx: ctx,
-		tx:  tx,
-		bc:  bc,
+		tx: tx,
+		bc: bc,
 	}
 }
 
 // Close sets the response context to the APM transaction and
 // ends the transaction.
```
axw, line 36: Close ends the transaction.

module/apmfasthttp/context.go
```diff
@@ -85,9 +87,7 @@ func setResponseContext(ctx *fasthttp.RequestCtx, tx *apm.Transaction, bc *apm.B
 	tx.Context.SetHTTPResponseHeaders(headers)
 	tx.Context.SetHTTPStatusCode(statusCode)
 
-	if bc != nil {
-		bc.Discard()
-	}
+	bc.Discard()
```
axw, line 90: Now that setResponseContext is called from within the handler, I'm not sure if this is too early or not. To be on the safe side, I think we should probably remove this and just call from the closer.

module/apmfasthttp/server.go
```diff
@@ -93,6 +92,10 @@ func (h *apmHandler) handler(ctx *fasthttp.RequestCtx) {
 	}()
 
 	h.requestHandler(ctx)
+
+	if err := setResponseContext(ctx, tx, bc); err != nil {
+		return
+	}
```
axw, line 98: AFAICS setResponseContext never returns an error, so can you please remove its err result?

module/apmfasthttp/closer.go
axw: also maybe worth adding a comment to the type to explain why it's needed at all

module/apmfasthttp/server.go
```diff
@@ -65,16 +65,15 @@ func Wrap(handler fasthttp.RequestHandler, options ...ServerOption) fasthttp.Req
...
 	}
 
-	tx, bc, err := StartTransactionWithBody(h.tracer, h.requestName(ctx), ctx)
+	tx, bc, err := StartTransactionWithBody(ctx, h.tracer, h.requestName(ctx))
 	if err != nil {
 		ctx.Error(err.Error(), fasthttp.StatusInternalServerError)
-
 		return
 	}
+	defer tx.End()
```
axw, line 76:
```suggestion
```

closer will be called when the response body is fully written, which I think is right in terms of measuring the request duration

module/apmfasthttp/server_test.go
```diff
@@ -0,0 +1,73 @@
...
+		Name:    "test-server",
+	}
+
+	go func() {
+		shutdown <- s.Serve(ln)
+	}()
+
+	addr := ln.Addr().String()
+	resp, err := http.Get("http://" + addr)
+	require.NoError(t, err)
```
axw, line 60:
```suggestion
	resp, err := http.Get("http://" + addr)
	require.NoError(t, err)
	io.Copy(ioutil.Discard, resp.Body) // consume body to complete request handler
	resp.Body.Close()
```

axw (approved):
LGTM! Thanks for adding the test and comments

PR #1105, opened 2021-08-19
[apmazure] add blob storage

stuartnelson3:
Add azure blob storage instrumentation.

I added instrumentation for azurequeue and moved the code into the interface seen in this PR, and then realized I was using the wrong azure lib. I decided to just bring over the interface from the other code for now.

There are still 3 other azure storage services to implement before #885 is officially "done":

- Table
- File Storage
- Queue


module/apmazure/storage.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/Azure/azure-pipeline-go/pipeline"
+
+	"go.elastic.co/apm"
+	"go.elastic.co/apm/module/apmhttp"
+	"go.elastic.co/apm/stacktrace"
+)
+
+func init() {
+	stacktrace.RegisterLibraryPackage(
+		"github.com/Azure/azure-storage-blob-go/azblob",
```
axw, line 39: also "github.com/Azure/azure-pipeline-go"?

module/apmazure/storage.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return p.next.Do(ctx, methodFactory, req)
+	}
+
+	tx := apm.TransactionFromContext(ctx)
+	if tx == nil {
+		return p.next.Do(ctx, methodFactory, req)
+	}
+
+	span := tx.StartSpan(rpc.name(), rpc._type(), apm.SpanFromContext(ctx))
```
axw, line 68:
```suggestion
	span := apm.StartSpan(ctx, rpc.name(), rpc._type())
```

If ctx doesn't contain a transaction, this will return a dropped span. If it does, it will also check if ctx contains a span and use that as the parent.

module/apmazure/storage.go
```diff
@@ -0,0 +1,128 @@
...
+		ctx = apm.ContextWithSpan(ctx, span)
+		req.Request = apmhttp.RequestWithContext(ctx, req.Request)
+		span.Context.SetHTTPRequest(req.Request)
+	} else {
+		return p.next.Do(ctx, methodFactory, req)
+	}
+	span.Action = rpc.operation()
+	span.Subtype = rpc.subtype()
+	span.Context.SetDestinationService(apm.DestinationServiceSpanContext{
+		Name:     rpc.subtype(),
```
axw, line 80:
```suggestion
```

Name doesn't need to be set anymore does it?

module/apmazure/storage.go
```diff
@@ -0,0 +1,128 @@
...
+		Resource: rpc.subtype() + "/" + rpc.storageAccountName(),
+	})
+
+	if host, port, err := net.SplitHostPort(req.URL.Host); err == nil {
+		// strconv.Atoi returns 0 if err != nil
+		// TODO: How do we want to handle an error?
+		p, _ := strconv.Atoi(port)
+		span.Context.SetDestinationAddress(host, p)
+	}
```
axw, line 89:
```suggestion
```

`span.Context.SetHTTPRequest` implicitly sets the destination address

module/apmazure/storage.go
```diff
@@ -0,0 +1,128 @@
...
+
+	resp, err := p.next.Do(ctx, methodFactory, req)
+	if err != nil {
+		apm.CaptureError(ctx, err).Send()
+	}
+
+	if r := resp.Response(); r != nil {
```
axw, line 96: this should probably only execute `if err == nil`

module/apmazure/storage.go
```diff
@@ -0,0 +1,128 @@
...
+	operation() string
+}
+
+func newAzureRPC(req pipeline.Request) (azureRPC, error) {
+	split := strings.Split(req.Host, ".")
+	accountName, storage := split[0], split[1]
+	var rpc azureRPC
+	if storage == "blob" {
+		rpc = &blobRPC{
+			resourceName: req.URL.Path[1:], // remove /
```
axw, line 118: maybe use `strings.TrimPrefix`, so we don't panic in case of API misuse?

module/apmazure/blob_test.go
```diff
@@ -0,0 +1,293 @@
...
+		blobURL.GetTags(ctx, nil)
+	})
+	require.Len(t, errors, 1)
+	require.Len(t, spans, 1)
+	span := spans[0]
+
+	assert.Equal(t, "storage", span.Type)
+	assert.Equal(t, "AzureBlob GetTags mycontainer/readme.txt", span.Name)
+	// TODO: If we use a fake URL, the test is fast but we do not set a
+	// status code
```
axw, line 61: why don't we set a status code? can we not use an httptest server?

axw (commented):
Looks good! Just a handful of minor things. I think we should probably use httptest.Server in the tests, but I don't know the details.

module/apmazure/storage.go
stuartnelson3: That was my understanding too, but the tracing doc has it marked as "required": https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-azure.md#span-context-fields. I raised the question in slack but didn't receive a response.

module/apmazure/storage.go
stuartnelson3: That was my understanding too, but the tracing doc has it marked as "required": https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-azure.md#span-context-fields. I raised the question in slack but didn't receive a response.

module/apmazure/blob_test.go
stuartnelson3: Because the request errors, `resp.Response()` is nil, so we can't get a status code. We could use something like `example.blob.core.windows.net` and get a 4XX response, but then the test takes a while longer (~1sec).

If we use httptest.Server, the URL is a local IP addr, but we rely on the subdomains in the provided url to get the account name and storage type (blob, queue, eg): `https://fakeaccnt.blob.core.windows.net`. We could use an httptest server and check the status code is set correctly, but then we won't be able to parse out (in this case) `fakeaccnt` and `blob` for other span properties.

module/apmazure/storage.go
axw: Couldn't find your thread on Slack - I've asked agent devs too now, but I think let's assume we'll remove it as the field isn't used anywhere.

module/apmazure/blob_test.go
axw: I see. What about setting [PipelineOptions.HTTPSender](https://github.com/Azure/azure-storage-blob-go/blob/da3335192ee9609890f4172c43d0750c65941b6f/azblob/zc_pipeline.go#L21-L22)?

stuartnelson3: good idea!

axw (approved):

PR #1109, opened 2021-08-24
[apmazure] file storage

stuartnelson3:
Add azure file storage instrumentation.

Table Service still needs to be implemented before #885 is officially "done".


axw (approved):
LGTM!

PR #1110, opened 2021-08-25
re-flush until transaction has been closed

stuartnelson3:
the transaction is ended after the response body
is fully written, so a single call to
`tracer.Flush()` may not have an event yet
enqueued.

## testing this change

before this change, I was able to get the test to reliably panic with `go test . -count=10000 -failfast`. over several runs on my local machine after this change, I have not gotten the panic.

### DIFF
--- a/module/apmfasthttp/server_test.go
+++ b/module/apmfasthttp/server_test.go
@@ -67,6 +67,11 @@ func TestServerHTTPResponse(t *testing.T) {
 	tracer.Flush(nil)
 	payloads := transport.Payloads()
 
+	// the transaction is ended after the response body is fully written,
+	// so a single call to `tracer.Flush()` may not have an event yet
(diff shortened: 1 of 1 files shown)


module/apmfasthttp/server_test.go
```diff
@@ -64,8 +65,24 @@ func TestServerHTTPResponse(t *testing.T) {
 	resp.Body.Close()
 	assert.Equal(t, fasthttp.StatusUnauthorized, resp.StatusCode)
 	tracer.Flush(nil)
-
 	payloads := transport.Payloads()
+
+	timer := time.NewTimer(100 * time.Millisecond)
```
axw, line 70: Can you please add a comment above this to explain why it's needed, and that it's particular to fasthttp? I wouldn't want anyone to cargo cult it

axw (approved):

PR #1132, opened 2021-09-28
transaction: Collect stats about dropped spans

### LINKED ISSUE #1113: [META 497] Collecting statistics about dropped spans
As described in [this spec](https://github.com/elastic/apm/blob/master/specs/agents/handling-huge-traces/tracing-spans-dropped-stats.md), agents should collect statistics on dropped spans.

Meta issue: [issue details](https://github.com/elastic/apm/issues/497)

marclop:
## Description
Modifies the span timing reporting function to collect statistics about
dropped spans when they exceed the `transaction_max_spans`. Each dropped
span is now aggregated under the `transaction.dropped_spans_stats` array
with a maximum count of items of `128`.

Each dropped span is aggregated on 4 dimensions: `outcome`. `subtype`,
`type` and `destination_service_resource` and report 2 metrics, their
`duration.sum.us` and the `count` for each of the aggregated dropped
spans.

## Related issues

Closes #1113

### DIFF
--- a/breakdown.go
+++ b/breakdown.go
@@ -64,6 +64,10 @@ Try to name your transactions so that there are less distinct transaction names.
 type spanTimingsKey struct {
 	spanType    string
 	spanSubtype string
+
+	// These fields are only set for dropped spans.
+	spanDstResource string
+	spanOutcome     string
 }
 
 // spanTiming records the number of times a {spanType, spanSubtype} pair
--- a/modelwriter.go
+++ b/modelwriter.go
@@ -18,6 +18,8 @@
 package apm // import "go.elastic.co/apm"
 
 import (
+	"time"
+
 	"go.elastic.co/apm/internal/ringbuffer"
 	"go.elastic.co/apm/model"
 	"go.elastic.co/apm/stacktrace"
@@ -122,6 +124,10 @@ func (w *modelWriter) buildModelTransaction(out *model.Transaction, tx *Transact
 	out.Duration = td.Duration.Seconds() * 1000
 	out.SpanCount.Started = td.spansCreated
 	out.SpanCount.Dropped = td.spansDropped
+	if dss := buildDroppedSpansStats(td.droppedSpansStats); len(dss) > 0 {
+		out.DroppedSpansStats = dss
+	}
+
 	if sampled {
 		out.Context = td.Context.build()
 	}
@@ -272,4 +278,23 @@ func normalizeOutcome(outcome string) string {
 	default:
 		return "unknown"
 	}
+}
+
+func buildDroppedSpansStats(dss spanTimingsMap) []model.DroppedSpansStats {
+	out := make([]model.DroppedSpansStats, 0, 128)
+	for k, timing := range dss {
+		out = append(out, model.DroppedSpansStats{
+			Type:                       k.spanType,
+			Subtype:                    k.spanSubtype,
+			DestinationServiceResource: k.spanDstResource,
+			Outcome:                    k.spanOutcome,
+			Duration: model.AggregateDuration{
+				Count: int(timing.count),
+				Sum: model.DurationSum{
+					Us: time.Duration(timing.duration).Microseconds(),
+				},
+			},
+		})
+	}
+	return out
 }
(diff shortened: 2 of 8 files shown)


marclop:
We may remove the "type" and "subtype" dimensions since they're not sure on the `apm-server` to produce the metrics. The PR that introduces the change is https://github.com/elastic/apm/pull/515.

https://github.com/elastic/apm-server/blob/9e0b8e32b63877765ab5996d926341fac2d32a94/x-pack/apm-server/aggregation/spanmetrics/aggregator.go#L255-L265

span.go
```diff
@@ -381,7 +399,15 @@ func (s *Span) reportSelfTime() {
 	} else {
 		s.tx.childrenTimer.childEnded(endTime)
 	}
-	s.tx.spanTimings.add(s.Type, s.Subtype, s.Duration-s.childrenTimer.finalDuration(endTime))
+
+	// Collect span timings when dropped and the destination service isn't empty
+	spanDuration := s.Duration - s.childrenTimer.finalDuration(endTime)
+	if s.dropped() && s.IsExitSpan() {
+		dstStvRes := s.Context.destinationService.Resource
```
axw, line 406:
```suggestion
		resource := s.Context.destinationService.Resource
```

nit: the other var name is unpronounceable to me :smile:

span.go
```diff
@@ -381,7 +399,15 @@ func (s *Span) reportSelfTime() {
 	} else {
 		s.tx.childrenTimer.childEnded(endTime)
 	}
-	s.tx.spanTimings.add(s.Type, s.Subtype, s.Duration-s.childrenTimer.finalDuration(endTime))
+
+	// Collect span timings when dropped and the destination service isn't empty
+	spanDuration := s.Duration - s.childrenTimer.finalDuration(endTime)
```
axw, line 404:
```suggestion
	selfTime := s.Duration - s.childrenTimer.finalDuration(endTime)
```

This isn't the entire span duration, but the "self time": the span duration minus the overlap with child spans. Is that the correct duration to be accumulating for dropped spans as well as breakdown metrics?

breakdown.go
```diff
@@ -66,6 +66,11 @@ type spanTimingsKey struct {
 	spanSubtype string
 }
 
+type droppedSpanTimingsKey struct {
+	destination string
+	outcome     string
+}
```
axw, line 72: This isn't related to breakdown metrics. I think it probably belongs next to `droppedSpanTimingsMap`, in transaction.go

modelwriter.go
```diff
@@ -273,3 +277,22 @@ func normalizeOutcome(outcome string) string {
 		return "unknown"
 	}
 }
+
+func buildDroppedSpansStats(dss droppedSpanTimingsMap) []model.DroppedSpansStats {
+	out := make([]model.DroppedSpansStats, 0, 128)
```
axw, line 282:
```suggestion
	out := make([]model.DroppedSpansStats, 0, len(dss))
```

transaction_test.go
```diff
@@ -381,6 +382,131 @@ func TestTransactionDiscard(t *testing.T) {
...
+	//   of the existing spans are still aggregated.
+	tx, spans, _ := apmtest.WithTransaction(func(ctx context.Context) {
+		exitSpanOpts := apm.SpanOptions{ExitSpan: true}
+		for i := 0; i < 1000; i++ {
+			span, _ := apm.StartSpanOptions(ctx,
+				fmt.Sprintf("GET %d", i),
+				fmt.Sprintf("request_%d", i),
+				exitSpanOpts,
+			)
+			<-time.After(10 * time.Microsecond)
```
axw, line 401: what are the sleeps for? can we just set span.Duration?

span.go
marclop: I'm not sure what I did here, I'll just use the `s.Duration` since that accurately reflects the span duration, and revert the childrenTimer calculation to what it was set to.

breakdown_test.go
```diff
@@ -510,7 +510,7 @@ func spanSelfTimeMetrics(txName, txType, spanType, spanSubtype string, count int
 		},
 		Samples: map[string]model.Metric{
 			"span.self_time.count":  {Value: float64(count)},
-			"span.self_time.sum.us": {Value: sum.Seconds() * 1000000},
+			"span.self_time.sum.us": {Value: float64(sum) / 1e3},
```
marclop, line 513: I had to change this since there was a case where `510000` turned into `510.00000000000006` due to the `time.Duration.Seconds()` logic. In any case, this should be the same as the other operation, perhaps a little more efficient, since we're only doing 1 division.

PR #1134, opened 2021-10-05
span (experimental): Compress short exit spans

### LINKED ISSUE #972: [META 432] Implement compressed spans algorithm
Implement the compressed spans algorithm in the agent as described in https://github.com/elastic/apm/issues/432.

marclop:
## Description

Implements the span compression algorithm described in elastic/apm#432.
The implementation pretty close to what the spec describes, with a few
modifications due to the differences in the Go agent implementation. The
span compression is an experimental and is disabled by default with the
default thresholds for the different compression strategies set:

- `ELASTIC_APM_SPAN_COMPRESSION_ENABLED=false`
- `ELASTIC_APM_SPAN_COMPRESSION_EXACT_MATCH_MAX_DURATION=50ms`
- `ELASTIC_APM_SPAN_COMPRESSION_SAME_KIND_MAX_DURATION=5ms`

The implementation uses a `spanBuffer` structure as the storage for span
compression candidates which needs to be present in both Transaction and
Span structures. The algorithm is described in detail with comments on
the code and is well documented there.

Additionally, two helper functions have been added under the `apmtest`:
package: `WriteTraceTable` and `WriteTraceWaterfall`, which help with
understanding why a test is failing if they are and debugging those.

I tried running the tests with higher counts to add pressure to the Go
runtime and have the spans take longer than they otherwise would and
using `apmtest.WriteTraceWaterfall` was extremely helpful.

```
$ go test -count=1000 -run TestCompressSpanSameKindTunedMax .
[             transaction (1757cdb325fd5f4e) - 16.243125ms             ]
m
[          request GET /a - 11.294375ms           ]
                                                 [4*Calls to request - ]
...
```

## Related issues
Closes #972

### DIFF
--- a/span.go
+++ b/span.go
@@ -253,6 +253,9 @@ type Span struct {
 	transactionID SpanID
 	parentID      SpanID
 	exit          bool
+	ctxPropagated bool
+	composite     compositeSpan
+	buffer        spanBuffer
 
 	mu sync.RWMutex
 
@@ -266,6 +269,10 @@ func (s *Span) TraceContext() TraceContext {
 	if s == nil {
 		return TraceContext{}
 	}
+	// TODO(marclop) maybe use atomic
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.ctxPropagated = true
 	return s.traceContext
 }
 
@@ -333,6 +340,12 @@ func (s *Span) End() {
 	if len(s.stacktrace) == 0 && s.Duration >= s.stackFramesMinDuration {
 		s.setStacktrace(1)
 	}
+
+	compressionEnabled := s.tracer.instrumentationConfig().spanCompressionEnabled
+	if s.attemptCompress(compressionEnabled) {
+		return
+	}
+
 	if s.tx != nil {
 		s.reportSelfTime()
 	}
@@ -374,6 +387,13 @@ func (s *Span) reportSelfTime() {
 
 	s.tx.TransactionData.mu.Lock()
 	defer s.tx.TransactionData.mu.Unlock()
+	s.reportSelfTimeLockless(endTime)
+}
+
+// reportSelfTimeLockless is used by all spans (composite and regular spans) to
+// report their timers. Since the composite end is triggered by a span of the
+// same kind or its parent ending, the upstream locks are already acquired.
+func (s *Span) reportSelfTimeLockless(endTime time.Time) {
 	if s.parent != nil {
 		if !s.parent.ended() {
 			s.parent.childrenTimer.childEnded(endTime)
--- a/span_compressed.go
+++ b/span_compressed.go
@@ -1,0 +1,321 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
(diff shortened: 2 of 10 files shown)


span_compressed.go
```diff
@@ -0,0 +1,321 @@
...
+	te := struct {
+		*Span
+		*SpanData
+	}{
+		Span:     from,
+		SpanData: from.SpanData,
+	}
+	to.span = te.Span
+	to.span.SpanData = te.SpanData
+}
```
axw, line 296: isn't this equivalent to `to.span = from`? what am I missing? :)

span_compressed.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+	if !b.span.tx.ended() && b.span.tx.breakdownMetricsEnabled {
+		b.span.reportSelfTimeLockless(b.span.timestamp.Add(b.span.Duration))
+	}
+
+	b.span.enqueue()
+	b.span = s
+	if s != nil {
+		preserveSpanData(s, b)
+	}
```
axw, line 283: why do we want to store s in b here? after flushing, shouldn't the buffer be emptied?

span_compressed.go
```diff
@@ -0,0 +1,321 @@
...
+		) {
+			return compressedStrategyExactMatch
+		}
+		return 0
+	}
+	if s.isSameKind(sibling) {
+		if s.durationLowerOrEq(sibling,
+			s.tx.compressedSpans.sameKindMaxDuration,
+		) {
+			s.Name = "Calls to " + s.Context.destinationService.Resource
```
axw, line 218: can this be deferred to the flush, so we just do it once for a set of compressed spans, and not for each one? (it'll incur an allocation)

span.go
```diff
@@ -374,6 +387,13 @@ func (s *Span) reportSelfTime() {
 
 	s.tx.TransactionData.mu.Lock()
 	defer s.tx.TransactionData.mu.Unlock()
+	s.reportSelfTimeLockless(endTime)
+}
+
+// reportSelfTimeLockless is used by all spans (composite and regular spans) to
+// report their timers. Since the composite end is triggered by a span of the
+// same kind or its parent ending, the upstream locks are already acquired.
```
axw, line 395: reportSelfTime also locks TransactionData to update its childrenTimer. Are we doing that in span.End()?

span.go
```diff
@@ -253,6 +253,9 @@ type Span struct {
 	transactionID SpanID
 	parentID      SpanID
 	exit          bool
+	ctxPropagated bool
+	composite     compositeSpan
+	buffer        spanBuffer
```
axw, line 258: IIANM these are only used in Span.End -- is that correct? If so, I think we should move them to SpanData. SpanData is discarded when the span is ended and its memory is recycled.

axw (commented):
The approach looks good. I haven't looked at all of the finer details, just left a few initial comments. I'm not entirely convinced by the locking, but again I haven't pored over all the details.

span.go
marclop: great catch, I totally missed that one, I'll review the locks again

span_compressed.go
marclop: yeah I had a note there to optimize but I must've deleted it, I'll change where we do this.

span_compressed.go
marclop: yeah... You're not missing anything, I'm not sure why I went to all this trouble to do this, I'll simplify things.

PR #1135, opened 2021-10-05
tracer: Remove `statsMu` in favor of `atomic`

marclop:
## Description
Removes the `statsMu sync.Mutex` and uses the `atomic` package to load
and obtain the values for the stats.

### DIFF
--- a/tracer.go
+++ b/tracer.go
@@ -327,8 +327,7 @@ type Tracer struct {
 	breakdownMetrics  *breakdownMetrics
 	profileSender     profileSender
 
-	statsMu sync.Mutex
-	stats   TracerStats
+	stats TracerStats
 
 	// instrumentationConfig_ must only be accessed and mutated
 	// using Tracer.instrumentationConfig() and Tracer.setInstrumentationConfig().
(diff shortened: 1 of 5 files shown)


tracer.go
```diff
@@ -327,8 +327,7 @@ type Tracer struct {
 	breakdownMetrics  *breakdownMetrics
 	profileSender     profileSender
 
-	statsMu sync.Mutex
-	stats   TracerStats
+	stats TracerStats
```
axw, line 330: We should probably heap allocate this, so ensure its fields are properly aligned for atomic access. Otherwise changes in fields above may cause it to misalign on 32-bit architectures (if it isn't already).

axw (changes requested):
Looks great, thank you. Just one issue regarding alignment on 32-bit arch.

tracer.go
marclop: Wasn't aware of the atomic bug, I've moved it to a pointer type in the Tracer.

tracer.go
```diff
@@ -327,8 +327,7 @@ type Tracer struct {
 	breakdownMetrics  *breakdownMetrics
 	profileSender     profileSender
 
-	statsMu sync.Mutex
-	stats   TracerStats
+	stats *TracerStats
```
axw, line 330:
```suggestion
	// stats is heap-allocated to ensure correct alignment for atomic access.
	stats *TracerStats
```

axw (approved):
LGTM, thanks :)

PR #1145, opened 2021-10-28
add histogram

stuartnelson3:
WIP for adding histogram support for prometheus.

The change to `Metric` is very much "make it work". I'd prefer to either add some sort of metric type identifier within the `Metric` type and some way to enforce validity within the type (`Buckets` and `Counts` should both be set and the correct length, `Value` shouldn't be set at the same time as `Buckets` and `Counts`), or some sort of metric type interface.

Example sample:

```
Samples:map[histogram:{Value:0 Buckets:[0.0025 0.0075 0.0175 0.037500000000000006 0.07500000000000001 0.175 0.375 0.75 1.75 3.75 5] Counts:[0 0 0 0 3 3 3 3 3 3 3]}]
```


model/marshal.go
```diff
@@ -670,3 +670,35 @@ func writeHex(w *fastjson.Writer, v []byte) {
...
 	}
 }
+
+// MarshalFastJSON writes the JSON representation of v to w.
+func (v *Metric) MarshalFastJSON(w *fastjson.Writer) error {
+	w.RawByte('{')
+	w.RawString("\"type\":")
+	w.String(v.Type)
+	if len(v.Counts) > 0 {
+		w.RawString(",\"buckets\":")
```
axw, line 680: apm-server expects it to be called "values": https://github.com/elastic/apm-server/blob/217b3841177a1ac3dc068a8fa982c871427a2273/docs/spec/v2/metricset.json#L49

I think it might be prudent to rename all "buckets" to "values" to keep aligned

module/apmprometheus/gatherer.go
```diff
@@ -80,8 +81,60 @@ func (g gatherer) GatherMetrics(ctx context.Context, out *apm.Metrics) error {
...
+					} else {
+						le = buckets[i-1].GetUpperBound() + (le-buckets[i-1].GetUpperBound())/2.0
+						counts[i] = b.GetCumulativeCount() - buckets[i-1].GetCumulativeCount()
+					}
+					// TODO: This rounds to 10,000ths, ie.
+					// 0.00001, precision, to handle
+					// floating point math trailing values,
+					// ie. 0.001500000003.
+					// How do we want to handle this?
+					midpoints[i] = math.Round(le*10000) / 10000
```
axw, line 133: I'm wondering if it's worth rounding/truncating at all. There might be legitimate use cases for indexing histograms with lower than 10000ths, and we'd be artificially excluding them. What problem are we solving with this? Just keeping the payload size down?

module/apmprometheus/gatherer.go
```diff
@@ -80,8 +81,60 @@ func (g gatherer) GatherMetrics(ctx context.Context, out *apm.Metrics) error {
...
+				h := m.GetHistogram()
+				// Total count for all buckets in this
+				// histogram. We want the per bucket count.
+				if h.GetSampleCount() == 0 {
+					continue
+				}
+				labels := makeLabels(m.GetLabel())
+				buckets := h.GetBucket()
+				midpoints := make([]float64, len(buckets))
+				counts := make([]uint64, len(buckets))
```
axw, line 105: We can (and I think should) omit zero count buckets.

axw (commented):
Looks good overall, just a few small things.

> Tests are failing because of my use of math.Round, which was introduced in go1.10. Do we still want/need to support versions this old? go1.9 was released in august 2017.

I don't want to, but yes we need to until we release a v2 of the agent.

There's an internal package with a round implementation for older versions of Go: https://github.com/elastic/apm-agent-go/tree/master/internal/apmmath

model/marshal.go
```diff
@@ -670,3 +670,35 @@ func writeHex(w *fastjson.Writer, v []byte) {
 		w.RawByte(hextable[v&0x0f])
 	}
 }
+
+// MarshalFastJSON writes the JSON representation of v to w.
+func (v *Metric) MarshalFastJSON(w *fastjson.Writer) error {
+	w.RawByte('{')
+	w.RawString("\"type\":")
+	w.String(v.Type)
```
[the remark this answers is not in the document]
stuartnelson3: makes sense. if it's empty then just write a raw `""`?

model/marshal.go
stuartnelson3: ah oops, thanks

module/apmprometheus/gatherer.go
stuartnelson3: When performing the midpoint math on buckets, I was getting bucket values like what's in the comment, like `0.001500000003`, which is probably not desired, and also might cause for flaky tests on different hardware (I think?).

module/apmprometheus/gatherer.go
stuartnelson3: Thinking about this .. won't this make it impossible to calculate percentiles? I thought the full range of available buckets was required in order to calculate a percentile over the values present

model/marshal.go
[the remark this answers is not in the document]
axw: I meant, I'd rather just omit the whole "type" property in this case. Something like:

```go
if len(v.Counts) > 0 {
    // add "values" and "counts"
} else {
    // add "value"
}
if v.Type != "" {
    // add "type"
}
```

module/apmprometheus/gatherer.go
axw: OK, I see. I think the fact that there's floating point error is fine and expected, but we should certainly avoid flaky tests. Could we find some bucket values which won't have this issue in tests?

stuartnelson3: Updated, hopefully CI is also happy

module/apmprometheus/gatherer.go
stuartnelson3: Nevermind, reviewed 8th grade statistics :sweat_smile:

module/apmprometheus/gatherer.go
```diff
@@ -80,8 +80,70 @@ func (g gatherer) GatherMetrics(ctx context.Context, out *apm.Metrics) error {
...
+					} else {
+						le = values[i-1].GetUpperBound() + (le-values[i-1].GetUpperBound())/2.0
+						// prometheus counts buckets cumulatively.
+						count = count - values[i-1].GetCumulativeCount()
+					}
+					// re-check count. the cumulative count
+					// in the initial check may have been
+					// non-zero, but when we subtract the
+					// preceding bucket, it may end up
+					// having a zero count.
```
axw, line 137: I wonder if there's enough value in the initial count check, wouldn't the code between be cheap? Maybe just  do the check once here?

module/apmprometheus/gatherer_test.go
```diff
@@ -134,6 +134,92 @@ func TestLabels(t *testing.T) {
...
+		},
+	}, {
+		Labels: model.StringMap{
+			{Key: "code", Value: "302"},
+			{Key: "method", Value: "GET"},
+		},
+		Samples: map[string]model.Metric{
+			"histogram": {
+				Type:   "histogram",
+				Values: []float64{7.5, 10},
```
axw, line 184: I would expect this to be `7.5, 12.5`. I think there's an issue with the translation of the final bucket. We should only take the 2nd last bucket value when the final bucket's upper bound is +Inf, i.e. all the values greater than the greatest specified upper bound (15 in this case).

module/apmprometheus/gatherer.go
```diff
@@ -80,8 +80,70 @@ func (g gatherer) GatherMetrics(ctx context.Context, out *apm.Metrics) error {
...
+					// prometheus buckets.
+					if count == 0 {
+						continue
+					}
+					le := b.GetUpperBound()
+					if i == 0 {
+						if le > 0 {
+							le /= 2
+						}
+					} else if i == (len(values) - 1) {
```
axw, line 117: Based on my other comment, I suppose the final +Inf bucket is omitted if it's empty? In which case we'd need to check `math.IsInf(values[i].GetUpperBound(), 1)`

axw (changes requested):
Code looks in good shape, seems to be an issue with the value for the last bucket.

module/apmprometheus/gatherer.go
stuartnelson3: that's fine by me. also, the first check is of limited use since it only works until we get a bucket with a count, after which point all buckets' cumulative counts will be initially non-zero.

module/apmprometheus/gatherer_test.go
stuartnelson3: Updated the code to reflect this, thanks for catching it!

module/apmprometheus/gatherer.go
stuartnelson3: I commented the code explaining the current situation as I found it while running test code. It appears the +Inf bucket isn't specifically included, but the histogram's total observations sum + count reflect values in the +Inf bucket, so it can be inferred.

module/apmprometheus/gatherer.go
```diff
@@ -80,8 +80,85 @@ func (g gatherer) GatherMetrics(ctx context.Context, out *apm.Metrics) error {
...
+					bucketCount += count
+					counts = append(counts, count)
+					midpoints = append(midpoints, le)
+				}
+				// Check if there were observations that fell
+				// outside of the defined histogram buckets, so
+				// we need to modify the current final bucket,
+				// and add an additional bucket with these
+				// observations.
+				if infBucketCount := totalCount - bucketCount; infBucketCount > 0 {
```
axw, line 153: Is it worth adding a `valuesLen > 0` check to this? I don't see why anyone would do that, but we probably shouldn't panic if they do.

axw (approved):
perfect!

module/apmprometheus/gatherer.go
stuartnelson3: I don't think that can happen, since you can't create a histogram with 0 buckets, but it doesn't hurt to check

https://github.com/prometheus/client_golang/blob/master/prometheus/histogram.go#L221-L223

PR #1151, opened 2021-11-04
feat(apmgrpc): wrap the server-stream with transaction-ctx

### LINKED ISSUE #1148: Improve gRPC Stream Interceptor
The `NewStreamServerInterceptor` creates a Transaction, but does not pass this transaction on to the handler.
This means that I can not create spans inside my gRPC handler and make them attached to the transaction created by the interceptor.

I'd like `apmgrpc` to wrap the gRPC server-stream with a struct that returns the `context.Contex` it created when creating the transaction, whenever the streams `.Context()` method is called. This is how it is done in many other stream-interceptors, such as in this logrus-interceptor: https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/logging/logrus/server_interceptors.go#L58-L61

Current workarounds:
- Drop using a streaming-interceptor, and add parsing of trace-headers from metadata and create transactions myself in every streaming call.
- Drop using apmgrpc's streaming-interceptor, but blatantly copy it and edit it to add the context-wrapping lines myself.

bendiktv2:
This allows the handler to retrieve the transaction from it's context, to e.g. create spans.

This closes #1148

### DIFF
--- a/module/apmgrpc/server.go
+++ b/module/apmgrpc/server.go
@@ -36,8 +36,6 @@ import (
 
 	"go.elastic.co/apm"
 	"go.elastic.co/apm/module/apmhttp"
-
-	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 )
 
 var (
@@ -136,8 +134,8 @@ func NewStreamServerInterceptor(o ...ServerOption) grpc.StreamServerInterceptor 
 		tx, ctx := startTransaction(ctx, opts.tracer, info.FullMethod)
 		defer tx.End()
 
-		wrapped := grpc_middleware.WrapServerStream(stream)
-		wrapped.WrappedContext = ctx
+		wrapped := wrapServerStream(stream)
+		wrapped.wrappedContext = ctx
 
 		// TODO(axw) define span context schema for RPC,
 		// including at least the peer address.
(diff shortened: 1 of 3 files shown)


stuartnelson3 (commented):
Looks good to me, but I'll wait for @axw for final approval. He'll be back next week.

module/apmgrpc/server.go
```diff
@@ -36,6 +36,8 @@ import (
 
 	"go.elastic.co/apm"
 	"go.elastic.co/apm/module/apmhttp"
+
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
```
axw, line 40: go-grpc-middleware is currently only used in tests, and I'd like to keep it that way to minimise dependencies added to instrumented applications.

Can you please remove it and introduce a type similar to WrappedServerStream (but unexported)?

axw (changes requested):
Thanks for the PR! The approach looks good, I'd just like to avoid the dependency please.

module/apmgrpc/server.go
bendiktv2: Sure, I used it, since it was already imported, but have now updated the PR without the import. I also removed the check for if the stream was already a wrappedServerStream, since it's no longer needed when it's not exported.

axw (approved):
Thank you :)

PR #1160, opened 2021-11-25
Set default outcome according to spec

axw:
Apart from HTTP and gRPC, which set outcome based on the status code, set the default outcome to "success" by default, or "failure" if any errors were captured within the transaction or span.

This brings us into line with the agent spec for transaction and span outcome:
 - https://github.com/elastic/apm/blob/master/specs/agents/tracing-transactions.md#transaction-outcome
 - https://github.com/elastic/apm/blob/master/specs/agents/tracing-spans.md#span-outcome

> For other protocols, we can default to the following behavior:
>
> - failure when an error is reported
> - success otherwise

### DIFF
--- a/span.go
+++ b/span.go
@@ -293,6 +293,8 @@ func (s *Span) SetStacktrace(skip int) {
 	if s.ended() {
 		return
 	}
+	s.SpanData.mu.Lock()
+	defer s.SpanData.mu.Unlock()
 	s.SpanData.setStacktrace(skip + 1)
 }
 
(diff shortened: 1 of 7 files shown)


span.go
```diff
@@ -346,6 +348,13 @@ func (s *Span) End() {
 	}
 	if s.Outcome == "" {
 		s.Outcome = s.Context.outcome()
+		if s.Outcome == "" {
+			if s.errorCaptured {
```
marclop, line 352: nice, this access doesn't need to be coordinated because the writer acquires a `s.mu.Rlock` and modifications won't be allowed to happen since this function holds `s.mu.Lock`

marclop (approved):
Looks great, very cool use of read locks and locks

span.go
axw: Yup :) Maybe it would be worth commenting on the locks when they should be held. The SpanData/TransactionData locks only need to be held when holding an RLock on the Span/Transaction lock.

PR #1164, opened 2021-12-06
Increase test pipeline timeout

cachedout:
The original timeout was set three years ago. Since that time, test times have gradually been moving up:

Screen Shot 2021-12-06 at 1 51 01 PM

This isn't necessarily due to changes in the time the tests take. Instead, the time it takes to provision workers and prepare them can vary considerably.

Therefore, I'm increasing the timeout for tests from 1 hour to 2 hours just to prevent the recent failures that we've seen where timeouts have been exceeded, causing the test suite not to report a result.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,7 +20,7 @@ pipeline {
     SLACK_CHANNEL = '#apm-agent-go'
   }
   options {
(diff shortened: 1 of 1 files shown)


axw (approved):
thanks!

v1v (approved):

PR #1167, opened 2021-12-08
transaction: Stop recording breakdown metrics

### LINKED ISSUE #1139: [META 505] Stop recording transaction.duration.sum.us,count metrics
Implement https://github.com/elastic/apm/issues/505

marclop:
## Description
Stops recording the unused transaction breakdown metrics:
* `transaction.duration.{count,sum.us}`
* `transaction.breakdown.count`

Removes the 4 bytes of padding in `breakdownTiming` and changes the type
of `spanTiming.count` from `uintptr` to `uint64`, making 32 and 64 bit
architectures span counts equal in size, removing the need for padding.

## Related issues

Closes #1139


breakdown.go
```diff
@@ -116,9 +113,9 @@ func newBreakdownMetrics() *breakdownMetrics {
 
 type breakdownMetricsMap struct {
 	mu      sync.RWMutex
-	entries int
 	m       map[uint64][]*breakdownMetricsMapEntry
 	space   []breakdownMetricsMapEntry
+	entries int
```
marclop, line 118: This one saves us 8 bytes of padding, not a lot

breakdown.go
```diff
@@ -129,8 +126,8 @@ func newBreakdownMetricsMap() *breakdownMetricsMap {
 }
 
 type breakdownMetricsMapEntry struct {
-	breakdownTiming
 	breakdownMetricsKey
+	breakdownTiming
```
marclop, line 130: Moved this here to align the struct better and reduce its size, I thought it worth pursuing since it may have up to 1000 entries in memory:

```
struct with 80 pointer bytes could be 56
```

Doing so removes 24 bytes of padding per entry, which can amount up to 24Kbytes of savings, not huge but less memory usage nevertheless.

breakdown_internal_test.go
```diff
@@ -86,7 +86,7 @@ func testBreakdownMetricsAlignment(t *testing.T, arch string) {
 		breakdownMetricsMapEntryObj.Type(), false, pkg, "breakdownTiming",
 	)
 	require.NotNil(t, breakdownTimingObj)
-	assert.Equal(t, []int{0}, breakdownTimingFieldIndex)
+	assert.Equal(t, []int{1}, breakdownTimingFieldIndex)
```
marclop, line 89: Had to change the index of the field for the test to pass since I've reordered them.

breakdown.go
axw: nice :)

axw (approved):
LGTM!

breakdown.go
```diff
@@ -156,27 +153,16 @@ func (k breakdownMetricsKey) hash() uint64 {
...
-	_ pad32
-
 	// span holds the "span.self_time" metric values.
 	span spanTiming
 
-	// breakdownCount records the number of transactions for which we
-	// have calculated breakdown metrics. If breakdown metrics are
-	// enabled, this will be equal transaction.count.
-	breakdownCount uintptr
+	// Padding to ensure the span field below is 64-bit aligned.
```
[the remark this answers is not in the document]
marclop: I tried removing the padding, but on 32-bit architecture the `uintptr` is 4 bytes rather than 8 (64-bit), making the total struct size 12 bytes in 32-bit architectures instead of the 16 that we get by default on 32-bit architectures.

We could arguably change the implementation to always use an `uint64`, which gives us a maximum value of `18,446,744,073,709,551,615` which is quite large already, not sure what benefit `uintptr` would give us since we're not using it for pointer references anyway.

I think switching to uint64 makes sense since instead of allocating padding bytes in memory, we'd use them for the actual count rather than having padding on 32-bit architectures, I'll make the change, but re-request a review from you to make sure we're ok with `uint64` vs `uintptr`.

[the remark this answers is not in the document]
axw: Looks sensible, and simpler - thanks.

axw (approved):

axw:
I don't think this can usefully be tested manually, but I'm open to ideas. The server doesn't record these metrics any more.

PR #1168, opened 2021-12-08
transport: Introduce NewHTTPTransportOptions

marclop:
## Description
Introduces a new `transport.NewHTTPTransportOptions()` function which
accepts an Options struct to be able to customize the `httptransport`
without using environment variables


transport/http.go
```diff
@@ -71,6 +71,37 @@ var (
...
+type HTTPTransportOptions struct {
+	// APIKey holds the APIKey base64-encoded string.
+	APIKey string
+
+	// SecretToken holds the secret token configured in the APM Server.
+	SecretToken string
+
+	// ServerTLSCert can be set if you have configured your APM Server with a
+	// self signed TLS certificate, or you want to pin the server certificate,
+	// specify the path to the PEM-encoded certificate with `ServerCACert`.
```
axw, line 84: I'm having a hard time parsing this sentence. What's ServerCACert got to do with this? If the server certificate is being pinned, then there's no point in setting the CA cert as well is there?

Also, naming quibble: for consistency either this one should be ServerCert or the other one should be ServerTLSCACert. They're both to do with TLS as much as the other. I'd probably go with ServerCert, seeing as the env var is `ELASTIC_APM_SERVER_CERT`.

transport/http.go
```diff
@@ -71,6 +71,37 @@ var (
...
+	ServerTLSCert string
+
+	// ServerCACert can be set if you want to specify a path to the pem-encoded
```
axw, line 87:
```suggestion
	// ServerCACert can be set if you want to specify a path to the PEM-encoded
```

transport/http.go
```diff
@@ -144,30 +190,28 @@ func NewHTTPTransport() (*HTTPTransport, error) {
...
 		if !rootCAs.AppendCertsFromPEM(additionalCerts) {
-			return nil, fmt.Errorf("failed to load CA certs from %s", caCertPath)
+			return nil, fmt.Errorf("failed to load CA certs from %s", opts.ServerCACert)
 		}
 		tlsConfig.RootCAs = rootCAs
 	}
 
 	client := &http.Client{
-		Timeout: serverTimeout,
+		Timeout: opts.ServerTimeout,
```
axw, line 206: Should NewHTTPTransportOptions return an error if ServerTimeout is invalid (negative)? I feel like it probably should, to avoid surprising errors at runtime.

transport/http.go
```diff
@@ -187,12 +231,12 @@ func NewHTTPTransport() (*HTTPTransport, error) {
...
-		t.SetAPIKey(apiKey)
-	} else if secretToken := os.Getenv(envSecretToken); secretToken != "" {
-		t.SetSecretToken(secretToken)
+	if opts.APIKey != "" {
+		t.SetAPIKey(opts.APIKey)
+	} else if opts.SecretToken != "" {
+		t.SetSecretToken(opts.SecretToken)
 	}
-	t.SetServerURL(serverURLs...)
+	t.SetServerURL(opts.ServerURLs...)
```
axw, line 239: IIANM this will panic if HTTPTransportOptions.ServerURLs is empty. This should probably also return an error.

Maybe worth adding a Validate method, like:

```go
func (o HTTPTransportOptions) Validate() error {
    if len(o.ServerURLs) == 0 {
        // ...
    }
    if len(o.ServerTimeout) < 0 {
        // ...
    }
    return nil
}
```

axw (changes requested):

transport/http.go
```diff
@@ -116,26 +147,41 @@ func NewHTTPTransport() (*HTTPTransport, error) {
...
+		opts.APIKey = apiKey
+	} else if secretToken := os.Getenv(envSecretToken); secretToken != "" {
+		opts.SecretToken = secretToken
+	}
+	return NewHTTPTransportOptions(opts)
+}
+
+// NewHTTPTransportOptions returns a customized HTTPTransport which can be used
+// for streaming data to the APM Server.
+func NewHTTPTransportOptions(opts HTTPTransportOptions) (*HTTPTransport, error) {
```
simitt, line 179: How about `NewHttpTransportFromOptions`? When reading `NewHTTPTransportOptions` I'd expect a `HTTPTransportOptions` struct to be returned.

simitt (commented):
In my opinion `NewHTTPTransport` and `NewHTTPTransportOptions` should deliever the same default values for non-customized options, so also ensure that the same default values for `ServerURLs` and `ServerTimeout` are set. 
Can you please add a test for when using `NewHTTPTransportOptions`?

transport/http.go
axw: For better or worse, the naming is consistent with other existing functions like `NewTracerOptions`. I'd prefer to keep them consistent even if it's not the most clear language.

axw (changes requested):
LGTM, just a couple of minor things

axw (approved):

PR #1183, opened 2022-01-05
tracestate: Allow duplicate keys but rewrite them

### LINKED ISSUE #1182: Update `tracestate` parsing rules to comply with latest W3C spec
## Description

The latest revision of the W3C trace-context RFC (https://github.com/w3c/trace-context/commit/0c3135d8c2d18f51589098abac94083b3088dcb6) includes some changes state:

```
The test_tracestate_duplicated_keys was broken, it demanded to drop
the whole tracestate when a duplicate key was found. This does not match
what the specification mandates in that situation. Instead,
implementations MAY remove the duplicates from tracestate or even leave
them in as-is.
```

Hence, we need to modify the tracestate validation rules to not drop the tracestate, but rather remove the first instance of a duplicate key when found. https://github.com/elastic/apm-agent-go/blob/7e472e42beb69748652ff672e13b42afbfdfbbcf/tracecontext.go#L272-L274

marclop:
## Description

Changes the validation logic to allow duplicate keys to be present in
tracestate entries, modifying the `NewTraceState` constructor to rewrite
the duplicate keys according to the new W3C spec.

## Related issues

Closes #1182

### DIFF
--- a/tracecontext.go
+++ b/tracecontext.go
@@ -169,10 +169,34 @@ type TraceState struct {
 
 // NewTraceState returns a TraceState based on entries.
 func NewTraceState(entries ...TraceStateEntry) TraceState {
-	out := TraceState{}
+	var out TraceState
 	var last *TraceStateEntry
+	var modified []TraceStateEntry
+	recorded := make(map[string]uint8)
+	// Range over the entries and find duplicate keys. When a previous key is
+	// found, it is stored in a map with a count that is incremented every time
+	// the key is observed.
+	// When keys are observed more than once, the last value observed is stored
+	// in a slice of the modified keys.
 	for _, e := range entries {
-		e := e // copy
+		if count, ok := recorded[e.Key]; ok {
+			if count > 1 {
+				for i, m := range modified {
+					if m.Key == e.Key {
+						modified[i] = e
+					}
+				}
+			} else {
+				modified = append(modified, e)
+			}
+		}
+		recorded[e.Key]++
+	}
+	// After we've ranged over the whole entry slice, any duplicates present
+	// must be written first in a last to first order, thus we range the slice
+	// backwards.
+	for i := len(modified) - 1; i >= 0; i-- {
+		e := modified[i] // copy
 		if last == nil {
 			out.head = &e
 		} else {
@@ -180,12 +204,26 @@ func NewTraceState(entries ...TraceStateEntry) TraceState {
 		}
 		last = &e
 	}
+	// Now, we can write the rest of the entries in the linked list ignoring
+	// any keys that have already been handled.
+	var parsedESTraceState bool
 	for _, e := range entries {
-		if e.Key != elasticTracestateVendorKey {
+		// If the tracestate entry key has been recorded more than 1 time, it
+		// has already been written to the linked list.
+		if count, ok := recorded[e.Key]; ok && count > 1 {
 			continue
 		}
-		out.parseElasticTracestateError = out.parseElasticTracestate(e)
-		break
+		e := e // copy
+		if last == nil {
+			out.head = &e
+		} else {
+			last.next = &e
+		}
+		last = &e
+		if !parsedESTraceState && e.Key == elasticTracestateVendorKey {
+			out.parseElasticTracestateError = out.parseElasticTracestate(e)
+			parsedESTraceState = true
+		}
 	}
 	return out
 }
@@ -245,14 +283,12 @@ func (s TraceState) String() string {
 
 // Validate validates the trace state.
 //
-// This will return non-nil if any entries are invalid,
-// if there are too many entries, or if an entry key is
-// repeated.
+// This will return non-nil if any entries are invalid or
+// if there are too many entries.
 func (s TraceState) Validate() error {
 	if s.head == nil {
 		return nil
 	}
-	recorded := make(map[string]int)
 	var i int
 	for e := s.head; e != nil; e = e.next {
 		if i == 32 {
@@ -269,10 +305,6 @@ func (s TraceState) Validate() error {
 				return errors.Wrapf(err, "invalid tracestate entry at position %d", i)
 			}
 		}
-		if prev, ok := recorded[e.Key]; ok {
-			return fmt.Errorf("duplicate tracestate key %q at positions %d and %d", e.Key, prev, i)
-		}
-		recorded[e.Key] = i
 		i++
 	}
 	return nil
--- a/tracecontext_test.go
+++ b/tracecontext_test.go
@@ -76,9 +76,15 @@ func TestTraceStateDuplicateKey(t *testing.T) {
 		apm.TraceStateEntry{Key: "x", Value: "b"},
 		apm.TraceStateEntry{Key: "a", Value: "b"},
 		apm.TraceStateEntry{Key: "y", Value: "b"},
-		apm.TraceStateEntry{Key: "a", Value: "b"},
+		apm.TraceStateEntry{Key: "a", Value: "c"},
+		apm.TraceStateEntry{Key: "z", Value: "w"},
+		apm.TraceStateEntry{Key: "a", Value: "d"},
+		apm.TraceStateEntry{Key: "c", Value: "first"},
+		apm.TraceStateEntry{Key: "r", Value: "first"},
+		apm.TraceStateEntry{Key: "c", Value: "second"},
 	)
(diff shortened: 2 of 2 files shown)


tracecontext.go
```diff
@@ -169,23 +169,61 @@ type TraceState struct {
 
 // NewTraceState returns a TraceState based on entries.
 func NewTraceState(entries ...TraceStateEntry) TraceState {
-	out := TraceState{}
+	var out TraceState
 	var last *TraceStateEntry
+	var modified []TraceStateEntry
+	recorded := make(map[string]uint8)
```
marclop, line 175: This could also be named `seen` or `observed`, I think it works either way but I'm happy to rename it.

tracecontext.go
```diff
@@ -169,23 +169,61 @@ type TraceState struct {
...
+			} else {
+				modified = append(modified, e)
+			}
+		}
+		recorded[e.Key]++
+	}
+	// After we've ranged over the whole entry slice, any duplicates present
+	// must be written first in a last to first order, thus we range the slice
+	// backwards.
+	for i := len(modified) - 1; i >= 0; i-- {
```
marclop, line 198: I chose to use two different slices to avoid mutating the slice that is passed down the call stack, since it may not be expected or desired for that slice to be mutated.

tracecontext_test.go
```diff
@@ -76,9 +76,15 @@ func TestTraceStateDuplicateKey(t *testing.T) {
...
+		apm.TraceStateEntry{Key: "a", Value: "c"},
+		apm.TraceStateEntry{Key: "z", Value: "w"},
+		apm.TraceStateEntry{Key: "a", Value: "d"},
+		apm.TraceStateEntry{Key: "c", Value: "first"},
+		apm.TraceStateEntry{Key: "r", Value: "first"},
+		apm.TraceStateEntry{Key: "c", Value: "second"},
 	)
-	assert.EqualError(t, ts.Validate(), `duplicate tracestate key "a" at positions 1 and 3`)
+	assert.NoError(t, ts.Validate())
+	assert.Equal(t, "c=second,a=d,x=b,y=b,z=w,r=first", ts.String())
```
marclop, line 87: This test should cover all scenarios.

tracecontext.go
```diff
@@ -169,23 +169,61 @@ type TraceState struct {
...
+	// the key is observed.
+	// When keys are observed more than once, the last value observed is stored
+	// in a slice of the modified keys.
 	for _, e := range entries {
-		e := e // copy
+		if count, ok := recorded[e.Key]; ok {
+			if count > 1 {
+				for i, m := range modified {
+					if m.Key == e.Key {
+						modified[i] = e
```
stuartnelson3, line 186: Once we've found a matching key, we could `continue`, right? Since we're only wanting to record the latest entry `e` for a specific key `e.Key`?

tracecontext.go
```diff
@@ -169,23 +169,61 @@ type TraceState struct {
...
+					}
+				}
+			} else {
+				modified = append(modified, e)
+			}
+		}
+		recorded[e.Key]++
+	}
+	// After we've ranged over the whole entry slice, any duplicates present
+	// must be written first in a last to first order, thus we range the slice
```
stuartnelson3, line 196: nit: `must be written in a last to first order`, I think it makes the comment a bit simpler since we're already saying `last` and `first`

stuartnelson3 (approved):
couple small comments, but nothing that requires a re-review

tracecontext.go
axw, line 175: WDYT about not checking duplicates for non-Elastic keys? We don't really care about them. If we're not required to check for duplicates I think it might be better to only check our own keys, as we can do it without any allocations: just keep track of whether we've seen an Elastic key already.

marclop: that's probably better than treating all the keys the same, since each system should do their own key _cleaning_. I'll change the implementation and simplify the logic.

tracecontext.go
```diff
@@ -169,23 +169,61 @@ type TraceState struct {
...
+	// backwards.
```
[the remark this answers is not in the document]
marclop: exactly!

tracecontext.go
marclop: yeah we could if we kept the implementation as is, I'm going to simplify it and change it to only act on the 'es' vendor key which is the only one we care about, it'll also eliminate allocations :)

tracecontext.go
```diff
@@ -169,9 +169,40 @@ type TraceState struct {
 
 // NewTraceState returns a TraceState based on entries.
 func NewTraceState(entries ...TraceStateEntry) TraceState {
-	out := TraceState{}
+	var out TraceState
 	var last *TraceStateEntry
+	var recorded uint8
+	// Range over the entries and find duplicate `es` vendor keys. When 'es'
```
axw, line 175: I'm finding this method quite difficult to follow now. Would this alternative work?

```go
var haveElastic bool
var last *TraceState
for _, e := range entries {
	if e.Key == elasticTracestateVendorKey {
		if haveElastic {
			// Discard duplicate "es" entries; keep the last entry's value.
			out.head.value = e.value
			continue
		}
		haveElastic = true
		e.next = last
		last = nil // cause elastic to be moved to head
	}
	e := e // copy
	if last == nil {
		out.head = &e
	} else {
		last.next = &e
	}
	last = &e
}
if haveElastic {
	out.parseElasticTracestateError = out.parseElasticTracestate(out.head)
}
```

axw: BTW we didn't previously move the elastic entry to the head because IIANM that's only expected when _mutating_ tracestate.

marclop: I mostly kept the suggestion intact, but had to make some changes in 9270b6a to not lose any entries that were found previous to the `es` key.

tracecontext.go
```diff
@@ -171,37 +171,24 @@ type TraceState struct {
...
-	}
+	var haveElastic bool
 	for _, e := range entries {
 		if e.Key == elasticTracestateVendorKey {
-			if !parsedESTraceState {
-				out.parseElasticTracestateError = out.parseElasticTracestate(e)
-				parsedESTraceState = true
-			} else {
+			if haveElastic {
+				// Discard duplicate `es`` entries; keep the last entry's value.
```
axw, line 178:
```suggestion
				// Discard duplicate `es` entries; keep the last entry's value.
```

tracecontext.go
```diff
@@ -171,37 +171,24 @@ type TraceState struct {
...
+				out.head.Value = e.Value
 				continue
 			}
+			haveElastic = true
+			e := e            // copy
+			e.next = out.head // move the current head reference to `es`.next.
+			out.head = &e     // swap the head with the current `es` entry.
+			// To preserve the previous entries int he linked list, set the
```
axw, line 186:
```suggestion
			// To preserve the previous entries in the linked list, set the
```

axw (approved):
thanks for the updates!

axw:
Verified using go.elastic.co/apm/v2 v2.0.0-20220209134054-640a916049c9 (current main), along with apmhttp/v2. To get apmhttp/v2 to work, I had to add a replace stanza:

`replace go.elastic.co/apm/v2 v2.0.0 => go.elastic.co/apm/v2 v2.0.0-20220209134054-640a916049c9`

I compared the v2 behaviour against the latest v1 agent.

```
package main

import (
        "fmt"
        "io"
        "io/ioutil"
        "net/http"
        "net/http/httptest"

        //"go.elastic.co/apm"
        //"go.elastic.co/apm/module/apmhttp"
        "go.elastic.co/apm/module/apmhttp/v2"
        "go.elastic.co/apm/v2"
)

func main() {
        srv := httptest.NewServer(apmhttp.Wrap(http.HandlerFunc(
                func(w http.ResponseWriter, r *http.Request) {
                        tx := apm.TransactionFromContext(r.Context())
                        fmt.Println(tx.TraceContext().State)
                },
        )))
        req, _ := http.NewRequest("GET", srv.URL, nil)
        req.Header.Set("Tracestate", "foo=bar,foo=bar")
        req.Header.Set("Traceparent", "00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01")
        resp, _ := http.DefaultClient.Do(req)

        io.Copy(ioutil.Discard, resp.Body)
        resp.Body.Close()
}
```

- With v1, `tracestate` is cleared out.
- With v2, `tracestate` is maintained as-is.

PR #1187, opened 2022-01-11
remove stacktrace.ContextSetter

### LINKED ISSUE #646: Remove stacktrace.ContextSetter
It's currently possible to set stacktrace "context" — surrounding lines of source code — by supplying a stacktrace.ContextSetter to Tracer. This is almost certainly not being used, and will become irrelevant once we have integration with Code Search. We should remove the ContextSetter bits from the agent in a future major version.

stuartnelson3:
closes #646


modelwriter.go
```diff
@@ -262,19 +259,6 @@ func stacktraceCulprit(frames []model.StacktraceFrame) string {
...
-func (w *modelWriter) setStacktraceContext(stack []model.StacktraceFrame) {
-	if w.cfg.contextSetter == nil || len(stack) == 0 {
-		return
-	}
-	err := stacktrace.SetContext(w.cfg.contextSetter, stack, w.cfg.preContext, w.cfg.postContext)
-	if err != nil {
-		if w.cfg.logger != nil {
-			w.cfg.logger.Debugf("setting context failed: %v", err)
-		}
-		w.stats.Errors.SetContext++
```
axw, line 274: can we get rid of the SetContext field now?

tracer.go
```diff
@@ -525,7 +524,6 @@ type tracerConfig struct {
 	metricsInterval         time.Duration
 	logger                  WarningLogger
 	metricsGatherers        []MetricsGatherer
-	contextSetter           stacktrace.ContextSetter
 	preContext, postContext int
```
axw, line 527: Do we still need preContext/postContext? I think we can get rid of all related config and env vars?

axw (commented):
Thanks, looks good. I think there might be a few more related things we can get rid of.

tracer.go
stuartnelson3: seems like it, removed :magic_wand:

axw (approved):
:tada: LGTM!

stuartnelson3:
I didn't see any additional environment variables, but I cleaned up the pre/post context stuff, and code related to Errors.SetContext

simitt:
@axw anything you think we should test for this? Nothing comes to my mind, so I'd skip manual testing on this one.

axw:
Nope, nothing to test here, we just took away an API.

Issue #465, opened 2019-03-06
Add special case value to disable ELASTIC_APM_SPAN_FRAMES_MIN_DURATION


axw:
Introduce a means of disabling capturing stack traces, by setting ELASTIC_APM_SPAN_FRAMES_MIN_DURATION to a special value of 0.

The other agents interpret a value of 0 to mean disable capturing stack traces, and -1 to always capture. We should do the same.

axw:
See https://github.com/elastic/apm/issues/94

PR #1188, opened 2022-01-11
handle special cases for span frames min duration

stuartnelson3:
cases:

0: never capture stacktraces
-1: always capture stacktraces

Closes https://github.com/elastic/apm-agent-go/issues/465

### DIFF
--- a/span.go
+++ b/span.go
@@ -356,9 +356,17 @@ func (s *Span) End() {
 			}
 		}
 	}
-	if !s.dropped() && len(s.stacktrace) == 0 &&
-		s.Duration >= s.stackFramesMinDuration {
+	switch s.stackFramesMinDuration {
+	case -1:
+		// Always set stacktrace
 		s.setStacktrace(1)
+	case 0:
+		// If s.stackFramesMinDuration == 0, we never set stacktrace.
+	default:
+		if !s.dropped() && len(s.stacktrace) == 0 &&
+			s.Duration >= s.stackFramesMinDuration {
+			s.setStacktrace(1)
+		}
 	}
 	// If this span has a parent span, lock it before proceeding to
 	// prevent deadlocking when concurrently ending parent and child.
--- a/span_test.go
+++ b/span_test.go
@@ -196,6 +196,38 @@ func TestStartExitSpan(t *testing.T) {
 	assert.True(t, span.IsExitSpan())
 }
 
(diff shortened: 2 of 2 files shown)


span.go
```diff
@@ -356,9 +356,17 @@ func (s *Span) End() {
...
+	case -1:
+		// Always set stacktrace
 		s.setStacktrace(1)
+	case 0:
+		// If s.stackFramesMinDuration == 0, we never set stacktrace.
+	default:
+		if !s.dropped() && len(s.stacktrace) == 0 &&
+			s.Duration >= s.stackFramesMinDuration {
+			s.setStacktrace(1)
+		}
```
stuartnelson3, line 369: This could have been squished into a single conditional, and I'm happy to do so. This just seemed to be more easy to read.

axw: Looks good to me. Maybe we can revisit when we get to https://github.com/elastic/apm-agent-go/issues/1142

span_test.go
```diff
@@ -196,6 +196,39 @@ func TestStartExitSpan(t *testing.T) {
...
+	tx = tracer.StartTransaction("name", "type")
+	span = tx.StartSpan("span2", "span2", nil)
+	span.End()
+	tx.End()
+
+	tracer.Flush(nil)
+
+	spans = tracer.Payloads().Spans
+	require.Len(t, spans, 1)
+	assert.Len(t, spans[0].Stacktrace, 4)
```
axw, line 229: Maybe just check it's non-empty? The Go runtime or testing package might add more frames in the future.

axw (approved):

axw:
Verified using `go.elastic.co/apm/v2 v2.0.0-20220209134054-640a916049c9` (current main):

```go
package main

import (
        "time"

        "go.elastic.co/apm/v2"
)

func main() {
        tracer := apm.DefaultTracer()
        defer tracer.Flush(nil)

        tx := tracer.StartTransaction("name", "type")
        defer tx.End()

        span := tx.StartSpan("name", "type", nil)
        time.Sleep(5 * time.Millisecond)
        span.End()
}
```

- By default, the span has a stack trace
- When setting ELASTIC_APM_SPAN_FRAMES_MIN_DURATION=-1ms, the span still has a stack trace
- When setting ELASTIC_APM_SPAN_FRAMES_MIN_DURATION=0ms, the span does not have a stack trace

PR #1189, opened 2022-01-11
Replace DefaultTracer with initialization func

### LINKED ISSUE #775: RFC: make DefaultTracer a function, defer init to first call
This would be an alternative to https://github.com/elastic/apm-agent-go/issues/773. Rather than initialising the default tracer at package init time, make DefaultTracer a function and initialise the tracer on first call. We would probably also need to supply a SetDefaultTracer companion function.

stuartnelson3:
Additionally, add `SetDefaultTracer` so users can set the tracer returned by `DefaultTracer()`.

closes #775

### DIFF
--- a/tracer.go
+++ b/tracer.go
@@ -48,20 +48,33 @@ const (
 )
 
 var (
-	// DefaultTracer is the default global Tracer, set at package
-	// initialization time, configured via environment variables.
-	//
-	// This will always be initialized to a non-nil value. If any
-	// of the environment variables are invalid, the corresponding
-	// errors will be logged to stderr and the default values will
-	// be used instead.
-	DefaultTracer *Tracer
+	defaultTracer *Tracer
 )
 
-func init() {
-	var opts TracerOptions
-	opts.initDefaults(true)
-	DefaultTracer = newTracer(opts)
+// DefaultTracer returns the default global Tracer, set the first time the
+// function is called. It is configured via environment variables.
+//
+// This will always be initialized to a non-nil value. If any of the
+// environment variables are invalid, the corresponding errors will be logged
+// to stderr and the default values will be used instead.
+func DefaultTracer() *Tracer {
+	if defaultTracer == nil {
+		var opts TracerOptions
+		opts.initDefaults(true)
+		defaultTracer = newTracer(opts)
+	}
+	return defaultTracer
+}
+
+// SetDefaultTracer sets the tracer returned by DefaultTracer(). If another
+// tracer has already been initialized, any events are flushed and the tracer
+// closed.
+func SetDefaultTracer(t *Tracer) {
+	if defaultTracer != nil {
+		defaultTracer.Flush(nil)
+		defaultTracer.Close()
+	}
+	defaultTracer = t
 }
 
 // TracerOptions holds initial tracer options, for passing to NewTracerOptions.
@@ -378,12 +391,6 @@ type Tracer struct {
 // This is equivalent to calling NewTracerOptions with a
 // TracerOptions having ServiceName and ServiceVersion set to
 // the provided arguments.
-//
-// NOTE when this package is imported, DefaultTracer is initialised
-// using environment variables for configuration. When creating a
-// tracer with NewTracer or NewTracerOptions, you should close
-// apm.DefaultTracer if it is not needed, e.g. by calling
-// apm.DefaultTracer.Close() in an init function.
 func NewTracer(serviceName, serviceVersion string) (*Tracer, error) {
 	return NewTracerOptions(TracerOptions{
 		ServiceName:    serviceName,
@@ -394,12 +401,6 @@ func NewTracer(serviceName, serviceVersion string) (*Tracer, error) {
 // NewTracerOptions returns a new Tracer using the provided options.
 // See TracerOptions for details on the options, and their default
 // values.
-//
-// NOTE when this package is imported, DefaultTracer is initialised
-// using environment variables for configuration. When creating a
-// tracer with NewTracer or NewTracerOptions, you should close
-// apm.DefaultTracer if it is not needed, e.g. by calling
-// apm.DefaultTracer.Close() in an init function.
 func NewTracerOptions(opts TracerOptions) (*Tracer, error) {
 	if err := opts.initDefaults(false); err != nil {
 		return nil, err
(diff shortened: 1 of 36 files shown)


marclop (commented):
Looks good except for a potential data race. Maybe I'm being overly zealous, but should we add a package `tracerMu sync.Mutex` to avoid potential races between `apm.DefaultTracer()` and `apm.SetDefaultTracer()`?

I was able to simulate a data race easily with:

```go
func TestConcurrentDefaultTracer(t *testing.T) {
	go func() { apm.DefaultTracer() }()
	go func() { apm.SetDefaultTracer(nil) }()
}
```

```console
$ go test -count=1 -race -run TestConcurrentDefaultTracer .
PASS
==================
WARNING: DATA RACE
Write at 0x000103942dd8 by goroutine 19:
  go.elastic.co/apm.SetDefaultTracer()
      /Users/marclop/repos/elastic/apm-agent-go/tracer.go:77 +0x98
  go.elastic.co/apm_test.TestConcurrentDefaultTracer.func2()
      /Users/marclop/repos/elastic/apm-agent-go/tracer_test.go:511 +0x2c

Previous read at 0x000103942dd8 by goroutine 18:
  go.elastic.co/apm.DefaultTracer()
      /Users/marclop/repos/elastic/apm-agent-go/tracer.go:61 +0x3c
  go.elastic.co/apm_test.TestConcurrentDefaultTracer.func1()
      /Users/marclop/repos/elastic/apm-agent-go/tracer_test.go:510 +0x28

Goroutine 19 (running) created at:
  go.elastic.co/apm_test.TestConcurrentDefaultTracer()
      /Users/marclop/repos/elastic/apm-agent-go/tracer_test.go:511 +0x4c
  testing.tRunner()
      /Users/marclop/.go/src/testing/testing.go:1259 +0x198

Goroutine 18 (finished) created at:
  go.elastic.co/apm_test.TestConcurrentDefaultTracer()
      /Users/marclop/repos/elastic/apm-agent-go/tracer_test.go:510 +0x38
  testing.tRunner()
      /Users/marclop/.go/src/testing/testing.go:1259 +0x198
==================
Found 1 data race(s)
FAIL	go.elastic.co/apm	0.706s
FAIL
```

When would this happen? Perhaps on initialization if the tracer is set by users on a goroutine and imported modules or other parts of the code call `apm.DefaultTracer()`. I think is a fairly common case.

axw (commented):
Looks great! I also think we should have a mutex for the DefaultTracer/SetDefaultTracer functions.

Let's make it a RWMutex, as I would expect many more calls to DefaultTracer than to SetDefaultTracer.

stuartnelson3:
Cool, added the RWMutex. I tried to enable `-race` for `make test`, but it found some other races. I've opened an issue: https://github.com/elastic/apm-agent-go/issues/1191

tracer.go
```diff
@@ -45,20 +45,41 @@ const (
...
+	tracerMu.RLock()
+	dt := defaultTracer
+	tracerMu.RUnlock()
+
+	if dt == nil {
+		var opts TracerOptions
+		opts.initDefaults(true)
+		SetDefaultTracer(newTracer(opts))
+	}
+	return defaultTracer
```
axw, line 68: this is accessing `defaultTracer` without a lock, which is racy

I think it should be:

```go
	tracerMu.RLock()
	tracer := defaultTracer
	tracerMu.RUnlock()
	if tracer != nil {
		return tracer
	}

	var opts TracerOptions
	opts.initDefaults(true)
	tracer = newTracer(opts)
	SetDefaultTracer(tracer)
	return tracer
```

tracer.go
```diff
@@ -45,20 +45,41 @@ const (
...
+
+// SetDefaultTracer sets the tracer returned by DefaultTracer(). If another
+// tracer has already been initialized, any events are flushed and the tracer
+// closed.
+func SetDefaultTracer(t *Tracer) {
+	tracerMu.Lock()
+	defer tracerMu.Unlock()
+
+	if defaultTracer != nil {
+		defaultTracer.Flush(nil)
```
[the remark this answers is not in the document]
stuartnelson3: That works for me. I wasn't sure if we should flush for them or not. Thanks for the clarification!

tracer.go
```diff
@@ -45,20 +45,42 @@ const (
...
 	var opts TracerOptions
 	opts.initDefaults(true)
-	DefaultTracer = newTracer(opts)
+	tracer = newTracer(opts)
+	SetDefaultTracer(tracer)
+	return tracer
+}
+
+// SetDefaultTracer sets the tracer returned by DefaultTracer(). If another
+// tracer has already been initialized, any events are flushed and the tracer
```
axw, line 74: Comment needs updating now that we're not flushing. Probably best to explicitly state that queued events won't be flushed, and users need to call DefaultTracer().Flush() if they need to.

axw (approved):
LGTM with one small docs update. Thanks!

axw:
Verified using go.elastic.co/apm/v2 v2.0.0-20220209134054-640a916049c9 (current main), along with apmhttp/v2. To get apmhttp/v2 to work, I had to add a replace stanza:

`replace go.elastic.co/apm/v2 v2.0.0 => go.elastic.co/apm/v2 v2.0.0-20220209134054-640a916049c9`

```go
package main

import (
        "io"
        "io/ioutil"
        "net/http"
        "net/http/httptest"

        "go.elastic.co/apm/module/apmhttp/v2"
        "go.elastic.co/apm/v2"
)

func main() {
        tracer1, _ := apm.NewTracer("svc1", "")
        defer tracer1.Flush(nil)
        apm.SetDefaultTracer(tracer1)

        tracer2, _ := apm.NewTracer("svc2", "")
        defer tracer2.Flush(nil)

        srv := httptest.NewServer(apmhttp.Wrap(http.NewServeMux()))
        resp, _ := http.Get(srv.URL)
        io.Copy(ioutil.Discard, resp.Body)
        resp.Body.Close()

        tracer2.StartTransaction("svc2_name", "svc2_type").End()
        time.Sleep(10 * time.Second)
}
```

- The apmhttp-instrumented HTTP server uses the default tracer, confirmed by checking the indexed transaction has `service.name: svc1`.
- Confirmed that the non-default tracer sends a transaction with `service.name: svc2`
- Confirmed there were two `GET /config/v1/agents` requests to the server, with `url.query: service.name=svc1` and `url.query: service.name=svc2`.
- Direct, implicit configuration of DefaultTracer() was tested indirectly in https://github.com/elastic/apm-agent-go/pull/1188#issuecomment-1034419654

Issue #655, opened 2019-10-16
Drop support for older versions of Go


axw:
We should drop older versions of Go - at least 1.8.x - when we bump major version.

PR #1190, opened 2022-01-12
Drop support for versions of Go prior to 1.15.0

axw:
Closes https://github.com/elastic/apm-agent-go/issues/655


marclop (approved):
Thanks!

axw:
It's a little up in the air as to what the minimum supported version for v2 will be. I'm going to merge this as it is, and we'll nail that down before releasing v2.0.0.

PR #1193, opened 2022-01-13
[apmfasthttp] fix: set response context with closer close

savsgio:
Fix and set correctly the response context data to the apm transacction after the request finish (included Hijack and Stream connections). 

Relationed issue: https://github.com/elastic/apm-agent-go/pull/1104
Fasthttp issues: https://github.com/valyala/fasthttp/pull/1176,  https://github.com/valyala/fasthttp/pull/1199


stuartnelson3:
Can you add a test to verify the fixed behavior?

savsgio:
> Can you add a test to verify the fixed behavior?

The test added in https://github.com/elastic/apm-agent-go/pull/1104 tests the same case!

axw:
@savsgio the test you reference was succeeding without your change. Please add a test which would fail without your change, and succeeds with your change.

savsgio:
@axw could you review it again?, please!

axw:
Thanks @savsgio, I will take a look soon. We have a few other things going on too, so it may take a little while.

module/apmfasthttp/context.go
```diff
@@ -111,7 +109,7 @@ func StartTransactionWithBody(
 		return nil, nil, err
 	}
 
-	ctx.SetUserValue(txKey, newTxCloser(tx, bc))
+	ctx.SetUserValue(txKey, newTxCloser(ctx, tx, bc))
```
axw, line 112: The fasthttp docs state:

> VERY IMPORTANT! Fasthttp disallows holding references to RequestCtx or to its' members after returning from RequestHandler. Otherwise data races are inevitable. Carefully inspect all the net/http request handlers converted to fasthttp whether they retain references to RequestCtx or to its' members after returning.

With your change, isn't StartTransactionWithBody now causing a reference to RequestCtx to be held (in ctx itself) after the handler returns?

savsgio: The docs reference to use a RequestCtx after the RequestHandler finish (parallel gorutines that use the RequestCtx), but the context user values are allowed, because it's reset before being put back into the pool. So it's safe!

savsgio: Sure! 😉

module/apmfasthttp/context.go
```diff
@@ -78,10 +77,9 @@ func setResponseContext(ctx *fasthttp.RequestCtx, tx *apm.Transaction, bc *apm.B
 		headers.Set(sk, sv)
 	})
 
+	tx.Result = apmhttp.StatusCodeResult(statusCode)
```
axw, line 80: Result should still be set for non-sampled transactions. Any particular reason why you changed this method?

module/apmfasthttp/server_test.go
```diff
@@ -30,67 +32,159 @@ import (
...
-			break
-		}
-		time.Sleep(10 * time.Millisecond)
+
+		count++
+
+		time.Sleep(100 * time.Millisecond)
 		tracer.Flush(nil)
 		payloads = transport.Payloads()
 	}
```
axw, line 90: Why has this changed? Doesn't seem related to the PR. I'm concerned that your changes may lead to failures on slow CI machines, so please revert this unless it's needed.

axw (changes requested):
Thanks for the changes! Looks good apart from the couple of comments I've left.

module/apmfasthttp/context.go
savsgio: Done!

module/apmfasthttp/server_test.go
savsgio: Done!

axw (approved):
Sorry for the delay, LGTM. Thanks!

axw:
@savsgio CI is unhappy:

```
17:47:36  + make install precheck check-modules
17:47:36  go get -v -t ./...
17:47:36  goimports differs:
17:47:36   - module/apmfasthttp/closer.go
17:47:36  Makefile:12: recipe for target 'check-goimports' failed
```

Please run `make fmt` to fix this issue, and `make check` to ensure there are no others.

axw:
@savsgio the change has broken the apmfiber tests. Can you please take a look at those?

axw:
Thanks @savsgio!

savsgio:
Thank you so much @axw!

Issue #1136, opened 2021-10-12
[META 527] User-agent header for transport


SylvainJuge:
Implement [514](https://github.com/elastic/apm/pull/514) for all APM agents.

axw:
I think this is going to require breaking changes to our API. Service name and version are properties of the tracer. A transport (which defines the User-Agent) can be shared by multiple tracers. We use `transport.Default` if one is not specified when constructing a tracer, so we can't reasonably update its User-Agent.

I think we'll need to get rid of the `transport.Default` singleton, and always construct one when creating a tracer.

PR #1196, opened 2022-01-18
Include service name and version in User-Agent

axw:
Update User-Agent to match the [agent spec](https://github.com/elastic/apm/blob/main/specs/agents/transport.md#user-agent), including service name and service version if it is specified. User-Agent no longer includes the Go runtime version, but this is still available in `service.runtime.version`.

The `transport.NewHTTPTransportOptions` function is removed, and `transport.NewHTTPTransport` is updated to take HTTPTransportOptions instead now that we can break the API. Use env vars as defaults only if values are not specified in the options struct, aligning with behaviour of `apm.NewTracerOptions`.

Update to the latest shared agents Gherkin spec.

Closes https://github.com/elastic/apm-agent-go/issues/1136

### DIFF
--- a/config.go
+++ b/config.go
@@ -18,6 +18,7 @@
 package apm // import "go.elastic.co/apm"
 
 import (
+	"fmt"
 	"net/url"
 	"os"
 	"path/filepath"
@@ -140,9 +141,14 @@ var (
 )
 
 func initialTransport(serviceName, serviceVersion string) (transport.Transport, error) {
-	// TODO(axw) pass service name and version, to append to the User-Agent header.
+	// User-Agent should be "apm-agent-go/<agent-version> (service-name service-version)".
+	service := serviceName
+	if serviceVersion != "" {
+		service += " " + serviceVersion
+	}
+	userAgent := fmt.Sprintf("%s (%s)", transport.DefaultUserAgent(), service)
 	httpTransport, err := transport.NewHTTPTransport(transport.HTTPTransportOptions{
-		UserAgent: "user-agent",
+		UserAgent: userAgent,
 	})
 	if err != nil {
 		return nil, err
--- a/config.go
+++ b/config.go
@@ -141,7 +141,9 @@ var (
 
 func initialTransport(serviceName, serviceVersion string) (transport.Transport, error) {
 	// TODO(axw) pass service name and version, to append to the User-Agent header.
-	httpTransport, err := transport.NewHTTPTransport()
+	httpTransport, err := transport.NewHTTPTransport(transport.HTTPTransportOptions{
+		UserAgent: "user-agent",
+	})
 	if err != nil {
 		return nil, err
 	}
--- a/config.go
+++ b/config.go
@@ -22,6 +22,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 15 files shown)


config.go
```diff
@@ -139,9 +141,21 @@ var (
 	}()
 )
 
+// Regular expression matching comment characters to escape in the User-Agent header value.
+//
+// See https://httpwg.org/specs/rfc7230.html#field.components
+var httpComment, _ = regexp.Compile("[^\\t \\x21-\\x27\\x2a-\\x5b\\x5d-\\x7e\\x80-\\xff]")
```
axw, line 147: cargo culted from https://github.com/elastic/apm-agent-java/pull/2262/commits/817feefbbc14837a56521ec9f18367905655a8f9

transport/http.go
```diff
@@ -73,37 +72,71 @@ var (
...
+	// 30 seconds if the environment variable is not set. Negative values are
+	// not allowed, and will cause NewHTTPTransport to return an error.
 	ServerTimeout time.Duration
 
-	// SkipServerVerify skips TLS certificate validation of the APM Server.
+	// SkipServerVerify skips TLS certificate verification of the APM Server.
+	// By default, the server's TLS certificate will be verified.
+	//
+	// If false, SkipServerVerify will be initialized using the
+	// ELASTIC_APM_VERIFY_SERVER_CERT environment variable.
```
stuartnelson3, line 132: The behavior I'm used to (and I think what we generally do in apm-server / apm-agent-go) is that an explicitly configured value in the constructor takes precedence over an environment variable. I would expect that setting `false` explicitly for `SkipServerVerify` would ensure the value is false, which isn't the case (the environment variable can overrule an explicitly set `false`). That behavior is clearly described in the comment, so that's fine, but I'm wondering if this is something we want to consider.

stuartnelson3 (approved):
looks good! small question, but that's all.

transport/http.go
axw: @marclop also raised this in standup yesterday.

What do you both think about removing `ServerCert`, `ServerCACert`, and `SkipVerifyCert` fields, and instead introducing a `TLS *tls.Config` field? Only if `TLS` field is nil would `ELASTIC_APM_SERVER_CERT`, `ELASTIC_APM_SERVER_CA_CERT_FILE`, and `ELASTIC_APM_VERIFY_SERVER_CERT` be used.

That would make the very simplest things a little bit more verbose, but provides more flexibility to users for more complicated use cases.

marclop: sounds goo to me!

marclop (approved):

config.go
axw: most definitely, I just had a brain fade

transport/http.go
stuartnelson3: :+1: that sounds like a good solution to me

stuartnelson3 (approved):

stuartnelson3 (changes requested):
taking that back :sweat_smile: this is good to go once the tls config options get rolled into a `*tls.Config`

axw:
@stuartnelson3 merging as I think you concluded that all necessary changes were already made. Let me know if I misinterpreted.

marclop:
ah I didn't read the UserAgent comment, it was unexpected to me since thought it should've also extended the UserAgent when not explicitly set and the `ServiceName` and `ServiceVersion` was specified in `HTTPTransportOptions`.

```go
tracer, err := apm.NewTracerOptions(apm.TracerOptions{
	Transport: t,
}) // Yields UserAgent = `apm-agent-go/2.0.0`

tracer, err := apm.NewTracerOptions(apm.TracerOptions{
	ServiceName:    "marc-test",
	ServiceVersion: "5.1.3",
	Transport:      t,
}) // Yields UserAgent = `apm-agent-go/2.0.0 (marc-test 5.3.1)`

tracer, err := apm.NewTracerOptions(apm.TracerOptions{
	ServiceName:    "marc-test",
	ServiceVersion: "5.1.3",
	Transport:      t,
	UserAgent:      "custom-user-agent"
}) // Yields UserAgent = `custom-user-agent`
```

It isn't a regression then since the mechanics are clearly defined in the godocs.

axw:
Maybe we can improve this later, e.g. by adding a `SetUserAgent` method to the Transport interface. Then NewTracerOptions could update it. I prefer it being explicitly specified, but I can see how it's a bit surprising.

PR #1209, opened 2022-02-07
add ParseTraceState() fn

stuartnelson3:
for setting TraceState when bridging another
tracing library implementation, for example
OpenTelemetry.

cf. https://github.com/elastic/apm-agent-go/pull/1203/commits/e27194e12120c865addece26b85fd7188fe5901d

### DIFF
--- a/tracecontext.go
+++ b/tracecontext.go
@@ -34,8 +34,9 @@ const (
 )
 
 var (
-	errZeroTraceID = errors.New("zero trace-id is invalid")
-	errZeroSpanID  = errors.New("zero span-id is invalid")
+	errZeroTraceID    = errors.New("zero trace-id is invalid")
+	errZeroSpanID     = errors.New("zero span-id is invalid")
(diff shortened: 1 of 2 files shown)


tracecontext.go
```diff
@@ -204,6 +205,58 @@ func NewTraceState(entries ...TraceStateEntry) TraceState {
 	return out
 }
 
+// ParseTraceState attempts to decode a TraceState from the passed
+// string. It returns an error if the input is invalid according to the W3C
+// Trace Context specification.
+// Adapted from https://github.com/open-telemetry/opentelemetry-go/blob/trace/v1.3.0/trace/tracestate.go.
```
axw, line 211: We have our own parse function already: `apmhttp.ParseTracestateHeader`. Does this do something we don't already do? If we need it elsewhere, should we just move it to the top-level package?

stuartnelson3: ahh, I only looked for this in the top-level package. nope, looks like it does the thing!

PR #1210, opened 2022-02-07
Add otel type

stuartnelson3:
otel type will be populated with data when using
`apmotel` module

cf. https://github.com/elastic/apm-agent-go/commit/cb65b4002280b587b7b119fc3506da332f0b5684


context.go
```diff
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"net/http"
 
+	"go.opentelemetry.io/otel/attribute"
```
axw, line 24: I'm not too keen on adding OTel as a dependency for every user. Would it be reasonable to use `map[string]interface{}` in this code, and convert from `go.opentelemetry.io/otel/attribute` in apmotel?

context.go
```diff
@@ -83,6 +86,16 @@ func (c *Context) reset() {
 	}
 }
 
+// SetOtelAttributes sets the provided OpenTelemetry attributes.
```
axw, line 89:
```suggestion
// SetOTelAttributes sets the provided OpenTelemetry attributes.
```

Nit, OpenTelemetry is commonly abbreviated OTel with a capital T.

stuartnelson3: :+1:

modelwriter.go
```diff
@@ -150,6 +151,7 @@ func (w *modelWriter) buildModelSpan(out *model.Span, span *Span, sd *SpanData)
 	out.Duration = sd.Duration.Seconds() * 1000
 	out.Outcome = normalizeOutcome(sd.Outcome)
 	out.Context = sd.Context.build()
+	out.OTel = sd.Context.otel
```
axw, line 154: I suppose it's too late now, but we probably should have nested `otel.attributes` underneath `context` on the wire.

axw (approved):
LGTM, but I think SetSpanKind should probably have OTel in the name.

modelwriter.go
stuartnelson3: Would the apm-server then move `otel` up a level when mapping, to be right under `transaction`/`span`? From the spec:

>OpenTelemetry attributes should be stored in `otel.attributes` as a flat key-value pair mapping added to `span` and `transaction` objects

Back when this was implemented in the apm-server, I took this to mean that `otel.attributes` would already be on the incoming wire format for the event. It could still be changed, but would require coordination with any agents that may have already implemented this.

For my edification, why should this be underneath `context` on the wire?

axw: > Would the apm-server then move otel up a level when mapping, to be right under transaction/span? From the spec:

This isn't about the Elasticsearch mapping right? Just the intake protocol? What I'm saying is instead of

```json
{"transaction":{
  // [...] other span/transaction attributes
  "otel": {
    "span_kind": "CLIENT",
    "attributes": {
      "db.system": "mysql",
      "db.statement": "SELECT * from table_1"
    }
  }
}}
```

we should have had

```json
{"transaction":{
  "context": {
    // [...] other span/transaction attributes
    "otel": {
      "span_kind": "CLIENT",
      "attributes": {
        "db.system": "mysql",
        "db.statement": "SELECT * from table_1"
      }
    }
  }
}}
```

> Back when this was implemented in the apm-server, I took this to mean that otel.attributes would already be on the incoming wire format for the event. It could still be changed, but would require coordination with any agents that may have already implemented this.
>
>
> For my edification, why should this be underneath context on the wire?

So I was thinking that we should have it under `context` to align with labels, HTTP request/response context, etc. All these things that describe details of a transaction or span, but which aren't required attributes and won't break anything if we drop them.

I think maybe I was wrong though: if we were to drop context, and `otel.attributes` with it, then we wouldn't be able to infer the transaction type. @felixbarny @SylvainJuge do you know if this was the intention?

SylvainJuge: I don't think we had in mind to make `otel` attribute mandatory by making it a top-level element of the `transaction` / `span` objects.

From what I remember it seemed just relevant to put it here as a way to mean "this object is built from OTel data", in hindsight we could definitely have moved that into the `context` field as most of the otel fields and attributes values represent things that we usually store in specialized fields in `context`.

On the other end, for `{span,transaction}.type` or `span.subtype` which are inferred from values in `{span,transaction}.otel`, it might be slightly easier to avoid having "top-level attributes" to have their values inferred from closer sibling/cousin fields rather than small-small nephews deeper in the family tree (sorry, I'm out of good metaphors here).

I think that it would be definitely worth revisiting this if and when we want to merge the concepts of transaction & spans.

PR #1215, opened 2022-02-22
Update go module dependencies

axw:
- Update `golang.org/x/tools` to the latest (v0.1.9), so it works with Go 1.18
- Update `go.mod` to specify `go 1.15`, aligning with our removal of support for Go <1.15
- Drop use of `golint`, which panics with Go 1.18, and is deprecated and frozen


marclop (approved):
Looks good, but I've noticed that a lot of the `module/.../go.mod` files specify their own `go` version requirement, should we update them all to `1.15`? They all rely on `go.elastic.co/apm/v2` which already requires 1.15.

axw:
Good call, I'll update them all to `go 1.15` before merging. I may also update genmod to check this.

PR #1231, opened 2022-03-23
use updated default service name

### LINKED ISSUE #1162: [META 571] Align on default service name
[issue details](https://github.com/elastic/apm/issues/571)

[issue details](https://github.com/elastic/apm/issues/572)

As defined in #499, agents don't require users to configure the `service_name`. Ideally, the a descriptive value can be inferred form the application. But in some cases that's not possible. In these cases, agents must now align on the naming scheme for the default service name. This enables the UI to show inline help for services that it identifies as unnamed services based on the naming scheme.

stuartnelson3:
default service name is
`unknown-${service.agent.name}-service`, as
described in https://github.com/elastic/apm/blob/34f54c7b177ada87162f92595317edae5216311e/specs/agents/configuration.md#zero-configuration-support

closes #1162

### DIFF
--- a/config.go
+++ b/config.go
@@ -21,9 +21,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
-	"path/filepath"
 	"regexp"
-	"runtime"
 	"strconv"
 	"strings"
 	"sync/atomic"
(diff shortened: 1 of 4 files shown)


marclop (approved):

axw:
Can we hold off on merging please? This doesn't seem like an improvement for Go, to me. I'd like to have a look at the spec and maybe discuss with other agent devs.

axw:
From the spec:

> If possible, agents SHOULD detect sensible defaults for service.name and service.version. In any case agents MUST include service.name - if discovering it is not possible then the default value: unknown-${service.agent.name}-service MUST be used.

We were already doing the first bit ("detect sensible defaults") for service.name, so the fallback is not necessary.

PR #1232, opened 2022-03-23
Revert "use updated default service name"

axw:
Reverts elastic/apm-agent-go#1231

From the spec:

> If possible, agents SHOULD detect sensible defaults for service.name and service.version. In any case agents MUST include service.name - if discovering it is not possible then the default value: unknown-${service.agent.name}-service MUST be used.

We were already doing the first bit ("detect sensible defaults") for service.name, so the fallback is not necessary. It's only necessary to use `unknown-${service.agent.name}-service` if a "sensible default" cannot be detected.


marclop (approved):
I had misread the shoulds / musts in the spec language, thank you for reverting.

PR #1239, opened 2022-04-07
Fetch initial server version async

### LINKED ISSUE #1238: `NewTracer(Options)` blocks up to 10 seconds when server is unresponsive
When `NewTracer` is called, it starts a background loop and sends the initial config over a channel. We may block the loop startup with a server version query: https://github.com/elastic/apm-agent-go/blob/3d60cbda513315c34ce80b35a192a483c875689f/tracer.go#L894

e.g.

```go
package main                                                                                        
                                                                                                    
import (                                                                                            
        "fmt"                                                                                       
        "net/http"                                                                                  
        "net/http/httptest"                                                                         
        "os"                                                                                        
        "time"                                                                                      
                                                                                                    
        "go.elastic.co/apm/v2"                                                                      
)                                                                                                   
                                                                                                    
func main() {                                                                                       
        srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {   
                <-r.Context().Done()                                                                
        }))                                                                                         
        os.Setenv("ELASTIC_APM_SERVER_URL", srv.URL)                                                
                                                                                                    
        before := time.Now()                                                                        
        apm.DefaultTracer()                                                                         
        fmt.Println(time.Since(before))                                                             
}                                                                                                   
```

This should exit immediately, but it blocks for 10 seconds.

axw:
Fixes #1238

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -23,6 +23,8 @@ endif::[]
 
 https://github.com/elastic/apm-agent-go/compare/v2.0.0...main[View commits]
 
+- Fetch initial server version async to prevent blocking NewTracer for 10 seconds {pull}1239[#(1239)]
+
 [[release-notes-2.x]]
 === Go Agent version 2.x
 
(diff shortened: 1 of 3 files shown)


marclop (approved):
Thanks for the fix!

PR #1248, opened 2022-05-20
apm: fix race in DefaultTracer

### LINKED ISSUE #1247: Concurrent calls to DefaultTracer results in 2 tracers existing with one being inactive
**Describe the bug**
When 2 goroutines run at the same time at start-up and both call the apm.DefaultTracer call, then a race-condition occurs where both calls attempt to create a new Tracer. In this scenario, one tracer loses out and disables itself while the other tracer continues to work.

I encountered this during a start-up scheduled task and creation of a gin handler. The gin tracer was being marked as inactive while the scheduled process continues.

There is a read-lock on the tracer. However, this is only when it is read, and is a very short lock. When the first caller runs, it gets null and releases the lock. While the first caller is creating the new tracer, the second caller runs this method, and passes through the lock, also seeing null. This caller then creates a new tracer. When SetDefaultTracer is called after it creates it, it will terminate the tracer that the first caller generated and made default.

**To Reproduce**
Following code:
```golang
go func() {
  web := gin.New()
  web.Use(apmgin.Middleware(web))
}

go func() {
  trans := apm.DefaultTracer().StartTransaction("gameCacheSync", "schedule")
  defer trans.End()
}
```

**Expected behavior**
Single tracker exists, which sends transactions from both Gin and scheduled process.

axw:
Fix a race in DefaultTracer which could lead to two tracers being created, one of which would be closed.
We must hold the write lock while creating the default tracer, checking again that it was not set after releasing the read lock.

Fixes #1247

### DIFF
--- a/tracer.go
+++ b/tracer.go
@@ -50,30 +50,41 @@ var (
 )
 
 // DefaultTracer returns the default global Tracer, set the first time the
-// function is called. It is configured via environment variables.
+// function is called, or after calling SetDefaultTracer(nil).
 //
-// This will always be initialized to a non-nil value. If any of the
-// environment variables are invalid, the corresponding errors will be logged
-// to stderr and the default values will be used instead.
+// The default tracer is configured via environment variables, and will always
+// be non-nil. If any of the environment variables are invalid, the
+// corresponding errors will be logged to stderr and the default values will be
+// used instead.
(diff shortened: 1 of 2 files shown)


marclop (approved):
Thanks for the quick fix!

PR #1255, opened 2022-05-30
scripts: add workflow to add to APM github project

simitt:
Add all newly created issues to the APM project board, labeled with team `server` to ensure we keep track of issues in our planned work. 

The script is copied from the [APM Server workflow](https://github.com/elastic/apm-server/blob/main/.github/workflows/add-to-apm-project.yml).

### DIFF
--- a/.github/workflows/addToAPMProject.yml
+++ b/.github/workflows/addToAPMProject.yml
@@ -1,0 +1,46 @@
+name: Add to APM Project
+on:
+  issues:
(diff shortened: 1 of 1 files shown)


axw (approved):
LGTM. I wonder if we should be using elastic/assign-one-project-github-action like in https://github.com/elastic/apm-agent-go/blob/main/.github/workflows/addToProject.yml?

PR #1261, opened 2022-06-29
Expose method to get metadata collected by agent

lahsivjar:
## Motivation

Agent collects rich metadata about service, process and system. The metadata can be used to enrich information collected by other libraries. One possible use case can be to enrich logs shipped by custom log appenders.

### DIFF
--- a/tracer.go
+++ b/tracer.go
@@ -849,6 +849,85 @@ func (t *Tracer) Stats() TracerStats {
 	return t.stats.copy()
 }
 
+// GetECSMetadata returns ECS compliant metadata collected by the agent
+func (t *Tracer) GetECSMetadata() map[string]interface{} {
+	m := make(map[string]interface{})
+
+	service := make(map[string]interface{})
+	service["name"] = t.service.Name
+	service["version"] = t.service.Version
+	service["environment"] = t.service.Environment
+	m["service"] = service
+
+	if t.process != nil {
+		process := make(map[string]interface{})
+		process["title"] = t.process.Title
+		process["parent"] = map[string]interface{}{
+			"pid": t.process.Ppid,
+		}
+		m["process"] = process
+	}
+
+	if mdCloud := getCloudMetadata(); mdCloud != nil {
+		cloud := make(map[string]interface{})
+		cloud["provider"] = mdCloud.Provider
+		cloud["availability_zone"] = mdCloud.AvailabilityZone
+		cloud["region"] = mdCloud.Region
+		if mdCloud.Project != nil {
+			project := make(map[string]interface{})
+			project["id"] = mdCloud.Project.ID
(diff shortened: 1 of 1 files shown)


tracer.go
```diff
@@ -849,6 +849,85 @@ func (t *Tracer) Stats() TracerStats {
...
+
+	service := make(map[string]interface{})
+	service["name"] = t.service.Name
+	service["version"] = t.service.Version
+	service["environment"] = t.service.Environment
+	m["service"] = service
+
+	if t.process != nil {
+		process := make(map[string]interface{})
+		process["title"] = t.process.Title
```
felixbarny, line 864:
```suggestion
		process["pid"] = t.process.Pid
```

tracer.go
```diff
@@ -849,6 +849,85 @@ func (t *Tracer) Stats() TracerStats {
...
+			orchestrator["type"] = "kubernetes"
+			orchestrator["namespace"] = mdSystem.Kubernetes.Namespace
+			if pod := mdSystem.Kubernetes.Pod; pod != nil {
+				resource := make(map[string]interface{})
+				resource["id"] = pod.UID
+				resource["name"] = pod.Name
+				resource["type"] = "pod"
+				orchestrator["resource"] = resource
+			}
+			m["orchestrator"] = orchestrator
```
felixbarny, line 919: It seems like events from APM Server and Filebeat don't use the official orchestrator namespace but an unofficial `kubernetes.*` namespace.
Let's leave this in as-is for now but it might make correlation harder.

axw (commented):
@lahsivjar I'm not too keen on making this change without us having a plan to use it. One of the nice things about doing the transformation to Elasticsearch docs in APM Server is that very old agents continue to work without change. Beyond that, I'd be a bit concerned about spreading the logic of mapping around.

felixbarny:
@lahsivjar and I talked about this a while ago and agreed that it's better to instead extend the intake API to add support for `log` events next to `transaction`, `span`, `error` and `metricset`. This way, the APM Server can be the central place of mapping the metadata that agents send to ECS.

So in order to make the Go agent send logs to APM Server, the Go agent would expose appenders for popular log libraries that the user can configure using the ecs-logging-go-* formatters. The appenders would then just format the log events using the ecs logging formatter and add the string/byte[] log event to the APM Server reporter queue that also holds other APM events.

I've prototyped this for the Java agent: https://github.com/elastic/apm-agent-java/pull/2694
It works a bit differently as Java offers options to make it more magical using byte code instrumentation. So for the Java agent, all the user needs to do is to set the option `ship_logs=true`.

axw:
:+1: @lahsivjar should we close this then?

lahsivjar:
Thanks, @felixbarny for the details. @axw Yup, closing this.

PR #1268, opened 2022-07-19
Added span request name option to echo tracer

aneshas:
Adding a span name option to echo apm middleware such as the one apmhttp has

### DIFF
--- a/module/apmechov4/middleware.go
+++ b/module/apmechov4/middleware.go
@@ -179,7 +179,7 @@ func WithTracer(t *apm.Tracer) Option {
 }
 
 // WithRequestName returns an Option which sets r as the function
-// to use to obtain the span name for the given echo request.
+// to use to obtain the transaction name for the given echo request.
 func WithRequestName(r RequestNameFunc) Option {
 	if r == nil {
 		panic("r == nil")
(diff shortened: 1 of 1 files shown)


module/apmechov4/middleware.go
```diff
@@ -167,6 +178,18 @@ func WithTracer(t *apm.Tracer) Option {
 	}
 }
 
+// WithRequestName returns an Option which sets r as the function
+// to use to obtain the span name for the given echo request.
```
axw, line 182:
```suggestion
// to use to obtain the transaction name for the given echo request.
```

axw (approved):
Thanks @aneshas! Looks good, just one small nitpick about a doc comment.

module/apmechov4/middleware.go
aneshas: Yes makes sense

axw:
Thanks :)

aneshas:
@axw might be a stupid question, but how does this actually get released ?

axw:
@aneshas not a stupid question! We generally create new releases around the time of Elastic Stack minor releases (next one being 8.4.0). That's expected within weeks, so if you would like to update sooner, I would suggest using the unreleased commit (update modules to `@latest`) in your go.mod.

aneshas:
Ah tnx

PR #1269, opened 2022-07-19
feat: implement Span Links API

kruskall:
This is an attempt to implement the Span Links API, mostly inspired by other PRs to the other agents and the spec/docs for OT.

Closes https://github.com/elastic/apm-agent-go/issues/1243

### DIFF
--- a/span.go
+++ b/span.go
@@ -236,7 +236,7 @@ func (t *Tracer) startSpan(name, spanType string, transactionID SpanID, opts Spa
 	span.transactionID = transactionID
 	span.timestamp = opts.Start
 	span.Type = spanType
-	span.Links = opts.Links
+	span.links = opts.Links
 	if dot := strings.IndexRune(spanType, '.'); dot != -1 {
 		span.Type = spanType[:dot]
 		span.Subtype = spanType[dot+1:]
@@ -582,7 +582,7 @@ type SpanData struct {
 	// Context describes the context in which span occurs.
 	Context SpanContext
 
-	Links []SpanLink
+	links []SpanLink
 
 	mu            sync.Mutex
 	stacktrace    []stacktrace.Frame
--- a/span.go
+++ b/span.go
@@ -219,6 +219,9 @@ type SpanOptions struct {
 	// transaction timestamp. Calculating the timstamp in this way will ensure
 	// monotonicity of events within a transaction.
 	Start time.Time
+
+	// Links, if non-nil, holds a list of spans linked to the span.
+	Links []SpanLink
 }
 
 func (t *Tracer) startSpan(name, spanType string, transactionID SpanID, opts SpanOptions) *Span {
@@ -233,6 +236,7 @@ func (t *Tracer) startSpan(name, spanType string, transactionID SpanID, opts Spa
 	span.transactionID = transactionID
 	span.timestamp = opts.Start
 	span.Type = spanType
+	span.Links = opts.Links
 	if dot := strings.IndexRune(spanType, '.'); dot != -1 {
 		span.Type = spanType[:dot]
 		span.Subtype = spanType[dot+1:]
@@ -577,6 +581,8 @@ type SpanData struct {
 
 	// Context describes the context in which span occurs.
 	Context SpanContext
+
+	Links []SpanLink
 
 	mu            sync.Mutex
 	stacktrace    []stacktrace.Frame
--- a/span_test.go
+++ b/span_test.go
@@ -122,51 +122,28 @@ func TestSpanLink(t *testing.T) {
 	tracer := apmtest.NewRecordingTracer()
 	defer tracer.Close()
 
-	runFunc := func(s apm.SpanOptions) *apm.Span {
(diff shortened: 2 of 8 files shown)


span.go
```diff
@@ -578,6 +582,8 @@ type SpanData struct {
 	// Context describes the context in which span occurs.
 	Context SpanContext
 
+	Links []SpanLink
```
axw, line 585: I think we should we unexport this. The spec says:

> The public API MUST NOT allow adding span links *after* span creation.

tracecontext.go
```diff
@@ -138,6 +138,12 @@ func (id SpanID) MarshalText() ([]byte, error) {
 	return text, nil
 }
 
+// SpanLink describes a linked span.
+type SpanLink struct {
+	TraceID TraceID
+	SpanID  SpanID
```
axw, line 144: Perhaps for consistency with the TraceContext type's field names, call these `Trace` and `Span`?

axw (changes requested):
Looking good!

We'll need to add this for transactions too.

Can you please also add tests (one for spans, one for transactions)? We should show that the span links are included in the events sent, which we can verify using the apmtest.RecordingTracer. See `span_test.go` and `transaction_test.go` for examples. Since the spec says "Links SHOULD preserve the order in which they are set", it would be good to show that in the tests too.

span.go
kruskall: Good catch! Fixed!

kruskall:
@axw Thank you for the feedback! :bow:  

I've pushed some changes and added the tests. 

> We'll need to add this for transactions too.

Thank you for pointing this out, I reread the spec and I completely missed that part :facepalm:

span_test.go
```diff
@@ -118,6 +118,60 @@ func TestSpanParentID(t *testing.T) {
 	assert.Equal(t, model.SpanID(parentID), payloads.Spans[0].ParentID)
 }
 
+func TestSpanLink(t *testing.T) {
```
axw, line 121: I think this test could be simplified, so it just asserts bits related to span links. Assertions on parent ID being non-empty and such shouldn't be necessary; that is covered by other tests.

Suggestions:
 - create a single span, like `tracer.StartSpan("name", "type", apm.SpanID{}, apm.SpanOptions{Links: ...})`
 - hand craft the span links, like `apm.SpanLink{Trace: apm.TraceID{...}, Span: apm.SpanID{...}}`

You should end up with something like:

```go
links := []apm.SpanLink{...} // mock IDs
span := tracer.StartSpan("name", "type", apm.SpanID{}, apm.SpanOptions{Links: links})
defer span.End()

tracer.Flush(nil)
payloads := tracer.Payloads()
require.Len(t, payloads.Spans, 1)

expectedLinks := []model.SpanLink{...} // also mock IDs, same values as in links
assert.Equal(t, expectedLinks, payloads.Spans[0].Links)
```

transaction_test.go
```diff
@@ -371,6 +371,54 @@ func TestTransactionSampleRateOmission(t *testing.T) {
 	}
 }
 
+func TestTransactionSpanLink(t *testing.T) {
```
axw, line 374: same as above

axw (commented):
Thanks, looks good. I'd like to see the tests simplified, but otherwise LGTM.

span_test.go
kruskall: Thanks for the tip! :bow: 

I've pushed some changes. I still had to keep the loop as `apm.SpanLink` and `model.SpanLink` are two different types so the equality check would fail.

transaction_test.go
kruskall: Thanks! Changed :+1:

kruskall:
Thanks for the feedback! I've pushed some changes :ping_pong:

span_test.go
```diff
@@ -118,6 +118,37 @@ func TestSpanParentID(t *testing.T) {
...
+	payloads := tracer.Payloads()
+	require.Len(t, payloads.Spans, 1)
+	require.Len(t, payloads.Spans[0].Links, len(links))
+
+	// Assert equality and elements order
+	for i, sl := range links {
+		l := payloads.Spans[0].Links[i]
+		assert.Equal(t, model.SpanID(sl.Span), l.SpanID)
+		assert.Equal(t, model.TraceID(sl.Trace), l.TraceID)
+	}
```
axw, line 149:
```suggestion
	// Assert span links are identical.
	expectedLinks := []model.SpanLink{
	        {TraceID: model.TraceID{1}, SpanID: model.SpanID{1}},
        	{TraceID: model.TraceID{2}, SpanID: model.SpanID{2}},
	}
	assert.Equal(t, expectedLinks, payloads.Spans[0].Links)
```

This does repeat the values, but IMO it's a bit clearer what's being asserted.

axw (approved):
LGTM! Just a suggestion about the test, I'll leave it to you to decide which way you prefer.

span_test.go
kruskall: Agree, I like that! :+1: 

Changed along with the transaction test!

PR #1270, opened 2022-07-22
feat: implement trace_continuation_strategy

kruskall:
Add ELASTIC_APM_TRACE_CONTINUATION_STRATEGY variable to change the
continuation strategy. (Default to 'continue')

Add test to verify trace ids are different and span links are
populated properly.

Closes https://github.com/elastic/apm-agent-go/issues/1221

### DIFF
--- a/config.go
+++ b/config.go
@@ -258,15 +258,15 @@ func initContinuationStrategy() (string, error) {
 	if value == "" {
 		return defaultContinuationStrategy, nil
 	}
-	return parseContinuationStrategy(value)
+	return value, validateContinuationStrategy(value)
 }
 
-func parseContinuationStrategy(value string) (string, error) {
+func validateContinuationStrategy(value string) error {
 	switch value {
 	case "continue", "restart", "restart_external":
-		return value, nil
+		return nil
 	default:
-		return "", fmt.Errorf("unknown continuation strategy: %s", value)
+		return fmt.Errorf("unknown continuation strategy: %s", value)
 	}
 }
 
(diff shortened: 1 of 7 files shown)


config.go
```diff
@@ -251,6 +253,23 @@ func initialSanitizedFieldNames() wildcard.Matchers {
...
+}
+
+func parseContinuationStrategy(value string) (string, error) {
+	switch value {
+	case "continue", "restart", "restart_external":
+		return value, nil
+	default:
+		return "", fmt.Errorf("unknown continuation strategy: %s", value)
+	}
+}
```
axw, line 271:
```suggestion
func validateContinuationStrategy(value string) error {
	switch value {
	case "continue", "restart", "restart_external":
		return nil
	default:
		return fmt.Errorf("unknown continuation strategy: %s", value)
	}
}
```

Seems a bit unnecessary to return the value that's passed in.

module/apmhttp/handler_test.go
```diff
@@ -502,6 +502,50 @@ func TestHandlerTraceparentHeader(t *testing.T) {
 	}
 }
 
+func TestHandleContinuationStrategy(t *testing.T) {
```
axw, line 505: I would prefer that we have this test in `package apm`, since that's where the logic lives. Can we test the functionality using `Tracer.StartTransactionOptions`? Maybe as a table test, with a combination of trace states with/without es, and the continuation strategy.

axw (changes requested):
Looks great, just a couple of comments on the tests.

Can you please also add a subtest to `TestTracerCentralConfigUpdate` in config_test.go?

config.go
kruskall: Ah :facepalm: , good catch!

module/apmhttp/handler_test.go
kruskall: I've reworked the test :+1:

kruskall: I've moved the test and expanded the test cases :+1:

transaction_test.go
```diff
@@ -85,6 +85,98 @@ func startTransactionInvalidTraceContext(t *testing.T, traceContext apm.TraceCon
...
 }
 
+func TestContinuationStrategy(t *testing.T) {
+	testCases := map[string]struct {
+		traceContext     apm.TraceContext
+		strategy         string
+		expectNewTraceID bool
+		expectSpanLink   bool
+	}{
+		"restart": {
```
axw, line 95: Should we include an "es" tracestate entry for this test, to show that "restart" is indiscriminate?

axw (changes requested):
Thanks, the new test looks great. ICYMI:

> Can you please also add a subtest to TestTracerCentralConfigUpdate in config_test.go?

Then this will be good to merge.

transaction_test.go
kruskall: Great point! Added!

kruskall:
Thank you for the feedback! :bow: 

I added the config test and pushed a few changes :ping_pong:

axw (approved):
LGTM!

Issue #1211, opened 2022-02-08
Agent can send empty transaction and span types


axw:
**Describe the bug**

`StartTransaction("name", "")` will yield a transaction document without a transaction type field. Similar story for `StartSpan("name", "")`. The UI rightfully expects a non-empty transaction type and span type.

**To Reproduce**
1. Using the agent's API, call `StartTransaction("name", "").End()` and `StartSpan("name", "").End()`
2. Wait for ~15 seconds
3. Check that the transaction and span are visible in the APM UI

**Expected behavior**

A default transaction/span type should be set, so the events are visible in the UI.

felixbarny:
The server might be a better way to fix that as it then covers all agents. But fixing it on both sides doesn't hurt, either.

simitt:
created https://github.com/elastic/apm-server/issues/7376

simitt:
from the [agent spec](https://github.com/elastic/apm/blob/main/specs/agents/tracing-spans.md#transaction-and-span-type-and-subtype-fields):

> If no transaction.type or span.type is provided or the value is an empty string, the agent needs to set a default value custom.

PR #1272, opened 2022-07-25
fix: default to custom for empty transaction and span types

kruskall:
According to the spec:

If no transaction.type or span.type is provided or the value
is an empty string, the agent needs to set a default value
custom.

Add test to assert the behaviour is compliant to the spec.

Closes https://github.com/elastic/apm-agent-go/issues/1211

### DIFF
--- a/span_test.go
+++ b/span_test.go
@@ -118,6 +118,22 @@ func TestSpanParentID(t *testing.T) {
 	assert.Equal(t, model.SpanID(parentID), payloads.Spans[0].ParentID)
 }
 
+func TestSpanEnsureType(t *testing.T) {
+	tracer := apmtest.NewRecordingTracer()
+	defer tracer.Close()
+
+	tx := tracer.StartTransaction("name", "type")
+	span := tx.StartSpan("name", "", nil)
+	span.End()
+	tx.End()
+	tracer.Flush(nil)
+
+	payloads := tracer.Payloads()
+	require.Len(t, payloads.Spans, 1)
+
+	assert.NotEmpty(t, payloads.Spans[0].Type)
+}
+
 func TestSpanLink(t *testing.T) {
 	tracer := apmtest.NewRecordingTracer()
 	defer tracer.Close()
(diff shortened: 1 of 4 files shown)


span_test.go
```diff
@@ -118,6 +118,22 @@ func TestSpanParentID(t *testing.T) {
...
+	tx := tracer.StartTransaction("name", "type")
+	span := tx.StartSpan("name", "", nil)
+	span.End()
+	tx.End()
+	tracer.Flush(nil)
+
+	payloads := tracer.Payloads()
+	require.Len(t, payloads.Spans, 1)
+
+	assert.NotEmpty(t, payloads.Spans[0].Type)
```
axw, line 134:
```suggestion
	assert.Equal(t, "custom", payloads.Spans[0].Type)
```

transaction_test.go
```diff
@@ -143,6 +143,19 @@ func TestTransactionEnsureParent(t *testing.T) {
...
+	tracer, transport := transporttest.NewRecorderTracer()
+	defer tracer.Close()
+
+	tx := tracer.StartTransaction("name", "")
+	tx.End()
+	tracer.Flush(nil)
+
+	payloads := transport.Payloads()
+	require.Len(t, payloads.Transactions, 1)
+	assert.NotEmpty(t, payloads.Transactions[0].Type)
```
axw, line 156:
```suggestion
	assert.Equal(t, "custom", payloads.Transactions[0].Type)
```

axw (commented):
Looks good, but the `module/apmot` tests are failing, as they are depending on Type to be empty after construction.

Maybe there's a better alternative, but we I think we may need to set this default value when ending the transaction or span, if the value is empty. That's a bit more foolproof anyway, since users could directly set `Type = ""` after starting the transaction or span.

span_test.go
kruskall: Changed!

transaction_test.go
kruskall: Fixed!

axw (approved):
LGTM!

ugosan:
this should also fix this one https://github.com/elastic/kibana/issues/126805

Issue #1262, opened 2022-07-04
Nil-pointer panic in AWS SDK instrumentation


bengesoff:
**Describe the bug**
We're using the library to instrument our AWS DynamoDB calls, and we're sometimes getting a nil-pointer panic from the AWS SDK instrumentation. It doesn't happen all the time and I can't see how it happens, so my guess is that there is some kind of data race somewhere. I'm afraid I haven't managed to reproduce it in any other environment, or locally, so this won't be the most useful bug report but I didn't know if there might be a way to add a check to avoid the error being a runtime panic. Alternatively someone more familiar with the code might have more of an idea about how it could be triggered?

The code triggering the panic is here:

https://github.com/elastic/apm-agent-go/blob/41de5f09b4be67a9781a9e889a9c249a4c35176d/module/apmawssdkgo/session.go#L165-L167

Somehow the `span` variable is coming out as `nil` from the context, and panicking inside `span.End()` when trying to access the mutex. The `span.Dropped()` check is already verifying whether the `span` is `nil`, and avoiding trying to add it to the DynamoDB API HTTP request, but I think there should also be a check to avoid calling `span.End()` as well.

We're doing a `dynamodb.BatchGetItemWithContext` call, which is working perfectly locally and in some of our environments, but for some reason it's breaking frequently in one specific environment, which doesn't differ much from the others except that it might have some larger instance sizes. I've tried running code with just the DynamoDB call and various combinations of dropping transactions before the call etc, shrinking the APM agent's buffer to the minimum, running many parallel requests, injecting errors into the requests using `localstack`, and I'm afraid I can't reproduce it so I am out of ideas!

The specific error message is:
```
runtime error: invalid memory address or nil pointer dereference
```

The stacktrace is:
```
panic.go in gopanic at line 1038
panic.go in panicmem at line 221
signal_unix.go in sigpanic at line 735
span.go in (*Span).End at line 336
session.go in send at line 165
handlers.go in (*HandlerList).Run at line 267
request.go in (*Request).sendRequest at line 602
request.go in (*Request).Send at line 555
api.go in (*DynamoDB).BatchGetItemWithContext at line 279
```

**To Reproduce**
As above, I'm not sure.

**Expected behavior**
A temporary mitigation would be to check for `nil` before calling `span.End()` to avoid a runtime panic.

axw:
@bengesoff sorry for the delay in responding, this fell through the cracks.

[...]

PR #1273, opened 2022-07-26
module/apmawssdkgo: fix dropped span handling

axw:
In the case where a span is dropped in the "build" hook, we don't add the span to the request context. Subsequently, the "send" and "complete" hooks attempt to fetch the span from context, and modify and end it. This can have two negative outcomes:
 - a panic will occur if there are no other spans in the context
 - if there are other spans in the context, then they will be modified/ended instead of the AWS span that was dropped

I've changed the code to add another context key for the AWS span. The hooks use this instead of `apm.SpanFromContext`.

Fixes https://github.com/elastic/apm-agent-go/issues/1262

### DIFF
--- a/module/apmawssdkgo/session.go
+++ b/module/apmawssdkgo/session.go
@@ -18,7 +18,8 @@
 package apmawssdkgo // import "go.elastic.co/apm/module/apmawssdkgo/v2"
 
 import (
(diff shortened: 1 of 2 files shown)


lahsivjar (approved):
LGTM

Issue #1267, opened 2022-07-13
[META 659] Backend dependencies granularity for NoSQL and Messaging


elastic-apm-tech:
See meta issue for the description and details:
- Meta issue: https://github.com/elastic/apm/issues/659

kruskall:
Fixed by https://github.com/elastic/apm-agent-go/pull/1274, https://github.com/elastic/apm-agent-go/pull/1281, https://github.com/elastic/apm-agent-go/pull/1280 and https://github.com/elastic/apm-agent-go/pull/1278

For compressed spans and dropped spans: https://github.com/elastic/apm-agent-go/issues/1291 and https://github.com/elastic/apm-agent-go/issues/1292

axw:
We still need the fallback handling for Redis at least, reopening until https://github.com/elastic/apm-agent-go/pull/1294 is done

kruskall:
Thank you for spotting this 🙇 

#1294 has been merged. Should we close this ? cc @axw

axw:
Yes, I think we're good to close this now. Just didn't want that part to fall through the cracks :)

axw:
Manually tested apmelasticsearch with 8.5.0-BC5, connecting to ES in Elastic Cloud. Looks good: 
![image](https://user-images.githubusercontent.com/843579/196137313-a1877da6-614d-4925-83dd-4b15398c86c0.png)

I haven't manually tested the others, just checked the code against the spec. The instrumentation is already pretty well tested with integration tests.

PR #1274, opened 2022-07-26
feat: add service target fields

kruskall:
This should lay the groundwork for backend dependencies granularity.

Different backends have different specs/guidelines and the goal is to replace `SetDestinationService` with `SetServiceTarget` and use the new fields. 

Related to https://github.com/elastic/apm-agent-go/issues/1233  
Related to https://github.com/elastic/apm-agent-go/issues/1267

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -434,13 +434,15 @@ type DestinationSpanContext struct {
 	Cloud *DestinationCloudSpanContext `json:"cloud,omitempty"`
 }
 
-// The span service target fields replace the `span.destination.service.*`
-// fields that are deprecated.
+// ServiceSpanContext holds contextual information about the service
+// for a span that relates to an operation involving an external service.
 type ServiceSpanContext struct {
 	// Target holds the destination service.
 	Target *ServiceTargetSpanContext `json:"target,omitempty"`
 }
 
+// ServiceTargetSpanContext fields replace the `span.destination.service.*`
+// fields that are deprecated.
 type ServiceTargetSpanContext struct {
 	// Type holds the destination service type.
 	Type string `json:"type,omitempty"`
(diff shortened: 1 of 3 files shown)


spancontext.go
```diff
@@ -246,6 +267,13 @@ func (c *SpanContext) SetDestinationService(service DestinationServiceSpanContex
 	c.model.Destination = &c.destination
 }
 
+func (c *SpanContext) SetServiceTarget(service ServiceTargetSpanContext) {
```
axw, line 270: Can you please add comments for the new method and type?

Should we add a deprecation notice to `SpanContext.SetDestinationService` too?

spancontext.go
```diff
@@ -60,14 +62,33 @@ type DatabaseSpanContext struct {
 	User string
 }
 
+// The span service target fields replace the `span.destination.service.*`
```
axw, line 65:
```suggestion
// ServiceSpanContext holds ...
//
// The span service target fields replace the `span.destination.service.*`
```

model/model.go
```diff
@@ -431,17 +434,40 @@ type DestinationSpanContext struct {
 	Cloud *DestinationCloudSpanContext `json:"cloud,omitempty"`
 }
 
+// The span service target fields replace the `span.destination.service.*`
```
axw, line 437:
```suggestion
// ServiceSpanContext ...
//
// The span service target fields replace the `span.destination.service.*`
```

axw (commented):
Looks good, just a couple of quibbles about comments. I would like us to follow the guidelines at https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

spancontext.go
kruskall: Good catch! Added :+1:

spancontext.go
kruskall: Ah :facepalm: , thanks for spotting this!

model/model.go
kruskall: Updated!

axw (approved):

PR #1276, opened 2022-07-28
feat: add 'service_target_*' fields to dropped spans metrics

kruskall:
Part of the implementation plan for span service target.

See https://github.com/elastic/apm/blob/main/specs/agents/tracing-spans-service-target.md

Closes https://github.com/elastic/apm-agent-go/issues/1291


axw (commented):
Code looks good. Can you please add a test?

simitt:
@kruskall @axw is this blocked by anything or is the only part missing to add a test?

axw:
AFAICS this just needs a test.

kruskall:
I've updated the test!

Should we remove destination service resource from dropped spans ?

axw (approved):
> Should we remove destination service resource from dropped spans ?

I think we should keep it for now, for compatibility with older servers.

PR #1278, opened 2022-07-28
feat: add service target fields support to aws module

kruskall:
Sets service target fields for services in the aws module,
namely s3 and dynamodb.

See https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-db.md#aws-s3
See https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-db.md#aws-dynamodb

This is part of the work to improve services dependencies granularity.


module/apmawssdkgo/sns.go
```diff
@@ -59,6 +59,8 @@ func (s *apmSNS) spanName() string { return s.name }
 
 func (s *apmSNS) resource() string { return s.resourceName }
 
+func (s *apmSNS) targetName() string { return "" }
```
axw, line 62: https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-messaging.md#span-context-fields says this should be the queue (topic) name.

module/apmawssdkgo/sqs.go
```diff
@@ -72,6 +72,8 @@ func (s *apmSQS) spanName() string { return s.name }
 
 func (s *apmSQS) resource() string { return s.resourceName }
 
+func (s *apmSQS) targetName() string { return "" }
```
axw, line 75: As above, this should be set to the queue name.

axw (commented):
Looks good, may as well add the SNS/SQS target names while we're at it I think? Can you please add tests?

module/apmawssdkgo/sns.go
kruskall: I missed that document! Thank you for pointing it out! :bow:

axw (approved):
LGTM!

PR #1279, opened 2022-07-28
feat: add service target fields support to sql module

kruskall:
Sets service target fields for services using the
sql module.

See https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-db.md#sql-databases


axw (commented):
Code looks good - please add a test :)

axw (approved):
Thanks, LGTM!

axw:
Tested with postgresql (pq), sending data to an 8.5.0-BC5 deployment - looks good. `service.target.name` ("test_db" in my test) is set to the database name, and `service.target.type` is set to the database vendor ("postgresql" in my test). These both show up in the `service_destination metrics:

image

PR #1280, opened 2022-07-28
feat: add service target fields support to azure module

kruskall:
Sets service target fields in the azure module.

See https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-azure.md


module/apmazure/storage.go
```diff
@@ -108,6 +108,10 @@ func (p *apmPipeline) Do(
 	span.Context.SetDestinationService(apm.DestinationServiceSpanContext{
 		Resource: rpc.subtype() + "/" + rpc.storageAccountName(),
 	})
+	span.Context.SetServiceTarget(apm.ServiceTargetSpanContext{
+		Type: rpc.subtype(),
+		Name: rpc.storageAccountName(),
```
axw, line 113: Maybe we need to add another method to the azureRPC interface for this? The spec only mentions setting these fields for Azure Service Bus and Azure Queue Storage, and for those the name should be queue / topic.

Can you please have a dig into the spec, and/or check with the other agent devs, to find out what if anything we should set for other Azure services?

kruskall: That's correct, right now there's no explicit guidance on what those fields should be:

See https://github.com/elastic/apm/issues/661
See https://github.com/elastic/apm/issues/646

From what I could gather, `Type` should fallback to `subtype` if empty.

See https://github.com/elastic/apm/blob/main/specs/agents/tracing-spans-service-target.md#field-values

module/apmazure/storage.go
```diff
@@ -127,6 +131,7 @@ type azureRPC interface {
 	name() string
 	_type() string
 	subtype() string
+	serviceName() string
```
axw, line 134:
```suggestion
	targetName() string
```

Otherwise it sounds a bit like it would return the cloud service name, like "azure service bus".

module/apmazure/storage.go
axw: OK, so in the mean time let's just make sure to use the queue name for Queue storage. Currently we're using the account name, which does not match the updated spec.

We might want to use this to parse the queue name out of request URLs: https://github.com/Azure/azure-storage-queue-go/blob/636801874cdd196abace508ef19c3edca7ed0564/azqueue/parsing_urls.go#L24

module/apmazure/storage.go
kruskall: Good point! Changed :+1:

module/apmazure/storage.go
kruskall: Understood! :+1: 

I went with a lighter solution but looking at this a bit more, it seems the current `name` and `resource` fields are using the account name (https://github.com/elastic/apm-agent-go/blob/main/module/apmazure/storage.go#L109 and https://github.com/elastic/apm-agent-go/blob/main/module/apmazure/queue.go#L36) but they should be using the queue name too.

I've updated the service target fields and left those for a followup PR.

axw (approved):
LGTM, thanks!

PR #1281, opened 2022-07-28
feat: add service target fields support to elasticsearch module

kruskall:
Sets service target fields in the elasticsearch module.

See https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-db.md#elasticsearch


module/apmelasticsearch/client.go
```diff
@@ -92,6 +92,9 @@ func (r *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 		Name:     "elasticsearch",
 		Resource: "elasticsearch",
 	})
+	span.Context.SetServiceTarget(apm.ServiceTargetSpanContext{
+		Type:     "elasticsearch",
```
axw, line 96: Can you please take a look at the effort required to add the Name field too? (Following the Cluster Name logic: https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-db.md#cluster-name)

Extracting `x-found-handling-cluster` and `Host` headers should be fairly trivial.

`go-elasticsearch` has a node discovery loop which queries https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-info.html, so it _could_ extract the cluster name and make it available to clients somehow. Maybe we could request an enhancement.

axw (commented):
I think we should take a look at implementing at least some of the `service.target.name` algorithm too, while we're here. Can you please add a test?

module/apmelasticsearch/client.go
kruskall: Thanks for the links :bow: . I've added handling for `x-found-handling-cluster`.  

I see the cluster name is omitted from both `db.instance` and `service.resource`. Should we update those as well ? (Maybe only `db.instance`, since `service.resource` is deprecated).

I think the request to elasticsearch should be a separate issue since the spec is mentioning a cache but there's not much detail (size ?, TTL ? item expiration ?). Let me know what you think

axw: > I see the cluster name is omitted from both db.instance and service.resource. Should we update those as well ? (Maybe only db.instance, since service.resource is deprecated).

Good catch: let's set `db.instance` to the same value, and leave `destination.service.resource` as it is.

> I think the request to elasticsearch should be a separate issue since the spec is mentioning a cache but there's not much detail (size ?, TTL ? item expiration ?). Let me know what you think

Yes, happy for that to be a separate issue. FWIW, I _think_ the idea is not to query Elasticsearch for the information directly, but to make use of information that is already cached by the official Elasticsearch client.

e.g. the discovery loop I referred to above calls this code, which queries the Elasticsearch `_nodes` API: https://github.com/elastic/elastic-transport-go/blob/8245a9edc85c67c63fd154d9578b01dc2204b3a0/elastictransport/discovery.go#L53

That endpoint returns the cluster name, but the Elasticsearch client does not extract it. So we could ask them to extract it, and somehow make it available. Ideally for us that would be done in a way that avoids a direct dependency on go-elasticsearch.

kruskall: Understood! I've updated the code :+1:

module/apmelasticsearch/client.go
```diff
@@ -92,6 +92,10 @@ func (r *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 		Name:     "elasticsearch",
 		Resource: "elasticsearch",
 	})
+	span.Context.SetServiceTarget(apm.ServiceTargetSpanContext{
+		Type: "elasticsearch",
+		Name: req.Header.Get("x-found-handling-cluster"),
```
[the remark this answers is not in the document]
kruskall: Oh I see, thank you for the explanation :bow:  

Changed!

module/apmelasticsearch/client.go
```diff
@@ -92,7 +92,18 @@ func (r *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 		Name:     "elasticsearch",
 		Resource: "elasticsearch",
 	})
+
+	clustName := req.Header.Get("x-found-handling-cluster")
```
axw, line 96:
```suggestion
	clusterName := req.Header.Get("x-found-handling-cluster")
```

axw (approved):
LGTM, thanks!

PR #1283, opened 2022-08-06
test: add testcase for sqs delete_batch operation

kruskall:
Assert 'span.name' and 'span.action' are following the
spec for delete_batch operations.

The actions were not described in the original spec but
the output seems to be correct so we just add a test to
avoid regressions and verification.

Closes https://github.com/elastic/apm-agent-go/issues/1170


axw (approved):
LGTM!

PR #1285, opened 2022-08-06
feat: rename span_frames_min_duration to span_stack_trace_min_duration

kruskall:
rename the property and update the handling of special
values like 0 and negative values according to the spec.

Closes https://github.com/elastic/apm-agent-go/issues/1142


marclop (commented):
Implementation looks good. Would you mind adding a changelog entry so users are aware that the variable name changed?

kruskall:
Good catch! Added a changelog entry 👍

marclop (approved):

Issue #822, opened 2020-09-30
Expand k8s pod ID discovery regex


eyalkoren:
Implementing elastic/apm#344

PR #1288, opened 2022-08-06
fix: expand k8s pod discovery regex

kruskall:
update k8s pod ID discovery regex to the current status of the spec.

See https://github.com/elastic/apm/blob/main/specs/agents/metadata.md#containerkubernetes-metadata

Add a testcase to make sure the pod mentioned in the related issue is
discovered correctly.

Closes https://github.com/elastic/apm-agent-go/issues/822

### DIFF
--- a/internal/apmhostutil/container_linux.go
+++ b/internal/apmhostutil/container_linux.go
@@ -25,7 +25,6 @@ import (
 	"errors"
 	"io"
 	"os"
-	"path"
 	"regexp"
 	"strings"
 	"sync"
@@ -45,8 +44,8 @@ var (
 
 	kubepodsRegexp = regexp.MustCompile(
 		"" +
-			`(?:^/kubepods[\S]*/pod([^/]+)/$)|` +
-			`(?:kubepods[^/]*-pod([^/]+)\.slice/)`,
+			`(?:^/kubepods[\S]*/pod([^/]+)$)|` +
+			`(?:kubepods[^/]*-pod([^/]+)\.slice)`,
 	)
 
 	containerIDRegexp = regexp.MustCompile(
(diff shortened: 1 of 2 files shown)


internal/apmhostutil/container_linux.go
```diff
@@ -46,12 +46,12 @@ var (
...
-			`(?:^/kubepods\.slice/kubepods-[^/]+\.slice/kubepods-[^/]+-pod([^/]+)\.slice/$)`,
+			`(?:kubepods[^/]*-pod([^/]+)\.slice/)`,
 	)
 
 	containerIDRegexp = regexp.MustCompile(
 		"^" +
-			"[[:xdigit:]]{64}|" +
+			"[[:xdigit:]]{64}$|" +
 			"[[:xdigit:]]{8}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{4,}" +
 			"$",
```
axw, line 56: I think the original intention was that the "^" and "$" were to apply regardless of which of the other lines matched. Possibly there was originally a non-capturing group surrounding those options. Anyway, I think we should add "^" and "$" to each of the lines, so it's:

```go
	containerIDRegexp = regexp.MustCompile(
			"^[[:xdigit:]]{64}$|" +
			"^[[:xdigit:]]{8}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{4,}$",
	)
```

WDYT?

axw (commented):
Should we also add the test case from the description of https://github.com/elastic/apm/pull/508?

internal/apmhostutil/container_linux.go
kruskall: I agree, this should also be more in line with the spec 👍

kruskall:
It turns out that the test case in the description was not recognized as the discovery algorithm was outdated. I've updated the code, added comments and a test case 🏓

internal/apmhostutil/container_linux.go
```diff
@@ -99,32 +97,40 @@ func readCgroupContainerInfo(r io.Reader) (*model.Container, *model.Kubernetes,
...
+				continue
+			}
+		}
+
+		dirname, basename := cgroupPath[:idx], cgroupPath[idx+1:]
+
+		// If the basename ends with ".scope", check for a hyphen and remove everything up to
+		// and including that. This allows us to match .../docker-<container-id>.scope as well
+		// as .../<container-id>.
+		if strings.HasSuffix(basename, ".scope") {
```
axw, line 126: I think we can remove the `systemdScopeSuffix` constant?

axw (approved):

internal/apmhostutil/container_linux.go
kruskall: Good catch!

PR #1290, opened 2022-08-10
tracer: Parse global labels per tracer

marclop:
## Description
This patch removes the `globalLabel` variable and one-time parsing of
the global labels, in favor of parsing the global labels individually
on tracer construction. This has the advantage that tracers that are
created after start-up will parse environment variables at that time
and not only when the `apm` package is imported.

Encountered while writing a systemtest for elastic/apm-server#8139.


axw (approved):
LGTM! Can you please add a changelog entry for this?

kruskall:
Flaky test, I'll trigger another build

marclop:
Thanks!

Issue #776, opened 2020-05-18
False-positive container.id extraction from cgroup info


axw:
I observed a false positive `container.id` field being set when instrumenting an app on my Ubuntu 20.04 machine:

`container.id | spawn-75bc72bd-6642-4cf5-b62c-0674e11bfc84`

Corresponding cgroup line:

```
1:name=systemd:/user.slice/user-1000.slice/user@1000.service/apps.slice/apps-org.gnome.Terminal.slice/vte-spawn-75bc72bd-6642-4cf5-b62c-0674e11bfc84.scope
```

eyalkoren:
Thanks for reporting. This specific line wouldn't cause the creation of a false positive in the Java agent, as it applies similar logic to the Go agent with identifying the ID part of the cgroup path, but it would only create the ID if it matches the allowed ID regexes. 
However, based on our current logic, if the id part of this path would be `spawn-75bc72bd-6642-4cf5-b62c-0674e11bfc84.scope` (without the leading `vte-`), it would. What can you tell about these paths - is this a valid concern? How would you go about fixing that?

axw:
> However, based on our current logic, if the id part of this path would be spawn-75bc72bd-6642-4cf5-b62c-0674e11bfc84.scope (without the leading vte-), it would. What can you tell about these paths - is this a valid concern?

I don't know. This is the first time I'm seeing this issue, but I found it rather than it being reported in the wild, so who knows how prevalent the issue is. Users would probably only notice when looking at actions in Kibana.

> How would you go about fixing that?

Maybe make it more hard-coded: look for docker, crio, etc. I was really hoping to avoid that, but maybe it's inevitable.

PR #1293, opened 2022-08-10
test: verify Ubuntu cgroup line parsing for container ID

kruskall:
Ubuntu cgroup was returning a false positive container.id:

1:name=systemd:/user.slice/user-1000.slice/user@1000.service/apps.slice/apps-org.gnome.Terminal.slice/vte-spawn-75bc72bd-6642-4cf5-b62c-0674e11bfc84.scope

container.id | spawn-75bc72bd-6642-4cf5-b62c-0674e11bfc84

Add a test to verify the line is not returning a false positive
container ID. This is consistent with what other agents are doing
(e.g. the java agent).

Closes https://github.com/elastic/apm-agent-go/issues/776


axw (approved):
Thanks :)

PR #1294, opened 2022-08-11
feat: add fallback service target type for exit spans

kruskall:
This is the final step for complete service target fields
support. Exit spans will fallback to type and subtype similar
to what the deprecated destination resource was doing.

Assertions have been added to tests to verify the service target
type is not overwritten if previously set and that the fallback
mechanism is working properly.

related to https://github.com/elastic/apm-agent-go/pull/1276

Closes https://github.com/elastic/apm-agent-go/pull/1294

### DIFF
--- a/span.go
+++ b/span.go
@@ -350,7 +350,7 @@ func (s *Span) End() {
 		// manually set the destination.service.resource
 		s.setExitSpanDestinationService()
 	}
-	if s.exit && !s.Context.setServiceTargetCalled {
+	if s.exit {
 		// The span was created as an exit span, but the user did not
 		// manually set the service.target fields.
 		s.setExitSpanServiceTarget()
@@ -500,8 +500,32 @@ func (s *Span) setExitSpanServiceTarget() {
 	if fallbackType == "" {
 		fallbackType = s.Type
 	}
+
+	// Service target fields explicitly provided.
+	if s.Context.setServiceTargetCalled {
+		// if the user calls SetServiceTarget with a non-empty name, but empty type,
+		// we'll use the specified name and infer the type
+		if s.Context.serviceTarget.Type == "" && s.Context.serviceTarget.Name != "" {
+			s.Context.SetServiceTarget(ServiceTargetSpanContext{
+				Type: fallbackType,
+				Name: s.Context.serviceTarget.Name,
+			})
+		}
+		return
+	}
+
+	var fallbackName string
+	if s.Context.database.Type != "" { // database spans
+		fallbackName = s.Context.database.Instance
+	} else if s.Context.message.Queue != nil { // messaging spans
+		fallbackName = s.Context.message.Queue.Name
+	} else if s.Context.http.URL != nil { // http spans
+		fallbackName = s.Context.http.URL.Host
+	}
+
 	s.Context.SetServiceTarget(ServiceTargetSpanContext{
 		Type: fallbackType,
+		Name: fallbackName,
 	})
 }
 
--- a/span.go
+++ b/span.go
@@ -350,6 +350,11 @@ func (s *Span) End() {
 		// manually set the destination.service.resource
 		s.setExitSpanDestinationService()
 	}
+	if s.exit && !s.Context.setServiceTargetCalled {
+		// The span was created as an exit span, but the user did not
+		// manually set the service.target fields.
+		s.setExitSpanServiceTarget()
+	}
 	if s.Duration < 0 {
 		s.Duration = time.Since(s.timestamp)
 	}
@@ -487,6 +492,16 @@ func (s *Span) setExitSpanDestinationService() {
 	}
 	s.Context.SetDestinationService(DestinationServiceSpanContext{
 		Resource: resource,
+	})
+}
+
+func (s *Span) setExitSpanServiceTarget() {
+	fallbackType := s.Subtype
+	if fallbackType == "" {
+		fallbackType = s.Type
+	}
+	s.Context.SetServiceTarget(ServiceTargetSpanContext{
+		Type: fallbackType,
 	})
 }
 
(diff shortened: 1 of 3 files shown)


span.go
```diff
@@ -350,6 +350,11 @@ func (s *Span) End() {
 		// manually set the destination.service.resource
 		s.setExitSpanDestinationService()
 	}
+	if s.exit && !s.Context.setServiceTargetCalled {
+		// The span was created as an exit span, but the user did not
+		// manually set the service.target fields.
+		s.setExitSpanServiceTarget()
+	}
```
axw, line 357: This doesn't quite follow the logic described at https://github.com/elastic/apm/blob/main/specs/agents/tracing-spans-service-target.md#field-values. I think users should be able to set just the Name, and have the Type be inferred from the span type/subtype. (I realise this is different to what we had with service destination, but let's just ignore that since it's being replaced.)

Rather than checking if the user has called the method, how about something like this:

```go
if s.exit {
    s.Context.setDefaultServiceTarget(s.Type, s.Subtype)
}

...

func (c *SpanContext) setServiceTargetDefaults(spanType, spanSubtype string) {
    if c.serviceTarget.Type == "" {
        c.serviceTarget = spanSubtype
        if c.serviceTarget.Type == "" {
            c.serviceTarget = spanType
        }
    }
    if c.serviceTarget.Name == "" {
        // ...
    }
    c.SetServiceTarget(ServiceTargetSpanContext{Type: c.serviceTarget.Type, Name: c.serviceTarget.Name})
}
```

kruskall: Ok, just a question to clarify: it seems that would infer the value even if an explicit empty name/type was set. Is that intended ?

axw: I thought so, since the type can always be inferred and the name is optional, but the spec also says this:

> Agents SHOULD provide an API entrypoint to set the value of span.context.service.target.type and span.context.service.target.name, setting an empty or null value on both of those fields allows the user to discard the inferred values.

@SylvainJuge why would a user need to disable the inference of `service.target.type` or `service.target.name`?

SylvainJuge: Hi !

I remember it was requested at some point for the Java agent in order to provide a work-around when our heuristics fail and or the user wants more control on how things are named (and maybe ignore a few dependencies):
- allow to split a dependency when the heuristic groups everything (for example when calls are made to a gateway, we currently only show the gateway, the user might want to manually split if the app has some knowledge about this).
- allow to limit cardinality : for example a single database server with dozens of databases that might be better grouped together, or a web-crawling app that has one dependency per DNS entry by default.

axw: @SylvainJuge let me know if I understand correctly: what you're saying is that in some cases the inferred type & name may not be ideal, so users may wish to override the values and provide either more specific values (e.g. in the case of a gateway) or less specific values (e.g. in the case of many databases instances).

That's a good argument for allowing users to manually set non-empty values, but I don't see why we should completely omit the `service.target.*` fields if a user sets them both to null/empty values.

SylvainJuge: > @SylvainJuge let me know if I understand correctly: what you're saying is that in some cases the inferred type & name may not be ideal, so users may wish to override the values and provide either more specific values (e.g. in the case of a gateway) or less specific values (e.g. in the case of many databases instances).

Yes, that's exactly what I meant.

> That's a good argument for allowing users to manually set non-empty values, but I don't see why we should completely omit the `service.target.*` fields if a user sets them both to null/empty values.

Omitting the fields when empty would be consistent with what we do for `resource` field ([spec](https://github.com/elastic/apm/blob/main/specs/agents/tracing-spans-destination.md)) for `context.destination.service.resource` :

> Agents SHOULD offer a public API to set this field so that users can customize the value if the generic mapping is not
sufficient. If set to `null` or an empty value, agents MUST omit the `span.destination.service` field altogether, thus
providing a way to manually disable the automatic setting/inference of this field (e.g. in order to remove a node
from a service map or an external service from the dependencies table).

Also, when set to empty/null values, omitting both `destination.service.resource` and `service.target.*` would avoid making the apm-server try to infer any of those fields which could lead to inconsitencies.

axw: Thanks @SylvainJuge.

How about this then:
 - if the user does not call SetServiceTarget at all, we'll use inferred values for both (where possible)
 - if the user calls SetServiceTarget with an empty name, this will disable inference
 - if the user calls SetServiceTarget with a non-empty name, but empty type, we'll use the specified name and infer the type

I think that covers the intended use case of controlling the `service.target.name` cardinality.

kruskall: > How about this then:
> 
>     * if the user does not call SetServiceTarget at all, we'll use inferred values for both (where possible)
> 
>     * if the user calls SetServiceTarget with an empty name, this will disable inference
> 
>     * if the user calls SetServiceTarget with a non-empty name, but empty type, we'll use the specified name and infer the type
> 
> 
> I think that covers the intended use case of controlling the `service.target.name` cardinality.

I've updated the code to reflect this!

span_test.go
```diff
@@ -227,6 +227,10 @@ func TestStartExitSpan(t *testing.T) {
 	// the exit span will assign the value.
 	assert.Equal(t, spans[0].Context.Destination.Service.Resource, "type")
 
+	// When the context's DestinationService is not explicitly set, ending
```
axw, line 230:
```suggestion
	// When the context's ServiceTarget is not explicitly set, ending
```

span.go
```diff
@@ -490,6 +495,40 @@ func (s *Span) setExitSpanDestinationService() {
...
+	if fallbackType == "" {
+		fallbackType = s.Type
+	}
+
+	// Service target fields explicitly provided.
+	if s.Context.setServiceTargetCalled {
+		// if the user calls SetServiceTarget with a non-empty name, but empty type,
+		// we'll use the specified name and infer the type
+		if s.Context.serviceTarget.Type == "" && s.Context.serviceTarget.Name != "" {
+			s.Context.SetServiceTarget(ServiceTargetSpanContext{
```
axw, line 509: Should we have a unit test for this case, where SetServiceTarget is called with only a name?

axw (approved):
LGTM :) Just a couple of minor things

span.go
kruskall: Good point!

PR #1295, opened 2022-08-12
fix: middleware panics on ended transaction

eugene-trifonov:
When transaction is ended already, apmechov4 middleware panics, because `tx.Result` is `tx.TransactionData.Result`, and `TransactionData == nil` when transaction is ended.
This commit fixes the problem.

### DIFF
--- a/module/apmechov4/middleware.go
+++ b/module/apmechov4/middleware.go
@@ -81,7 +81,10 @@ func (m *middleware) handle(c echo.Context) error {
 	}
 	name := m.requestName(c)
 	tx, body, req := apmhttp.StartTransactionWithBody(m.tracer, name, req)
-	defer tx.End()
+	defer func() {
+		body.Discard()
+		tx.End()
+	}()
 	c.SetRequest(req)
 
 	resp := c.Response()
@@ -106,11 +109,14 @@ func (m *middleware) handle(c echo.Context) error {
 			e.Handled = true
 			e.Send()
 		}
+
+		if txEnded := tx.TransactionData == nil; txEnded {
+			return
+		}
 		tx.Result = apmhttp.StatusCodeResult(resp.Status)
 		if tx.Sampled() {
 			setContext(&tx.Context, req, resp, body)
 		}
-		body.Discard()
 	}()
 
 	handlerErr = m.handler(c)
--- a/module/apmechov4/middleware_test.go
+++ b/module/apmechov4/middleware_test.go
@@ -165,6 +165,21 @@ func TestEchoMiddlewarePanic(t *testing.T) {
 	assertError(t, transport.Payloads(), "handlePanic", "boom", false)
 }
 
+func TestEchoMiddlewareEndedTxn(t *testing.T) {
+	tracer, _ := transporttest.NewRecorderTracer()
+	defer tracer.Close()
+
+	e := echo.New()
+	e.Use(apmecho.Middleware(apmecho.WithTracer(tracer)))
+	e.GET("/end-txn", handleEndTxn)
+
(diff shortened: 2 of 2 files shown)


eugene-trifonov:
> ❌ Author of the following commits did not sign a [Contributor Agreement](https://www.elastic.co/contributor-agreement): [27e8874](https://github.com/elastic/apm-agent-go/commit/27e8874a8c0afb1471cf04717e8499d5091013ce)
> 
> Please, read and sign the above mentioned agreement if you want to contribute to this project

Just signed one. But seems this check still shows ❌ :(

axw:
@eugene-trifonov thanks for opening the PR!

What is the use case for ending the transaction in the handler, and not in the middleware? It's expected that the middleware ends the transaction.

eugene-trifonov:
> What is the use case for ending the transaction in the handler, and not in the middleware? It's expected that the middleware ends the transaction.

I found it in case when http connection is patched to be websocket connection. And as a result such transaction longs too much time (hours).
I wanted to split such huge transaction to smaller ones to make it more convenient for reviewing, and first transaction is closing right before I patch http connection to websocket connection.
Once the websocket connection is ended, I receive panic from this middleware.

axw:
@eugene-trifonov thanks for the info.

> I wanted to split such huge transaction to smaller ones to make it more convenient for reviewing, and first transaction is closing right before I patch http connection to websocket connection.

Do you still want to capture the transaction created in the middleware, or would it be an option to not create a transaction in the first place for that route?

If that is an option, you could use https://pkg.go.dev/go.elastic.co/apm/module/apmechov4/v2#WithRequestIgnorer to not create these transactions. Then you can create transactions manually in your websocket handler.

eugene-trifonov:
> @eugene-trifonov thanks for the info.
> 
> > I wanted to split such huge transaction to smaller ones to make it more convenient for reviewing, and first transaction is closing right before I patch http connection to websocket connection.
> 
> Do you still want to capture the transaction created in the middleware, or would it be an option to not create a transaction in the first place for that route?
> 
> If that is an option, you could use https://pkg.go.dev/go.elastic.co/apm/module/apmechov4/v2#WithRequestIgnorer to not create these transactions. Then you can create transactions manually in your websocket handler.

Yeah, I could, actually I've tried already, but such solution has some duplications I would want to have.
Currently I use apm middleware before logging middleware, so any log has `trace.id`/`transaction.id`, and logger writes log about call to endpoint. So either I need to copy-paste logging, checking error for different logging levels and tracing ending with/without error, either I fix apm middleware.

I have chosen second option, because I was thinking it's not cool that middleware can panic and have memory leak (on `body.Discard()`) just because transaction is ended already. This fix just adds flexibility to the middleware I think.

axw:
> Currently I use apm middleware before logging middleware, so any log has trace.id/transaction.id, and logger writes log about call to endpoint. So either I need to copy-paste logging, checking error for different logging levels and tracing ending with/without error, either I fix apm middleware.

Won't that `transaction.id` be wrong/irrelevant though? i.e. it would refer to the automatically-generated transaction, which I assume you're discarding?

> I have chosen second option, because I was thinking it's not cool that middleware can panic and have memory leak (on body.Discard() isn't calling when panics) just because transaction is ended already. This fix just adds flexibility to the middleware I think.

Yes it adds flexibility, but it's also adding a bit of complexity and sets a precedent for other middleware - hence why I'm a bit wary of the change. I don't think it's unreasonable for code to panic if used in a way that is not intended. I understand it doesn't work for your use case so let's try and explore different options.

Would you be able to share a small program which exhibits the problem? Ideally with the logging you're expecting as well.

PR #1297, opened 2022-08-12
fix: readd deprecated span_frames_min_duration option as fallback for older configuration

kruskall:
We cannot remove a config option right away as that's a breaking change
and it would create issue for older agents updating.
Add a deprecation notice to the config doc and add the old option as fallback.

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -426,6 +426,25 @@ The minimum duration allowed for this setting is 1 microsecond (`us`).
 
 [float]
 [[config-span-frames-min-duration-ms]]
+=== `ELASTIC_APM_SPAN_FRAMES_MIN_DURATION`
+
+<<dynamic-configuration, image:./images/dynamic-config.svg[] >>
+
+[options="header"]
+|============
+| Environment                                 | Default
+| `ELASTIC_APM_SPAN_FRAMES_MIN_DURATION`      | `5ms`
+|============
+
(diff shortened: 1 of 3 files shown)


docs/configuration.asciidoc
```diff
@@ -441,6 +441,9 @@ exceeds this configured value. While this is very helpful to find the exact
 place in your code that causes the span, collecting this stack trace does have
 some processing and storage overhead.
 
+NOTE: This configuration was previously known as `ELASTIC_APM_SPAN_FRAMES_MIN_DURATION`, which has been deprecated and will be removed in a future major
```
axw, line 444: Can you please add the old docs back in? While doing that, please rename this config's anchor to `[[config-span-stack-trace-min-duration]]`, and use `[[config-span-frames-min-duration-ms]]` for the deprecated config.

There's special syntax for deprecation notices which we should probably be using: https://github.com/elastic/docs#additions-and-deprecations

axw (changes requested):
Thanks! Code LGTM, I'd like to duplicate the docs for now though. We can remove the deprecated config handling & docs in the next major version of the agent.

docs/configuration.asciidoc
kruskall: Oh I see, thanks for spotting this! I've updated the doc 👍

axw (approved):
Thanks :)

PR #1299, opened 2022-08-16
docs: update correct env flag for loglevel

vigneshshanmugam:
+ fixes the env flag for log level in docs.


kruskall (approved):
Good catch!

Issue #854, opened 2020-11-22
Add support for native pgx database driver


StianOvrevage:
Related to https://github.com/elastic/apm-agent-go/issues/851 .

You have added support for the `database/sql` mode of `pgx` (a regular replacement for the deprecated `lib/pq`) which is great!

But the native mode of `pgx`, which is not compatible with `database/sql` has some very nice improvements that we use. Much easier type handling and support for batching queries and some other nice performance improvements.

It would be great to see support for the native mode of `pgx` as well :)

PR #1301, opened 2022-08-20
feat(module): add module for tracing SQL queries created by jackc/pgx

gvencadze:
This module provide enhanced support of `jackc/pgx` library and it's features such as `COPY FROM` and `BATCH` statements

Closes https://github.com/elastic/apm-agent-go/issues/854


axw:
@gvencadze sorry for the silence, this slipped my attention.

Can you please sign the CLA? As there are commits from multiple authors, both you and @EpicStep will need to do this.

gvencadze:
@axw hey, I've signed (tried with both of emails that's linked to account), but nothing changed. Do I need to restart pipeline or something else?

module/apmpgx/tracer.go
```diff
@@ -0,0 +1,162 @@
...
+		Type: postgresql,
+	})
+
+	if _, ok := data["err"]; ok {
+		e := apm.CaptureError(ctx, data["err"].(error))
+		e.Timestamp = stop
+		e.Send()
+	}
+
+	span.End()
```
axw, line 135: Most of the code is duplicated for each of the three methods - the main that differs is the span name and type.

Could you please take a look at extracting the common code into a separate method? e.g.

```go
// BatchTrace traces batch execution and creates spans for the whole batch.
func (t *tracer) BatchTrace(ctx context.Context, data map[string]interface{}) {
	span, ok := t.newSpan(ctx, "BATCH", batchSpanType)
	if !ok {
		return
	}
	defer span.End()

	if batchLen, ok := data["batchLen"].(int); ok {
		span.Context.SetLabel("batch.length", batchLen)
	}
}

func (t *tracer) newSpan(ctx context.Context, spanName, spanType string) (*apm.Span, bool) {
	stop := time.Now()

	duration, ok := data["time"].(time.Duration)
	if !ok {
		apm.CaptureError(ctx, ErrUnsupportedPgxVersion).Send()
		return nil, false
	}

	span, _ := apm.StartSpanOptions(ctx, spanName, spanType, apm.SpanOptions{
		Start: stop.Add(-duration),
		ExitSpan: true,
	})
	if span.Dropped() {
		span.End()
		return nil, false
	}

	span.Duration = duration
	span.Context.SetDatabase(apm.DatabaseSpanContext{
		Type: postgresql,
	})

	if err, ok := data["err"].(error); ok {
		e := apm.CaptureError(ctx, err)
		e.Timestamp = stop
		e.SetSpan(span)
		e.Send()
	}

	return span, true
}
```

module/apmpgx/tracer.go
```diff
@@ -0,0 +1,162 @@
...
+
+	span, _ := apm.StartSpanOptions(ctx, apmsql.QuerySignature(data["sql"].(string)), querySpanType, apm.SpanOptions{
+		Start: stop.Add(-data["time"].(time.Duration)),
+	})
+
+	span.Duration = data["time"].(time.Duration)
+	span.Context.SetDatabase(apm.DatabaseSpanContext{
+		Type:      postgresql,
+		Statement: data["sql"].(string),
+	})
```
axw, line 99: Type should be "sql"; the specific vendor can be derived from the span subtype.

This metadata is quite limited compared to what `apmsql` sets:

https://github.com/elastic/apm-agent-go/blob/99546a3cffd13007614d18aa926bf4dcc6c8d240/module/apmsql/conn.go#L73-L89

What do you think about either passing a [pgconn](https://pkg.go.dev/github.com/jackc/pgconn).[Config](https://pkg.go.dev/github.com/jackc/pgconn#Config) into `apmpgx.NewTracer`, or changing the signature to something like this:

```go
package apmpgx

// Instrument wraps cfg.Logger with an implementation of pgx.Logger that traces database operations.
func Instrument(cfg *pgx.ConnConfig) {
    // extract metadata from cfg.Config, update cfg.Logger
}
```

axw (changes requested):
This looks great, thanks for the PR! I've left some comments, and in addition to those:

 - Would you please take a look at adding an integration test? You can adapt the apmsql/pq one here: https://github.com/elastic/apm-agent-go/blob/main/module/apmsql/pq/pq_test.go
 - Please run `make update-licenses update-modules fmt scripts/Dockerfile-testing`
 - Please add vanity import comments to the files, so that `make check` passes

gvencadze:
@axw hey, I've done what you have write above:
1. integration tests are already presented in `tracer_test.go`, they're located under `TestLog`
2. `make update-licenses update-modules fmt scripts/Dockerfile-testing` - done
3. vanity import - done

module/apmpgx/tracer.go
```diff
@@ -0,0 +1,173 @@
...
+	tableName, ok := data["tableName"].(pgx.Identifier)
+	if !ok {
+		return
+	}
+
+	span, ok := t.startSpan(ctx,
+		fmt.Sprintf("COPY TO %s", strings.Join(tableName, ", ")),
+		copySpanType,
+		data,
+	)
```
axw, line 119:
```suggestion
	)
	if !ok {
		return
	}
```

module/apmpgx/tracer.go
```diff
@@ -0,0 +1,173 @@
...
+
+	if span.Dropped() {
+		span.End()
+		return nil, false
+	}
+
+	span.Duration = duration
+	span.Context.SetDatabase(apm.DatabaseSpanContext{
+		Instance:  t.cfg.Database,
+		Statement: spanName,
```
axw, line 158: Should we add another parameter to `startSpan` for passing in the statement? For QueryTrace, spanName will not be the full statement/query. This field is meant to capture the full query.

docs/instrumenting.asciidoc
```diff
@@ -969,3 +970,53 @@ func main() {
...
+import (
+    "github.com/jackc/pgx/v4/pgxpool"
+
+    "go.elastic.com/apm/module/apmpgx/v2"
+)
+
+func main() {
+    c, err := pgxpool.ParseConfig("dsn_string")
+    ...
+    c.ConnConfig.Logger = apmpgx.NewTracer(nil)
```
axw, line 1000: Please update this to the new Instrument function.

docs/instrumenting.asciidoc
```diff
@@ -969,3 +970,53 @@ func main() {
...
+import (
+    "github.com/jackc/pgx/v4"
+
+    "go.elastic.com/apm/module/apmpgx/v2"
+)
+
+func main() {
+    c, err := pgx.Connect(context.Background, "dsn_string")
+    ...
+    c.Config().Logger = apmpgx.NewTracer(nil)
```
axw, line 1019: Please update this to the new Instrument function.

axw (changes requested):
Thanks @gvencadze, it's looking good. Just a couple more things.

> integration tests are already presented in tracer_test.go, they're located under TestLog

I would call those unit tests, as they're calling the logger functions directly. What I meant was a test that runs with a real PostgreSQL database, instruments pgx, calls the pgx API methods, and then checks that the expected spans are produced.

That's what https://github.com/elastic/apm-agent-go/blob/99546a3cffd13007614d18aa926bf4dcc6c8d240/module/apmsql/pq/pq_test.go#L34-L79 does, using the `database/sql` API.

I don't think it's strictly necessary for this PR, so if it's too much to ask then we can come back to it later.

module/apmpgx/tracer.go
gvencadze: I've added this parameter, but not sure that it's required for Copy and Batch trace, so provide empty string for them. We'll discuss this today with @EpicStep

gvencadze:
@axw I've added e2e tests for Query and Copy tracing, only the batch trace remain. Can you take a look? Maybe you will find something else that should be changed/discussed

axw (approved):
LGTM! Thank you, this looks great.

axw:
@gvencadze there's an issue with formatting, CI is failing with this error:

```
[2022-09-27T17:20:32.075Z] goimports differs:
[2022-09-27T17:20:32.076Z]  - module/apmpgx/e2e_test.go
```

Can you please make sure `make check-full` passes locally?

gvencadze:
> @gvencadze there's an issue with formatting, CI is failing with this error:
> 
> ```
> [2022-09-27T17:20:32.075Z] goimports differs:
> [2022-09-27T17:20:32.076Z]  - module/apmpgx/e2e_test.go
> ```
> 
> Can you please make sure `make check-full` passes locally?

Sure, I’ll take a look

module/apmpgx/tracer.go
```diff
@@ -0,0 +1,185 @@
...
+	statement, ok := data["sql"].(string)
+	if !ok {
+		apm.CaptureError(ctx,
+			fmt.Errorf("%w: expect string, got: %T",
+				ErrInvalidType,
+				data["sql"],
+			),
+		).Send()
+		return
+	}
```
gvencadze, line 99: @axw one more question - should I capture and send type casting errors into APM? I didn't find something like this in other modules, but I think it will be good for understanding that something inside module (or its dependency) is broken

axw: There's no precedent and I'm a bit on the fence about it, but I think it's fine to keep it for now at least. There's not really a good alternative at the moment. Generally errors are about the service, rather than about the instrumentation. I suppose ideally there would be some way to log errors about the instrumentation - we can revise it later.

axw:
@gvencadze I see you've sorted out the last of the tests :tada:

One last thing before I merge it: would you please add an entry to CHANGELOG.asciidoc, to the "Unreleased" section?

gvencadze:
@axw done

axw:
run elasticsearch-ci/docs

axw:
> They aren't available in V4 version, so I think we should make another one module, something called apmpgxv5 for example

I agree, that sounds like the way to go. Yes, please do open a separate issue and reference this PR.

axw:
Thank you very much for your contribution!

Issue #1310, opened 2022-09-13
apmelasticsearch: X-Found-Handling-Cluster should be extracted from response headers


axw:
I just noticed that the `X-Found-Handling-Cluster` header is being extracted from the request headers; it should be extracted from the _response_ headers: https://github.com/elastic/apm-agent-go/blob/831decc659a053ff6ce9f1c061001030e7f4ecb7/module/apmelasticsearch/client.go#L96

See https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-db.md#cluster-name

PR #1311, opened 2022-09-13
apmelasticsearch: query ES for cluster name

axw:
Follow on from https://github.com/elastic/apm-agent-go/pull/1309

Fix extraction of cluster name from `X-Found-Handling-Cluster` header by pulling from response headers rather than request headers.

If the response header is not present, query the nodes API and extract the `cluster_name` field from the response body. This is cached in an LRU (with a fixed size of 100), keyed by the `Host` request header. We cache failures, and there is currently no retrying.

Closes https://github.com/elastic/apm-agent-go/issues/1310

### DIFF
--- a/module/apmelasticsearch/clustername.go
+++ b/module/apmelasticsearch/clustername.go
@@ -1,0 +1,193 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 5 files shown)


module/apmelasticsearch/clustername.go
```diff
@@ -0,0 +1,193 @@
...
+		// Query /_nodes/http. This request is expected to fail for
+		// clients that are not authorized.
+		u := url.URL{
+			Scheme: requestURL.Scheme,
+			User:   requestURL.User,
+			Host:   requestURL.Host,
+			Path:   "/_nodes/http",
+		}
+
+		req, err := http.NewRequest(http.MethodGet, u.String(), nil)
```
marclop, line 72: Should we use the context that is passed as an argument for the request?
```suggestion
		req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
```

marclop (approved):

module/apmelasticsearch/clustername.go
axw: Yikes, we should indeed. Thanks for the catch!

Issue #1315, opened 2022-09-19
Spans representing calls to external services are not being marked as exit spans


axw:
**Describe the bug**

https://github.com/elastic/apm/blob/main/specs/agents/tracing-spans.md#exit-spans says

> Exit spans are spans that describe a call to an external service, such as an outgoing HTTP request or a call to a database.

We are only setting `ExitSpan: true` in module/apmredigov8. We should update all other modules that create these kinds of spans.

Note that the spec also says:

> A span is considered an exit span if it has explicitly been marked as such; a span's status should not be inferred.

This does not apply to spans created by built-in instrumentation; it applies to custom spans created through the API.

**Expected behavior**

We should be marking spans representing calls to external services as exit spans.

PR #1317, opened 2022-09-20
fix: mark spans representing calls to external services as exit spans

kruskall:
Exit spans are spans that describe a call to an external service, such as an outgoing HTTP request or a call to a database.

Marking a span as an exit span also triggers service target fields and destination fields inference.
Update tests to reflect the new fields.

Disable compression in some tests as exit spans can be aggregated based on certain fields, leading to unexpected results.

Additionally, make sure exit spans are not being used as parents of other spans as that would cause the spans to be dropped.

Closes https://github.com/elastic/apm-agent-go/issues/1315

### DIFF
--- a/apmtest/withtransaction.go
+++ b/apmtest/withtransaction.go
@@ -29,6 +29,16 @@ func WithTransaction(f func(ctx context.Context)) (model.Transaction, []model.Sp
 	return WithTransactionOptions(apm.TransactionOptions{}, f)
 }
 
+// WithUncompressedTransaction is equivalent to calling WithTransactionOptions with compression disabled.
+func WithUncompressedTransaction(f func(ctx context.Context)) (model.Transaction, []model.Span, []model.Error) {
+	tracer := NewRecordingTracer()
+	// Do not drop short exit spans by default.
+	tracer.SetExitSpanMinDuration(0)
+	tracer.SetSpanCompressionEnabled(false)
+	defer tracer.Close()
+	return tracer.WithTransactionOptions(apm.TransactionOptions{}, f)
+}
+
 // WithTransactionOptions calls f with a new context containing a transaction
 // and transaction options, flushes the transaction to a test server, and returns
 // the decoded transaction and any associated spans and errors.
--- a/module/apmgormv2/apmgorm_test.go
+++ b/module/apmgormv2/apmgorm_test.go
@@ -84,7 +84,7 @@ func TestWithContext(t *testing.T) {
 }
 
 func testWithContext(t *testing.T, dsnInfo apmsql.DSNInfo, dialect gorm.Dialector, config *gorm.Config) {
-	_, spans, errors := apmtest.WithTransaction(func(ctx context.Context) {
+	_, spans, errors := apmtest.WithUncompressedTransaction(func(ctx context.Context) {
 		db, err := gorm.Open(dialect, config)
 		require.NoError(t, err)
 		ddb, _ := db.DB()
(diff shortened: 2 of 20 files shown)


apmtest/withtransaction.go
```diff
@@ -29,6 +29,16 @@ func WithTransaction(f func(ctx context.Context)) (model.Transaction, []model.Sp
 	return WithTransactionOptions(apm.TransactionOptions{}, f)
 }
 
+// WithUncompressedTransaction is equivalent to calling WithTransactionOptions with compression disabled.
```
axw, line 32: nice :)

module/apmhttp/clienttrace.go
```diff
@@ -60,7 +60,7 @@ func withClientTrace(ctx context.Context, tx *apm.Transaction, parent *apm.Span)
 			r.mu.Lock()
 			defer r.mu.Unlock()
 			if !r.ended {
-				r.DNS = tx.StartSpan(fmt.Sprintf("DNS %s", i.Host), "external.http.dns", parent)
+				r.DNS = tx.StartExitSpan(fmt.Sprintf("DNS %s", i.Host), "external.http.dns", parent)
```
axw, line 63: The spec says:

> Exit spans MAY have child spans that have the same type and subtype. For example, an HTTP exit span may have child spans with the action request, response, connect, dns. These spans MUST NOT have any destination context, so that there's no effect on destination metrics.

If they mustn't have destination context, does that mean they should not be exit spans?

module/apmhttp/client.go
```diff
@@ -104,7 +104,15 @@ func (r *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
...
-	span := tx.StartSpan(name, r.spanType, apm.SpanFromContext(ctx))
+	var span *apm.Span
+	if r.traceRequests {
+		// Trace requests are enabled, use this as the parent span and mark the trace requests as
+		// exit spans.
+		span = tx.StartSpan(name, r.spanType, apm.SpanFromContext(ctx))
+	} else {
+		// No trace requests, mark the span as exit span.
+		span = tx.StartExitSpan(name, r.spanType, apm.SpanFromContext(ctx))
+	}
```
axw, line 115: See my other comment in clienttrace.go - I'm pretty sure this span should always be an exit span. Because the net/http/httptrace spans have the same type and subtype, only a different action, they can be children of the exit span.

module/apmgormv2/apmgorm_test.go
```diff
@@ -121,12 +121,10 @@ func testWithContext(t *testing.T, dsnInfo apmsql.DSNInfo, dialect gorm.Dialecto
...
 		assert.Equal(t, dsnInfo.User, span.Context.Database.User)
-		if dsnInfo.Address == "" {
-			assert.Nil(t, span.Context.Destination)
-		} else {
-			assert.Equal(t, dsnInfo.Address, span.Context.Destination.Address)
-			assert.Equal(t, dsnInfo.Port, span.Context.Destination.Port)
-		}
+		assert.Equal(t, dsnInfo.Address, span.Context.Destination.Address)
+		assert.Equal(t, dsnInfo.Port, span.Context.Destination.Port)
+		assert.Equal(t, dsnInfo.User, span.Context.Destination.Service.Name)
```
axw, line 126:
```suggestion
		assert.Equal(t, dsnInfo.Database, span.Context.Destination.Service.Name)
```

Right?

axw (commented):
Thanks, this is great. Just a few questions.

module/apmgormv2/apmgorm_test.go
kruskall: This is actually destination service name, the old deprecated field. It should use the driver name.

module/apmhttp/client.go
kruskall: Good point! Updated!

module/apmhttp/clienttrace.go
kruskall: It is not specified but I think it makes sense for them not to be exit spans, especially considering the related features (compression, etc.)

kruskall:
This uncovered another bug with the agent which is not respecting the spec.

Blocked by https://github.com/elastic/apm-agent-go/pull/1320

axw (commented):
Thanks for the changes! LGTM, modulo the httptrace issue you uncovered.

kruskall:
> This uncovered another bug with the agent which is not respecting the spec.
> 
> Blocked by #1320

PR merged. Unblocking

axw (approved):
LGTM. I think the integration tests need updating: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-go%2Fapm-agent-go-mbp/detail/PR-1317/8/pipeline

axw:
There's a few cases where we're not marking spans as exit spans, and I found a panic related to ending a child of an exit span. I'll open a PR.

Issue #1319, opened 2022-09-22
Child spans of exit spans are silently dropped


kruskall:
**Describe the bug**

The agent is silently dropping child spans of exit spans. This is blocking https://github.com/elastic/apm-agent-go/pull/1317 because of trace requests in the apmhttp module being child spans of an exit span.

**Expected behavior**

The spec clarified child spans handling with https://github.com/elastic/apm/pull/433

> Exit spans MUST not have child spans that have a different `type` or `subtype`.

> Exit spans MAY have child spans that have the same `type` and `subtype`.
> For example, an HTTP exit span may have child spans with the `action` `request`, `response`, `connect`, `dns`.
> These spans MUST NOT have any destination context, so that there's no effect on destination metrics.

simitt:
@kruskall we are past the `8.5` milestone, I add it to `8.6` though.

PR #1320, opened 2022-09-22
fix: allow exit spans to have child spans with the same type and subtype

kruskall:
Do not silently drop child spans of exit spans. The spec allows for exit spans to have child spans if they have the same type and subtype. One example of this is http trace requests being child spans of an exit span.

Update exit span test to make sure the feature is working as intended.

Closes https://github.com/elastic/apm-agent-go/issues/1319

### DIFF
--- a/span.go
+++ b/span.go
@@ -70,8 +70,14 @@ func (tx *Transaction) StartSpanOptions(name, spanType string, opts SpanOptions)
 	}
 
 	if opts.parent.IsExitSpan() {
-		spanType, after, _ := strings.Cut(spanType, ".")
-		subType, _, _ := strings.Cut(after, ".")
+		arr := strings.SplitN(spanType, ".", 3)
+
+		spanType := arr[0]
+		subType := ""
+
+		if len(arr) > 1 {
+			subType = arr[1]
+		}
 
 		if spanType != opts.parent.Type || subType != opts.parent.Subtype {
 			return newDroppedSpan()
--- a/span.go
+++ b/span.go
@@ -69,21 +69,6 @@ func (tx *Transaction) StartSpanOptions(name, spanType string, opts SpanOptions)
 		return newDroppedSpan()
 	}
 
-	if opts.parent.IsExitSpan() {
-		arr := strings.SplitN(spanType, ".", 3)
-
-		spanType := arr[0]
-		subType := ""
-
-		if len(arr) > 1 {
-			subType = arr[1]
-		}
-
-		if spanType != opts.parent.Type || subType != opts.parent.Subtype {
-			return newDroppedSpan()
-		}
-	}
-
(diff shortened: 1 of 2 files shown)


span.go
```diff
@@ -65,10 +65,25 @@ func (tx *Transaction) StartSpan(name, spanType string, parent *Span) *Span {
...
 
+	if opts.parent.IsExitSpan() {
+		arr := strings.SplitN(spanType, ".", 3)
+
+		spanType := arr[0]
+		subType := ""
+
+		if len(arr) > 1 {
+			subType = arr[1]
+		}
```
axw, line 80: There's already code for splitting type/subtype/action. Should we extract that into a function that can be reused here?

https://github.com/elastic/apm-agent-go/blob/31ed25dde1519583f6a6160f47976be428bace87/span.go#L240-L246

axw (commented):
Should we be checking the type/subtype in `Span.End`, rather than at creation time? Type/Subtype are mutable.

Also, should we check that the child span is not an exit span? The spec says that the child span must not have any destination context. That's to ensure we don't count both the exit span and its children in destination metrics.

What should happen if a child is created for a child of an exit span?

span.go
kruskall: This no longer applies

kruskall:
> Should we be checking the type/subtype in `Span.End`, rather than at creation time? Type/Subtype are mutable.

Good point! Changed.


> The spec says that the child span must not have any destination context. That's to ensure we don't count both the exit span and its children in destination metrics.

Updated the code to remove destination context :+1:  

> Also, should we check that the child span is not an exit span? 
> What should happen if a child is created for a child of an exit span?

The spec does not forbid that. The status of the child span has no effect on the decision of dropping the span.

span.go
```diff
@@ -345,6 +346,14 @@ func (s *Span) End() {
 	if s.Type == "" {
 		s.Type = "custom"
 	}
+	if s.parent.IsExitSpan() {
+		if s.Type != s.parent.Type || s.Subtype != s.parent.Subtype {
+			s.tracer = nil
+			return
+		}
+
+		s.Context.destination = model.DestinationSpanContext{}
```
axw, line 355:
```suggestion
		s.Context.model.Destination = nil
		s.Context.model.Service = nil
```

span.go
```diff
@@ -345,6 +346,14 @@ func (s *Span) End() {
 	if s.Type == "" {
 		s.Type = "custom"
 	}
+	if s.parent.IsExitSpan() {
+		if s.Type != s.parent.Type || s.Subtype != s.parent.Subtype {
+			s.tracer = nil
```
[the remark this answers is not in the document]
kruskall: Good point! I think we can just follow the normal control flow. I've updated the code

axw (approved):

PR #1334, opened 2022-10-13
Mark some more spans as exit spans

axw:
Also, note a bug in Span.End() when ending a child of an exit span.

Relates to https://github.com/elastic/apm-agent-go/issues/1315

### DIFF
--- a/span.go
+++ b/span.go
@@ -370,6 +370,9 @@ func (s *Span) End() {
 		s.Context.model.Destination = nil
 		s.Context.model.Service = nil
 
+		// BUG(axw) s.parent.SpanData must be accessed with s.parent.mu held.
+		// Even then, s.parent.SpanData will be nil if ended, so we need to
+		// find an alternative approach.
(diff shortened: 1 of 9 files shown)


span.go
```diff
@@ -370,6 +370,9 @@ func (s *Span) End() {
 		s.Context.model.Destination = nil
 		s.Context.model.Service = nil
 
+		// BUG(axw) s.parent.SpanData must be accessed with s.parent.mu held.
+		// Even then, s.parent.SpanData will be nil if ended, so we need to
+		// find an alternative approach.
```
simitt, line 375: Would you mind creating a gh issue for that bug? As I understand this won't be fixed within the same PR.

simitt (approved):

span.go
axw: I was intending to fix it in this PR. The other changes could slip to another release, but not this; this causes a panic if you end the parent before the child.

axw:
I haven't updated CHANGELOG, as the bug hasn't made it into a release.

simitt (approved):

Issue #1292, opened 2022-08-10
Handle span compression with new service_target_* fields


kruskall:
Handle span compression with new fields (stop relying on resource internally)

See spec for the description and details:

https://github.com/elastic/apm/blob/main/specs/agents/tracing-spans-service-target.md

PR #1339, opened 2022-10-23
feat: handle span compression with new service_target_* fields

kruskall:
Update compression logic to use service target fields and stop relying on resource internally.

The updated logic is compliant with the spec on span compression.
See https://github.com/elastic/apm/blob/main/specs/agents/handling-huge-traces/tracing-spans-compress.md#consecutive-same-kind-compression-strategy

Closes https://github.com/elastic/apm-agent-go/issues/1292


axw (commented):
I'm surprised that no tests have changed. Do we have any tests for "same kind" compression with target service?

simitt:
@kruskall have you had time to follow up with the question regarding the tests?

kruskall:
> I'm surprised that no tests have changed. Do we have any tests for "same kind" compression with target service?

Sorry for the late reply. I'm not sure what should change. We have tests to verify "same kind" compression is working as intended (https://github.com/elastic/apm-agent-go/blob/main/span_test.go#L536-L656) but they still pass since there is no "change" in behaviour.
Service target fields are fully supported and inferred when possible, same as the old destination service. Moving to the new fields would not change the "same kind" relantionship between spans created the same way.

axw:
It seems to me we have a (pre-existing) gap in our testing: AFAICS the "same kind" compression tests don't cover the code path where the span type and subtype are the same, but the target service name/type differ. I expected we would have a test for this, and that we should have changed the test to set service target rather than the deprecated destination fields.

We can merge this as is, but it would be good to improve the tests.

axw (approved):

kruskall:
> It seems to me we have a (pre-existing) gap in our testing: AFAICS the "same kind" compression tests don't cover the code path where the span type and subtype are the same, but the target service name/type differ. I expected we would have a test for this, and that we should have changed the test to set service target rather than the deprecated destination fields.
> 
> We can merge this as is, but it would be good to improve the tests.

Good point! I think we can add it here since it's related to the change and it's fairly small. I've added a test case, let me know what you think! :+1:

span_test.go
```diff
@@ -573,6 +573,32 @@ func TestCompressSpanSameKind(t *testing.T) {
 			currentTime = currentTime.Add(span.Duration)
 			span.End()
 		}
+		// Span is compressable and the next span is the same kind, but cannot be compressed
+		// since the service target fields are different (inferred by the db instance).
```
axw, line 577: Would we lose any coverage by explicitly setting the service target? I think it would make the test a bit easier to understand. We already test the DB -> service target inference elsewhere.

axw (approved):
LGTM, but I think we could simplify the test a little bit by not relying on inference.

span_test.go
kruskall: Good point! I don't think we'll lose anything. Updated :+1:

axw (approved):
Looks great, thank you :)

marclop:
Verified with 8.7.0 BC9 and the latest main (332875e4a3fb):


```go
package main

import (
	"fmt"
	"time"

	"go.elastic.co/apm/v2"
)

func main() {
	tracer := apm.DefaultTracer()
	tracer.SetSpanCompressionEnabled(true)
	tracer.SetExitSpanMinDuration(0)
	currentTime := time.Now()
	tx := tracer.StartTransactionOptions("8.7.0", "testplan", apm.TransactionOptions{
		Start: currentTime,
	})
	currentTime.Add(
		newSpan(tx, currentTime, "SELECT * FROM users", "mysql", "db", "myhost"),
	)
	currentTime.Add(
		newSpan(tx, currentTime, "SELECT * FROM users", "mysql", "db", "myhost"),
	)
	currentTime.Add(
		newSpan(tx, currentTime, "SELECT * FROM orders", "mysql", "db", "anotherhost"),
	)
	tx.End()
	tracer.Flush(nil)
	fmt.Printf("%+v\n", tracer.Stats())
}

func newSpan(tx *apm.Transaction, start time.Time, name, t, targetType, targetName string) time.Duration {
	span := tx.StartSpanOptions(name, t, apm.SpanOptions{
		ExitSpan: true, Start: start,
	})
	span.Duration = 2 * time.Microsecond
	span.Context.SetServiceTarget(apm.ServiceTargetSpanContext{
		Type: targetType,
		Name: targetName,
	})
	span.End()
	return span.Duration
}
```

image

PR #1344, opened 2022-10-28
feat(apmsql): Add Unwrap to return the wrapped driver.Conn

bongnv:
Since go 1.13, [`Conn.Raw`](https://pkg.go.dev/database/sql#Conn.Raw) was added to allow extracting the underlying connection. It becomes handy if it's used with libraries like ([pgx](https://github.com/jackc/pgx/blob/master/stdlib/sql.go)) which will allow us to run pgx-specific functionalities.

However, the current `apmsql` library adds a wrapper on top of that connection and it's a private type so there is no way to extract the raw connection from `pgx`. This PR will add `Unwrap` function which will allow getting the underlying connection.

### DIFF
--- a/module/apmsql/conn_test.go
+++ b/module/apmsql/conn_test.go
@@ -25,7 +25,7 @@ func TestConnUnwrap(t *testing.T) {
 	conn, err := db.Conn(ctx)
 	_ = conn.Raw(func(driverConn interface{}) error {
 		unwrappedConn := driverConn.(Unwrapper)
-		require.Equal(t, "*apmsql.connBeginTx", fmt.Sprintf("%T", unwrappedConn))
+		require.Equal(t, "*sqlite3.SQLiteConn", fmt.Sprintf("%T", unwrappedConn.Unwrap()))
 		return nil
 	})
 }
--- a/module/apmsql/conn_test.go
+++ b/module/apmsql/conn_test.go
@@ -1,0 +1,31 @@
+package apmsql_test
+
+import (
+	"context"
+	"database/sql/driver"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"go.elastic.co/apm/module/apmsql"
+	_ "go.elastic.co/apm/module/apmsql/sqlite3"
+)
+
+func TestConnUnwrap(t *testing.T) {
+	db, err := apmsql.Open("sqlite3", ":memory:")
+	require.NoError(t, err)
+	defer db.Close()
+
+	type Unwrapper interface {
+		Unwrap() driver.Conn
+	}
+
+	ctx := context.Background()
+	conn, err := db.Conn(ctx)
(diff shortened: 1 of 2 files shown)


module/apmsql/conn_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	type Unwrapper interface {
+		Unwrap() driver.Conn
+	}
+
+	ctx := context.Background()
+	conn, err := db.Conn(ctx)
+	_ = conn.Raw(func(driverConn interface{}) error {
+		unwrappedConn := driverConn.(Unwrapper)
+		require.Equal(t, "*apmsql.connBeginTx", fmt.Sprintf("%T", unwrappedConn))
```
axw, line 28: This is just type-asserting - should the unwrapper function be called, to get the underlying driver.Conn? And change the assertion below to check that the result is the sqlite3 Conn implementation?

axw (commented):
@bongnv thanks for the PR! Looks good, just one comment about the test.

Aside from that: we're not making enhancements to 1.x anymore, and all changes should go into main. Are you using v2 already? If not, is there anything in the agent preventing that?

module/apmsql/conn_test.go
bong-sc: Hi @axw , you're right. I've improved the assertion to verify against `*sqlite3.SQLiteConn`.

bongnv: Hi @axw , you're right. I've improved the assertion to verify against `*sqlite3.SQLiteConn`.

bongnv:
> 



> @bongnv thanks for the PR! Looks good, just one comment about the test.
> 
> Aside from that: we're not making enhancements to 1.x anymore, and all changes should go into main. Are you using v2 already? If not, is there anything in the agent preventing that?

Hi @axw , we've only noticed the existence of v2 recently and haven't had a plan to upgrade to v2. The major reason that we're not upgrading to v2 at the moment is that v2 and v1 aren't sharing the same context key in Go context. As the result, in order for apm to work properly, we must upgrade all references to apm at the same time which isn't an easy task to us.

If there is a way that we can use v2 and v1 together so we can gradually upgrade to v2, it will definitely make the upgrade easier.

axw:
> Hi @axw , we've only noticed the existence of v2 recently and haven't had a plan to upgrade to v2. The major reason that we're not upgrading to v2 at the moment is that v2 and v1 aren't sharing the same context key in Go context. As the result, in order for apm to work properly, we must upgrade all references to apm at the same time which isn't an easy task to us.

OK. Nevertheless, can you please retarget this PR to the main branch? Then we can look at how we can make the upgrade easier.

> If there is a way that we can use v2 and v1 together so we can gradually upgrade to v2, it will definitely make the upgrade easier.

One option may be to implement a module that provides most of the v1 API, and redirects it to the v2 API. Then you could use that as a replacement for the `go.elastic.co/apm` module.

There are some esoteric features from v1 that were not carried across, but for the basics this approach should work.

Issue #37, opened 2018-03-28
Feature: github.com/streadway/amqp tracing


axw:
Implement tracing for https://github.com/streadway/amqp.

This will probably have to be a collection of amqp.Channel method wrappers that:
 - add context.Context
 - create spans using said context
 - add headers (when we support distributed tracing)
 - call through to the underlying amqp.Channel

ekrucio:
Is this issue still open? This missing feature is a deal-breaker for our migration to Elastic APM.
If it's open and noone is working on it, I would like to implement it.
@axw @graphaelli

axw:
@ekrucio nobody else is working on it. I see you have opened a PR, thank you - I'll take a look shortly.

PR #1347, opened 2022-11-11
feat: Introduced new APM module to handle RabbitMQ messaging

ekrucio:
Messaging systems are one of the main components powering distributed microservices. We're tightly dependent on RabbitMQ messaging in our codebase, and especially this package [streadway/amqp](https://github.com/streadway/amqp) .

This PR is related to issue [Feature #37](https://github.com/elastic/apm-agent-go/issues/37) but does not resolve it, it is simply a POC. 
If you're okay with those changes I will add tests.


module/apmstreadwayamqp/utils.go
```diff
@@ -0,0 +1,65 @@
...
+var (
+	elasticTraceparentHeader = strings.ToLower(apmhttp.ElasticTraceparentHeader)
+	w3cTraceparentHeader     = strings.ToLower(apmhttp.W3CTraceparentHeader)
+	tracestateHeader         = strings.ToLower(apmhttp.TracestateHeader)
+	apmStreadwayAmqpPrefix   = "apmstreadwayamqp_transaction_"
+)
+
+func getMessageTraceparent(headers map[string]interface{}, header string) (apm.TraceContext, bool) {
+	headerValue, ok := headers[fmt.Sprintf("%s_%s", apmStreadwayAmqpPrefix, header)]
+	if ok {
```
tsetsoop, line 19: [minor]
```
if !ok {
  return apm.TraceContext{}, false
}
if headerValue == null {
  return apm.TraceContext{}, false
}
```

module/apmstreadwayamqp/extractor.go
```diff
@@ -0,0 +1,76 @@
...
+
+func еxtract(ctx context.Context, tracer *apm.Tracer, del amqp.Delivery) (*apm.Transaction, context.Context) {
+	if err := del.Headers.Validate(); err != nil {
+		return nil, ctx
+	}
+
+	traceContext, ok := getMessageTraceparent(del.Headers, w3cTraceparentHeader)
+	if !ok {
+		traceContext, ok = getMessageTraceparent(del.Headers, elasticTraceparentHeader)
+	}
```
tsetsoop, line 67: [minor] this could be since you check for both cases of ok

module/apmstreadwayamqp/extractor.go
```diff
@@ -0,0 +1,76 @@
...
+
+	if ok {
+		traceContext.State, _ = getMessageTracestate(del.Headers, tracestateHeader)
+	}
+
+	tx := tracer.StartTransactionOptions("rabbitmq.handle", "messaging", apm.TransactionOptions{TraceContext: traceContext})
+	ctx = apm.ContextWithTransaction(ctx, tx)
```
axw, line 74: This assumes a specific consumer pattern, but there are others too. Clients might consume a batch of messages and handle them in bulk (one transaction, many messages). Or a client might poll for messages, or synchronously fetch a message, within the context of another operation (messaging span, not transaction).

I think we should introduce a package-level function `func ExtractTraceContext(amqp.Delivery) (apm.TraceContext, error)`. That bit can be done independently of the consumer pattern.

Given streadway/amqp's fairly generic API, I think it might be best to leave the creation of transactions/spans to client (i.e. your application) code.

module/apmstreadwayamqp/utils.go
```diff
@@ -0,0 +1,64 @@
...
+	"go.elastic.co/apm/module/apmhttp/v2"
+	"go.elastic.co/apm/v2"
+	"strings"
+)
+
+var (
+	elasticTraceparentHeader = strings.ToLower(apmhttp.ElasticTraceparentHeader)
+	w3cTraceparentHeader     = strings.ToLower(apmhttp.W3CTraceparentHeader)
+	tracestateHeader         = strings.ToLower(apmhttp.TracestateHeader)
+	apmStreadwayAmqpPrefix   = "apmstreadwayamqp_transaction_"
```
axw, line 14: Why are you adding a prefix? This means that distributed tracing will not work when producers/consumers are written in different languages, use different clients, or use different tracers (e.g. OTel + Elastic APM). Can we just use the standard "traceparent" and "tracestate" names?

module/apmstreadwayamqp/utils.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"fmt"
+	"go.elastic.co/apm/module/apmhttp/v2"
+	"go.elastic.co/apm/v2"
+	"strings"
+)
+
+var (
+	elasticTraceparentHeader = strings.ToLower(apmhttp.ElasticTraceparentHeader)
```
axw, line 11:
```suggestion
```

I don't think we need that one for messaging.

module/apmstreadwayamqp/utils.go
```diff
@@ -0,0 +1,64 @@
...
+	w3cTraceparentHeader     = strings.ToLower(apmhttp.W3CTraceparentHeader)
+	tracestateHeader         = strings.ToLower(apmhttp.TracestateHeader)
+	apmStreadwayAmqpPrefix   = "apmstreadwayamqp_transaction_"
+)
+
+func getMessageTraceparent(headers map[string]interface{}, header string) (apm.TraceContext, bool) {
+	headerValue, ok := headers[fmt.Sprintf("%s_%s", apmStreadwayAmqpPrefix, header)]
+	if !ok || headerValue == nil {
+		return apm.TraceContext{}, false
+	}
```
axw, line 21:
```suggestion
	headerValue, ok := headers[fmt.Sprintf("%s_%s", apmStreadwayAmqpPrefix, header)].(string)
	if !ok {
		return apm.TraceContext{}, false
	}
```

Then you can remove the type switch below.

module/apmstreadwayamqp/utils.go
```diff
@@ -0,0 +1,64 @@
...
+			return trP, true
+		}
+	default:
+		return apm.TraceContext{}, false
+	}
+	return apm.TraceContext{}, false
+}
+
+func getMessageTracestate(headers map[string]interface{}, header string) (apm.TraceState, bool) {
+	headerValue, ok := headers[fmt.Sprintf("%s_%s", apmStreadwayAmqpPrefix, header)]
```
axw, line 34:
```suggestion
	headerValue, ok := headers[fmt.Sprintf("%s_%s", apmStreadwayAmqpPrefix, header)].(string)
```

Again, I think you can use a simpler type assertion here.

module/apmstreadwayamqp/pub.go
```diff
@@ -0,0 +1,72 @@
...
+func (c msgPublisher) WithContext(ctx context.Context) Publisher {
+	c.ctx = ctx
+	return c
+}
+
+func (c msgPublisher) Publish(exchange, key string, mandatory, immediate bool, msg amqp.Publishing) error {
+	return pub(c.ctx, c.ch, exchange, key, mandatory, immediate, msg)
+}
+
+func pub(ctx context.Context, ch *amqp.Channel, exchange, key string, mandatory, immediate bool, msg amqp.Publishing) error {
```
axw, line 45: Instead of creating an interface with a bound context, can we expose this function as a package-level function? i.e.

```go
// Publish calls ch.Publish.
//
// If ctx contains an apm.Transaction, Publish will create a span and inject trace context into msg.Headers.
//
// NOTE: ctx is not used for cancellation.
func Publish(ctx context.Context, ch *amqp.Channel, exchange, key string, mandatory, immediate bool, msg amqp.Publishing) error
```

module/apmstreadwayamqp/pub.go
```diff
@@ -0,0 +1,72 @@
...
+	span, ctx := apm.StartSpanOptions(ctx, "rabbitmq.pub", "messaging", apm.SpanOptions{
+		ExitSpan: true,
+	})
+	span.Subtype = "rabbitmq"
+	span.Outcome = "success"
+	defer span.End()
```
axw, line 51:
```suggestion
	span, ctx := apm.StartSpanOptions(ctx, "rabbitmq.pub", "messaging", apm.SpanOptions{
		ExitSpan: true,
	})
	span.Subtype = "rabbitmq"
	span.Outcome = "success"
	defer span.End()
```

Is it valid to assume that the broker is always RabbitMQ? I guess it's fine.

According to https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-messaging.md#-rabbitmq-naming-specifics, this span name should have the format `RabbitMQ SEND to <exchange>`. In case `exchange` is empty, it should be replaced with the string `"<default>"`.

module/apmstreadwayamqp/pub.go
```diff
@@ -0,0 +1,72 @@
...
+
+	setMessageTraceparent(span.TraceContext(), msg.Headers)
+	stateErr := setMessageTracestate(span.TraceContext().State, msg.Headers)
+	if stateErr != nil {
+		span.Outcome = "failure"
+		return stateErr
+	}
```
axw, line 58:
```suggestion
	setMessageTracestate(span.TraceContext().State, msg.Headers)
```

Setting tracestate should never fail.

module/apmstreadwayamqp/pub.go
```diff
@@ -0,0 +1,72 @@
...
+func (c msgPublisher) Publish(exchange, key string, mandatory, immediate bool, msg amqp.Publishing) error {
+	return pub(c.ctx, c.ch, exchange, key, mandatory, immediate, msg)
+}
+
+func pub(ctx context.Context, ch *amqp.Channel, exchange, key string, mandatory, immediate bool, msg amqp.Publishing) error {
+	span, ctx := apm.StartSpanOptions(ctx, "rabbitmq.pub", "messaging", apm.SpanOptions{
+		ExitSpan: true,
+	})
+	span.Subtype = "rabbitmq"
+	span.Outcome = "success"
```
axw, line 50:
```suggestion
```

Outcome defaults to "success" if there are no errors reported by the time it is ended.

module/apmstreadwayamqp/pub.go
```diff
@@ -0,0 +1,72 @@
...
+	pubErr := ch.Publish(
+		exchange,
+		key,
+		false, // mandatory
+		false, // immediate
+		msg,
+	)
+
+	if pubErr != nil {
+		span.Outcome = "failure"
```
axw, line 69: Should we report the error as an event to APM? i.e. `apm.CaptureError(ctx, pubErr).Send()`

If you do that, you don't need to explicitly set Outcome at all; it will be set to "failure" due to the error being reported.

module/apmstreadwayamqp/pub.go
```diff
@@ -0,0 +1,72 @@
...
+	if stateErr != nil {
+		span.Outcome = "failure"
+		return stateErr
+	}
+
+	pubErr := ch.Publish(
+		exchange,
+		key,
+		false, // mandatory
+		false, // immediate
```
axw, line 64:
```suggestion
		mandatory,
		immediate,
```

module/apmstreadwayamqp/utils.go
```diff
@@ -0,0 +1,64 @@
...
+
+func setMessageTracestate(tracestate apm.TraceState, headers map[string]interface{}) error {
+	if err := tracestate.Validate(); err != nil {
+		return err
+	}
+	if headers != nil {
+		headers[fmt.Sprintf("%s_%s", apmStreadwayAmqpPrefix, tracestateHeader)] = tracestate.String()
+	}
+	return nil
+}
```
axw, line 64:
```suggestion
func setMessageTracestate(tracestate apm.TraceState, headers map[string]interface{}) {
	if encoded := tracestate.String(); encoded != "" && headers != nil {
		headers[fmt.Sprintf("%s_%s", apmStreadwayAmqpPrefix, tracestateHeader)] = encoded
	}
}
```

Tracestate is validated when creating transactions, and will ignore it if it is invalid - in which case the empty tracestate will be stored, and String() will return "".

module/apmstreadwayamqp/pub.go
```diff
@@ -0,0 +1,72 @@
...
+func pub(ctx context.Context, ch *amqp.Channel, exchange, key string, mandatory, immediate bool, msg amqp.Publishing) error {
+	span, ctx := apm.StartSpanOptions(ctx, "rabbitmq.pub", "messaging", apm.SpanOptions{
+		ExitSpan: true,
+	})
+	span.Subtype = "rabbitmq"
+	span.Outcome = "success"
+	defer span.End()
+
+	setMessageTraceparent(span.TraceContext(), msg.Headers)
+	stateErr := setMessageTracestate(span.TraceContext().State, msg.Headers)
```
axw, line 54: If we introduce a package-level `ExtractTraceContext` function, perhaps we should also extract these into a `InjectTraceContext` function:

```go
func InjectTraceContext(apm.TraceContext, amqp.Delivery)
```

axw: Also, in some cases you should use the transaction context rather than the span context. Specifically, when the span is dropped due to the max spans per transaction having been reached. In this case the returned span won't have a trace context.

Here's how the apmhttp client instrumentation does it, for reference: https://github.com/elastic/apm-agent-go/blob/acf7d2b497fa70d5a1bb06088f700c152de5fb0c/module/apmhttp/client.go#L85-L120

axw (commented):
@ekrucio thanks for opening the PR!

The Extractor interface feels a bit too specific to the pattern you are using in your application. Because streadway/amqp provides a fairly general API, I think we may have to provide a toolkit of functions for creating transactions or spans for consumers, rather than always assuming a one-to-one mapping between messages and transactions (effectively this pattern, IIANM: https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-messaging.md#passive-message-reception)

Are you defining a `Publisher` interface in your code, so you can use either an `amqp.Channel` or an `amqpstreadway.Publisher`? If that is the case, I think the `Publisher` could stay - but it does not need to be an interface; it should instead be a struct type.

module/apmstreadwayamqp/extractor.go
ekrucio: I agree completely. 
If I understand correctly your idea is for us the extract and return the trace context, and the client should decide what to do with it. I like this more, it's more generic and gives more freedom to the client. 👍

module/apmstreadwayamqp/extractor.go
```diff
@@ -0,0 +1,76 @@
...
+	// Extract returns a transaction from the extracted
+	// trace information from the headers.
+	//
+	// The started transaction is stored in the ctx.
+	Extract(del amqp.Delivery) (*apm.Transaction, context.Context)
+}
+
+// Prepare returns an Extractor which could start apm.Transaction
+// related to previous transaction or span, or new one.
+func Prepare(headers map[string]interface{}) Extractor {
```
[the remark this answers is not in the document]
ekrucio: > What is the `headers` parameter? `headerExtractor.h` doesn't appear to be used anywhere.

`headers` are the amqp.Delivery headers, but I agree they're not used anywhere and are completely redundant, will remove them.

module/apmstreadwayamqp/pub.go
ekrucio: > Is it valid to assume that the broker is always RabbitMQ? I guess it's fine.
I think yes. 
Yes, I will update accordingly to follow the specifications.

module/apmstreadwayamqp/pub.go
ekrucio: I agree, this seems like the right thing to do, report it.

module/apmstreadwayamqp/extractor.go
axw: Yes you do understand me correctly :smile:

ekrucio:
> 



> @ekrucio thanks for opening the PR!
> 
> The Extractor interface feels a bit too specific to the pattern you are using in your application. Because streadway/amqp provides a fairly general API, I think we may have to provide a toolkit of functions for creating transactions or spans for consumers, rather than always assuming a one-to-one mapping between messages and transactions (effectively this pattern, IIANM: https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-messaging.md#passive-message-reception)
> 
> Are you defining a `Publisher` interface in your code, so you can use either an `amqp.Channel` or an `amqpstreadway.Publisher`? If that is the case, I think the `Publisher` could stay - but it does not need to be an interface; it should instead be a struct type.

My idea about the Publisher interface was to be some kind of a wrapper over the ch.Publish and thus the way to get access to this "custom publisher" we should call Wrap(ch *amqp.Channel).
// Wrap wraps amqp.Channel such that its Publish method calls streadway/amqp.Publish.
I do undertstand its somewhat constraining, but streadway/amqp is really extensive in what you can do with a streadway/amqp.Channel . Do we aim for this module to be a complete wrapper over them, or we should go "piece by piece" and the community could fill the needed, especially since the precedent/example would be available?

I completely agree with the unneeded and wrong "contract" initiated by the current implementation of the Extractor. Should the change to return trace context be suffiecient to fix this?

module/apmstreadwayamqp/utils.go
axw: I see, thanks for elaborating.

> to distribute a trace through a message

Users can still send trace events through messages - they would be in the message body. What we might break is having two separate instrumentation libraries (e.g. OTel + Elastic APM) enabled, which both want to set `traceparent`. Users will just have to choose one or the other.

I would prefer to always set the headers, and document this behaviour. Otherwise if an error were returned, then an application might start failing just because two instrumentation libraries have been used (perhaps inadvertently). While distributed tracing might be broken, I think that's preferable to breaking the user's application.

axw:
> My idea about the Publisher interface was to be some kind of a wrapper over the ch.Publish and thus the way to get access to this "custom publisher" we should call Wrap(ch *amqp.Channel).

:+1:

> I do undertstand its somewhat constraining, but streadway/amqp is really extensive in what you can do with a streadway/amqp.Channel . Do we aim for this module to be a complete wrapper over them, or we should go "piece by piece" and the community could fill the needed, especially since the precedent/example would be available?

I think starting out minimally is fine, and it can be filled in later.

For Publisher, I'm mostly just bothered about exporting an interface rather than a concrete struct type. What do you think about having a Channel wrapper struct type that embeds amqp.Channel, and overwrites its Publish method? Something like:

```go
// WrappedChannel wraps amqp.Channel such that Publish calls are traced,
// and trace context is added to published messages.
//
// Trace context must be supplied using Channel.WithContext.
type WrappedChannel struct {
    *amqp.Channel
    ctx context.Context
}

func WrapChannel(ch amqp.Channel) WrappedChannel {
    ...
}

func (c WrappedChannel) WithContext(ctx context.Context) WrappedChannel {
    return WrappedChannel{Channel: c.Channel, ctx: ctx}
}

// Publish publishes a message and returns an error in encountered.
func (c WrappedChannel) Publish(exchange, key string, mandatory, immediate bool, msg amqp.Publishing) error {
    ...
}
```

> I completely agree with the unneeded and wrong "contract" initiated by the current implementation of the Extractor. Should the change to return trace context be suffiecient to fix this?

:+1: Changing `Extractor` to some kind of `ExtractTraceContext` function would resolve that part.

module/apmstreadwayamqp/utils.go
ekrucio: > I would prefer to always set the headers, and document this behaviour. Otherwise if an error were returned, then an application might start failing just because two instrumentation libraries have been used (perhaps inadvertently). While distributed tracing might be broken, I think that's preferable to breaking the user's application. 

That's reasonable, I agree. 👍

ekrucio:
> For Publisher, I'm mostly just bothered about exporting an interface rather than a concrete struct type. What do you think about having a Channel wrapper struct type that embeds amqp.Channel, and overwrites its Publish method? Something like:
> 
> ```go
> // WrappedChannel wraps amqp.Channel such that Publish calls are traced,
> // and trace context is added to published messages.
> // ...
> ```

That is better, I agree, will do it that way 👍 .

ekrucio:
@axw Just fyi I made the necessary changes (or at least I think so).

module/apmstreadwayamqp/tx_ctx.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	if ok {
+		txCtx.State, _ = getMessageTracestate(del.Headers, tracestateHeader)
+	}
+	return txCtx, nil
+}
+
+func getMessageTraceparent(headers map[string]interface{}, header string) (apm.TraceContext, bool) {
+	headerValue, ok := headers[header].(string)
```
axw, line 54: Are header keys guaranteed to be received lowercase? If not, we might need to iterate over all keys and perform a case-insensitive comparison.

module/apmstreadwayamqp/tx_ctx.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	return apm.TraceContext{}, false
+}
+
+func getMessageTracestate(headers map[string]interface{}, header string) (apm.TraceState, bool) {
+	headerValue, ok := headers[header].(string)
+	if !ok {
+		return apm.TraceState{}, false
+	}
+	if trP, err := apmhttp.ParseTracestateHeader(strings.Split(headerValue, ",")...); err == nil {
```
axw, line 69: ParseTracestateHeader already splits on `,` internally, is it necessary to do it here?

module/apmstreadwayamqp/wrap.go
```diff
@@ -0,0 +1,58 @@
...
+
+func WrapChannel(ch *amqp.Channel) WrappedChannel {
+	return WrappedChannel{Channel: ch, ctx: context.Background()}
+}
+
+func (c WrappedChannel) WithContext(ctx context.Context) WrappedChannel {
+	return WrappedChannel{Channel: c.Channel, ctx: ctx}
+}
+
+// Publish publishes a message and returns an error in encountered.
```
axw, line 28:
```suggestion
// Publish publishes a message and returns an error if encountered.
//
// Publish will trace the operation as a span if the context associated with the channel
// (i.e. supplied with WithContext) contains an `*apm.Transaction.`. The trace context
// will be propagated as headers in the published message.
```

module/apmstreadwayamqp/wrap.go
```diff
@@ -0,0 +1,58 @@
...
+func (c WrappedChannel) Publish(exchange, key string, mandatory, immediate bool, msg amqp.Publishing) error {
+	ctx := c.ctx
+	var sn string
+	if len(exchange) == 0 {
+		sn = "default"
```
axw, line 33:
```suggestion
		sn = "<default>"
```

module/apmstreadwayamqp/wrap.go
```diff
@@ -0,0 +1,58 @@
...
+	} else {
+		sn = exchange
+	}
+	span, ctx := apm.StartSpanOptions(ctx, sn, "messaging", apm.SpanOptions{
+		ExitSpan: true,
+	})
+	span.Subtype = "rabbitmq"
+	defer span.End()
+
+	InjectTraceContext(span.TraceContext(), msg)
```
axw, line 43:
```suggestion
	tx := apm.TransactionFromContext(ctx)
	if tx == nil {
		return c.Channel.Publish(exchange, key, mandatory, immediate, msg)
	}
	
	traceContext := tx.TraceContext()
	if traceContext.Options.Recorded() {
		span, ctx := apm.StartSpanOptions(ctx, sn, "messaging", apm.SpanOptions{ExitSpan: true})
		if !span.Dropped() {
			traceContext = span.TraceContext()
			span.Subtype = "rabbitmq"
			defer span.End()
		} else {
			span.End()
		}
	}

	InjectTraceContext(span.TraceContext(), msg)
```

If there's no transaction in the context, the method call should just delegate directly to the wrapped Channel. If there's a transaction, we should use its TraceContext if the transaction is non-sampled, or if the span is dropped because the max span count is reached.

module/apmstreadwayamqp/tx_ctx.go
```diff
@@ -0,0 +1,73 @@
...
+// for trace propagation => "Traceparent"
+// If the provided msg contains header with this key
+// it will be overwritten
+func InjectTraceContext(tc apm.TraceContext, msg amqp.Publishing) {
+	if msg.Headers != nil {
+		msg.Headers[w3cTraceparentHeader] = apmhttp.FormatTraceparentHeader(tc)
+	}
+	if encoded := tc.State.String(); encoded != "" && msg.Headers != nil {
+		msg.Headers[tracestateHeader] = encoded
+	}
```
axw, line 30:
```suggestion
		msg.Headers[w3cTraceparentHeader] = apmhttp.FormatTraceparentHeader(tc)
		if encoded := tc.State.String(); encoded != "" {
			msg.Headers[tracestateHeader] = encoded
		}
	}
```

axw (changes requested):
@ekrucio thanks for the changes, this looks like it's in a good shape now. I've left a few more comments, otherwise it needs some tests, copyright headers, and doc comments.

module/apmstreadwayamqp/tx_ctx.go
ekrucio: I agree, case-insensitive comparison makes the contract more flexible 👍

ekrucio:
@axw Sorry, I didn't read you whole comment and requested a review before resolving everything mentioned here -> 
>  I've left a few more comments, otherwise it needs some tests, copyright headers, and doc comments.

Will add those 👍

module/apmstreadwayamqp/wrap_test.go
```diff
@@ -0,0 +1,50 @@
...
+	key := "key"
+	exch := "exchange"
+	msg := amqp.Publishing{
+		Headers: initialHeaders,
+	}
+
+	ch := amqp.Channel{}
+	wrCh := WrapChannel(&ch).WithContext(ctx)
+
+	defer func() {
```
axw, line 38: This is a bummer. We could perhaps use https://github.com/valinurovam/garagemq to run an in-memory AMQP server. I think what you have here is OK for now, but could you please also check the spans created?

Rather than using `apm.DefaultTracer` above, you should use `apmtest.WithTransaction` like in https://github.com/elastic/apm-agent-go/blob/b6f4f158b8f8d8d0c759deb4cd45811439ce70d4/module/apmhttp/client_test.go#L233. The function provided to WithTransaction is given a context that contains at transaction. WithTransaction returns any events produced during that function call.

axw (changes requested):

module/apmstreadwayamqp/tx_ctx.go
```diff
@@ -0,0 +1,101 @@
...
+	txCtx, ok := getMessageTraceparent(del.Headers, w3cTraceparentHeader)
+	if !ok {
+		txCtx, ok = getMessageTraceparent(del.Headers, elasticTraceparentHeader)
+	}
+
+	if ok {
+		txCtx.State, _ = getMessageTracestate(del.Headers, tracestateHeader)
+	}
+	return txCtx, nil
+}
```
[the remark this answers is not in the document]
ekrucio: This tottaly slipped past me, I agree 👍

ekrucio:
@axw I've added the proposed fixes. Also your idea about the in-memory AMQP server was top! I was ashamed of the way I did the test initially and it's much better that way. 👍

axw (commented):
Thanks @ekrucio, the main code all looks good, and the new test is nice :)

I have some concerns about the test being potentially flaky, for which I've left a suggested fix. 

Other than that, there's just a few other things that need to be updated:
- please run `make scripts/Dockerfile-testing update-modules update-licenses fmt` (you can verify the results by running `make precheck` -- this is required to pass by CI)
- please add a small entry to CHANGELOG.asciidoc, and to docs/instrumenting.asciidoc. See https://github.com/elastic/apm-agent-go/pull/1301/files for an example of where a new instrumentation module was added.

ekrucio:
>     * please run `make scripts/Dockerfile-testing update-modules update-licenses fmt` (you can verify the results by running `make precheck` -- this is required to pass by CI)

@axw 
Well this didn't go as planned. I'm not sure what Go version I should use, didn't find it mentioned anywhere and I think I have to use 1.15 to successfully complete those steps in the Makefile.
That's the problem, I'm using a MacBook M1 and there's no Go 1.15 for ARM processors.
[https://go.dev/dl/](https://go.dev/dl/)
I may be completely mistaken and I could just be doing something wrong. Can you assist me with those steps, or give me pointers if there's something else I'm missing.

CHANGELOG.asciidoc
```diff
@@ -23,6 +23,8 @@ endif::[]
 
 https://github.com/elastic/apm-agent-go/compare/v2.2.0...main[View commits]
 
+- Add `apmstreadwayamqp` module for tracing messages by streadway/amqp
```
axw, line 26:
```suggestion
- Add `apmstreadwayamqp` module for tracing messages by streadway/amqp {pull}1347[#1347]
```

module/apmhttp/go.mod
```diff
@@ -9,4 +9,4 @@ require (
 
 replace go.elastic.co/apm/v2 => ../..
 
-go 1.15
+go 1.17
```
axw, line 12: This should not be changed.

module/apmfasthttp/go.sum
```diff
@@ -36,6 +36,7 @@ github.com/prometheus/procfs v0.7.3 h1:4jVXhlkAyzOScmCkXBTOLRLTz8EeU+eyjrwB/EPq0
 github.com/prometheus/procfs v0.7.3/go.mod h1:cz+aTbrPOrUb4q7XlbU9ygM+/jj0fzG6c1xBZuNvfVA=
 github.com/santhosh-tekuri/jsonschema v1.2.4 h1:hNhW8e7t+H1vgY+1QeEQpveR6D4+OwKPXCfD2aieJis=
 github.com/santhosh-tekuri/jsonschema v1.2.4/go.mod h1:TEAUOeZSmIxTTuHatJzrvARHiuO9LYd+cIxzgEHCQI4=
+github.com/streadway/amqp v1.0.0/go.mod h1:AZpEONHx3DKn8O/DFsRAY58/XVQiIPMTMB1SddzLXVw=
```
axw, line 39: Huh, why is amqp in here... did something go awry with the make target I shared with you?

axw (commented):
> I may be completely mistaken and I could just be doing something wrong. Can you assist me with those steps, or give me pointers if there's something else I'm missing.

You can use a newer version of Go than 1.15, that's just the minimum for using the agent. I can take a look at fixing these things up and push to your branch.

Would you please add the docs/instrumenting.asciidoc section?

axw:
> You can use a newer version of Go than 1.15, that's just the minimum for using the agent. I can take a look at fixing these things up and push to your branch.

I can't push to your branch. If you're comfortable doing so, please enable that at https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

ekrucio:
@axw We had to raise the root Go version to Go 1.17 because one of the packages -> "golang.org/x/sys/unix" requires 1.17, otherwise it doesn't want to compile.
`# [golang.org/x/sys/unix](http://golang.org/x/sys/unix)
../../../pkg/mod/golang.org/x/sys@v0.2.0/unix/syscall.go:83:16: undefined: unsafe.Slice
../../../pkg/mod/golang.org/x/sys@v0.2.0/unix/syscall_darwin.go:95:8: undefined: unsafe.Slice
../../../pkg/mod/golang.org/x/sys@v0.2.0/unix/syscall_unix.go:118:7: undefined: unsafe.Slice
../../../pkg/mod/golang.org/x/sys@v0.2.0/unix/sysvshm_unix.go:33:7: undefined: unsafe.Slice
note: module requires Go 1.17`

axw:
> @axw We had to raise the root Go version to Go 1.17 because one of the packages -> "golang.org/x/sys/unix" requires 1.17, otherwise it doesn't want to compile.

Raising the minimum version for all modules is a showstopper. We support Go 1.15+, and changing that will require a major version bump.

Tests have been passing across the repo with Go 1.15. I don't see any changes to `golang.org/x/sys/unix` in the PR. Maybe you saw this error while updating module dependencies? Can you please remove the `go 1.17` change, and then let me know which specific command fails?

Also, while running the tests, I noticed some race conditions and other errors originating from garagemq and its dependencies. If you run `go test -race` in apmstreadwayamqp, you should see the same. I think it may be best to revert to your original approach for now.

Jamess-Lucass:
Sorry if this is a moot point but in the issue it does mention support for [streadway/amqp](https://github.com/streadway/amqp). However this is now a unmaintained fork and in the [README](https://github.com/streadway/amqp#beware-of-abandonware) states.

"This repository is NOT ACTIVELY MAINTAINED. Consider using a different fork instead: [rabbitmq/amqp091-go](https://github.com/rabbitmq/amqp091-go)."

I am not sure if this changes any implementation within this PR but would it be best to move forward with supporting  [rabbitmq/amqp091-go](https://github.com/rabbitmq/amqp091-go) instead.

I am quite busy with work at the moment but if this pull request is stale I may have a go at implementing a modulefor rabbitMQ over the coming months.

ekrucio:
> Sorry if this is a moot point but in the issue it does mention support for [streadway/amqp](https://github.com/streadway/amqp). However this is now a unmaintained fork and in the [README](https://github.com/streadway/amqp#beware-of-abandonware) states.
> 
> "This repository is NOT ACTIVELY MAINTAINED. Consider using a different fork instead: [rabbitmq/amqp091-go](https://github.com/rabbitmq/amqp091-go)."
> 
> I am not sure if this changes any implementation within this PR but would it be best to move forward with supporting [rabbitmq/amqp091-go](https://github.com/rabbitmq/amqp091-go) instead.
> 
> I am quite busy with work at the moment but if this pull request is stale I may have a go at implementing a module for RabbitMQ over the coming months.

I'm sorry, I've been pretty busy lately and the last I remember is I think the PR is ready to be merged with the exception of changes related to the Go modules which I didn't had to look at.
About the status of the amqp module which is instrumented, I'm not aware of the compatability of them both. My contribution is related to this RabbitMQ library only. 
@axw Hi and sorry for the 5 months gap. Since this repository stems from our organization fork, and I would not be able to grant you access, Are you okay with closing this PR and reopening it from my personal account and you could push the needed changes, as you proposed earlier.

axw:
> @axw Hi and sorry for the 5 months gap. Since this repository stems from our organization fork, and I would not be able to grant you access, Are you okay with closing this PR and reopening it from my personal account and you could push the needed changes, as you proposed earlier.

Sure, that's fine - go ahead and close/reopen when you're ready. It may take a while before I get to it though.

PR #1349, opened 2022-11-14
action: run matrix support

v1v:
### What

First stage to transition the CI from Jenkins to GitHub actions.

### Follow ups

- ~~Edge version is not supported yet while using the actions/setup-go -> https://github.com/actions/setup-go/issues/232~~
- ~~Report coverage.~~
- ~~Report test results.~~
- ~~Support Macos, see https://github.com/elastic/apm-agent-go/issues/1350~~
- Baremetals 
- Releases


axw (commented):
@v1v I would be quite happy to not convert test/coverage output to JUnit or Cobertura. We only did that because that was what we could consume in Jenkins.

I think we can just stick with plain test and coverage output, at least initially.

I'd like to reduce a bit more:
 - github.mod.* and setenv.sh: not needed if we don't need to convert
 - build.sh and test.sh: not needed, replace with `make check` on Linux and macOS
 - docker-test.sh: replace with an inline script:

```sh
./scripts/docker-compose-testing run -T --rm trace-context-harness
./scripts/docker-compose-testing up -d --build
./scripts/docker-compose-testing run -T --rm go-agent-tests make coverage GOFLAGS=-v
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,83 @@
...
+    paths-ignore:
+      - '**.md'
+      - '**.asciidoc'
+
+# limit the access of the generated GITHUB_TOKEN
+permissions:
+  contents: read
+
+env:
+  DEFAULT_GO_VERSION: '1.15.10'
```
axw, line 21: Instead of using this pattern, I think it might be better to use go-version-file: go.mod. That will cause the use of the minimum Go minor version supported by the project, as defined in go.mod. i.e. 1.15.

WDYT? (I realise you're carrying across the default we had in Jenkinsfile. May as well improve things while we're at it :))

axw (commented):
There have been a bunch more release since 1.15.10 -- there's 1.15.15. Perhaps updating will fix the segfaults we're seeing; there were some compiler and runtime fixes between those versions that look relevant. Not sure why it would affect GitHub Actions and not the Jenkins environment though.

I'm interested to know how we'll do benchmarks. Will we have a local runner on a baremetal machine?

.github/workflows/ci.yml
v1v: > WDYT? (I realise you're carrying across the default we had in Jenkinsfile. May as well improve things while we're at it :))

Totally in

v1v:
> I'm interested to know how we'll do benchmarks. Will we have a local runner on a baremetal machine?

Still something to be discussed, we are just learning how GH actions could be used for the daily tasks, so there are some constrains we need to find out. I'll keep this in draft for now

.github/workflows/ci.yml
v1v:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,83 @@
...
+        run: make check
+
+  coverage:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.DEFAULT_GO_VERSION }}
```
v1v, line 49:
```suggestion
          go-version-file: 'go.mod'
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,83 @@
...
+          ./scripts/docker-compose-testing run -T --rm go-agent-tests make coverage GOFLAGS=-v
+
+  test-windows:
+    runs-on: windows-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.DEFAULT_GO_VERSION }}
```
v1v, line 65:
```suggestion
          go-version-file: 'go.mod'
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,83 @@
...
+        run: go test -v ./...
+
+  test-macos:
+    runs-on: macos-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.DEFAULT_GO_VERSION }}
```
v1v, line 78:
```suggestion
          go-version-file: 'go.mod'
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,81 @@
...
+      - uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
+          check-latest: true
+          cache: true
+      - name: Build
+        run: go test -v ./...
+
+  test-macos:
+    runs-on: macos-latest
```
v1v, line 70: Let's try to reproduce the error using a different version

```suggestion
    runs-on: macos-12
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,81 @@
...
+        run: go test -v ./...
+
+  test-macos:
+    runs-on: macos-12
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
```
v1v, line 76:
```suggestion
          go-version: '1.16
          //go-version-file: 'go.mod'
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,82 @@
...
+        run: go test -v ./...
+
+  test-macos:
+    runs-on: macos-12
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '1.16
```
v1v, line 76:
```suggestion
          go-version: '1.16'
```

axw:
I made a few changes to the workflow in my fork, and got a green build: https://github.com/axw/apm-agent-go/actions/runs/3589944017/attempts/1

Changes:
 - Use `make precheck` in the "check" (now "precheck") step. This just does linting etc., and skips running the tests.
 - Use `make test` for the "test-*" steps, except for "test-windows"
 - Use `go-version-file: go.mod` for "precheck" and "coverage"
 - Use `go-version: '1.19'` for "test-macos" (seems there's some bugs with the older version of Go?)
 - Add `cache-dependency-path: '**/go.sum'` to all "setup-go" steps
 - Remove `check-latest: true` from all "steup-go" steps

axw (approved):

PR #1407, opened 2023-03-31
OpenTelemetry metrics exporter

dmathieu:
This PR is setting up an OpenTelemetry metrics exporter, based on the specification in https://github.com/elastic/apm/pull/742.


dmathieu:
Supporting Go versions down to 1.15 with OpenTelemetry is going to be very tricky.

Otel follows the [Golang release policy](https://go.dev/doc/devel/release#policy):
> Each major Go release is supported until there are two newer major releases.

Which means it only support 1.19 and 1.20, and has therefore started using generics, making it impossible to use this version with any Go < 1.18.
Going to a lower version of the SDK will not be possible either, as the metrics API wasn't finalized back then.

module/apmotel/gatherer.go
```diff
@@ -0,0 +1,116 @@
...
+// Gatherer is a gatherer/exporter which can act as an OpenTelemetry manual
+// reader, to retrieve metrics and expose them to the agent.
+type Gatherer struct {
+	metric.Reader
+}
+
+var _ metric.Reader = Gatherer{}
+var _ apm.MetricsGatherer = Gatherer{}
+
+// GatherMEtrics gathers metrics into out.
```
dubonzi, line 48: Small typo here.

axw (commented):
Very nice! Just a couple of comments.

module/apmotel/gatherer.go
```diff
@@ -0,0 +1,152 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
+
+	"go.elastic.co/apm"
```
axw, line 31:
```suggestion
	"go.elastic.co/apm/v2"
```

module/apmotel/go.mod
```diff
@@ -0,0 +1,40 @@
+module go.elastic.co/apm/module/apmotel
```
axw, line 1:
```suggestion
module go.elastic.co/apm/module/apmotel/v2
```

We version all instrumentation modules the same as the main module, which is currently on v2.

axw (commented):
@dmathieu great work, thank you. We just need to update the module major version suffixes, otherwise this is good to merge.

We can update the code later based on the outcome of the spec discussion about default bounds.

module/apmotel/config.go
```diff
@@ -0,0 +1,87 @@
...
+	case metric.InstrumentKindObservableGauge:
+		return aggregation.LastValue{}
+	case metric.InstrumentKindHistogram:
+		return aggregation.ExplicitBucketHistogram{
+			Boundaries: customHistogramBoundaries,
+			NoMinMax:   false,
+		}
+	}
+
+	log.Panicf("unknown instrument kind %q", ik)
```
marclop, line 85: I'm worried about this since instrumented applications will crash from the instrumentation code. I don't think that's what we want, right? Could we maybe just log, print or ignore altogether? I'm unsure what the current recommendation would be.

dmathieu: OpenTelemetry itself used a `panic` if that happens. I changed it to a `log.Panicf` so we could mention which instrument kind is at fault.

https://github.com/open-telemetry/opentelemetry-go/blob/65ebe5e50f1c4d5aeb01d9693551bc4aab888149/sdk/metric/reader.go#L147-L159

dmathieu: If a new instrument kind is somehow introduced, customers could get logs, but those wouldn't be actionable to them.
Introducing new instrument kinds would also break a lot of instrumentations, so I think that would probably cause a major release of otel anyway.

axw: Would it make sense to return [`aggregation.Default`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric/aggregation#Default) in this case? If there's no sensible option and opentelemetry-go will panic anyway, I guess it's OK, but I'd also prefer not to panic if we can avoid it.

module/apmotel/gatherer.go
axw: This still needs updating

module/apmotel/config.go
dmathieu: There's no way to test it, since we can't create an instrumentation kind which doesn't exist. But If I use it for sum and count, I'm getting errors:

```
could not create all aggregators: creating aggregator with instrumentKind: 2, aggregation {}: unrecognized aggregation: {}
```

```
func customAggregationSelector(ik metric.InstrumentKind) aggregation.Aggregation {
	switch ik {
	case metric.InstrumentKindHistogram:
		return aggregation.ExplicitBucketHistogram{
			Boundaries: customHistogramBoundaries,
			NoMinMax:   false,
		}
	default:
		return aggregation.Default{}
	}
}
```

Which may be an otel bug. The error is raised here, since `aggregation.Default{}` doesn't match anything in the case.
https://github.com/open-telemetry/opentelemetry-go/blob/65ebe5e50f1c4d5aeb01d9693551bc4aab888149/sdk/metric/pipeline.go#L391

dmathieu: I have opened a PR on otel to fix this issue: https://github.com/open-telemetry/opentelemetry-go/pull/3967
And changed this method to rely otel's default behavior (which panics as well)

axw (approved):
LGTM!

PR #1408, opened 2023-04-03
ci: remove notification and leftovers Jenkins

v1v:
## What does this PR do?

Remove GitHub build message in Jenkins. CI has been migrated to GH actions, but the micro-benchmarking 
that will happen eventually.

Move filter `when` to happen earlier in the pipeline

GH build message might be re-done to support GitHub actions, but for now the GH command is not accurate. Hence it's removed.


reakaleek (approved):

axw:
@v1v done

PR #1410, opened 2023-04-06
OpenTelemetry tracing bridge

dmathieu:
This setups a bridge (our own SDK actually) which allows us to use the OpenTelemetry APIs alongside our agent.


dmathieu:
Some areas are unfortunately impossible to test (setting tags on spans/transactions for example), because the data isn't readable within the test.

dmathieu:
I may have missed some things, but it's hard to say without real world usage. So I am marking this PR as ready for review.

marclop (commented):
just a quick review, haven't dived into all the details

module/apmotel/span.go
```diff
@@ -0,0 +1,323 @@
...
+	c := trace.NewEventConfig(opts...)
+	e := event{Name: name, Attributes: c.Attributes(), Time: c.Timestamp()}
+
+	s.mu.Lock()
+	s.events = append(s.events, e)
+	s.mu.Unlock()
+}
+
+func (s *span) IsRecording() bool {
+	return s.provider.apmTracer.Recording()
```
dmathieu, line 99: ping @marclop how do we get information whether a span is sampled or not on our agent?

module/apmotel/span.go
```diff
@@ -0,0 +1,323 @@
...
+	status := status{Code: code}
+	if code == codes.Error {
+		status.Description = description
+	}
+
+	s.status = status
+}
+
+func (s *span) SetName(name string) {
+	if s.span != nil {
```
dmathieu, line 138: Should lock a mutex in here.

module/apmotel/tracer.go
```diff
@@ -0,0 +1,112 @@
...
+					p = &span{tx: tx}
+				}
+			}
+		}
+
+		if p != nil {
+			if apmSpan, ok := p.(*span); ok {
+				// This is a child span. Create a span, not a transaction
+				spanOpts := apm.SpanOptions{
+					Parent: apmSpan.span.TraceContext(),
```
marclop, line 71: the parent may be a span or a tx, so the appropriate TraceContext should be used.

module/apmotel/tracer.go
```diff
@@ -0,0 +1,112 @@
...
+	s := &span{
+		provider: t.provider,
+
+		attributes:  config.Attributes(),
+		startTime:   startTime,
+		spanContext: psc,
+
+		tx: apmTx,
+	}
+	return trace.ContextWithSpan(ctx, s), s
```
marclop, line 111: I was having a bit of a hard time reasoning about the logic. I've made some changes and fixed a bug where the traceContext may not be passed down correctly. Let me know what you think and if I understood the flow correctly:

```suggestion
config := trace.NewSpanStartConfig(opts...)
	startTime := config.Timestamp()
	if startTime.IsZero() {
		startTime = time.Now()
	}

	s := &span{
		provider:   t.provider,
		attributes: config.Attributes(),
		startTime:  startTime,
	}
	var psc trace.SpanContext
	if config.NewRoot() {
		ctx = trace.ContextWithSpanContext(ctx, psc)
	} else {
		// If not root, check if a *span is already present, if nit's not,
		// attempt to obtain an APM transaction from the context.
		psc = trace.SpanContextFromContext(ctx)
		s.spanContext = psc
		var parent *span
		var ok bool
		if parent, ok = trace.SpanFromContext(ctx).(*span); !ok {
			// Try to find a obtain an APM transaction from the agent context.
			if tx := apm.TransactionFromContext(ctx); tx != nil {
				parent = &span{tx: tx}
			}
		}
		// Use the parent if it exists. Otherwise, we'll create a new
		// transaction using the trace context from `psc`.
		if parent != nil {
			// This is a child span. Create a span, not a transaction.
			// The parent may be a span or a transaction.
			var tc apm.TraceContext
			if parent.span != nil {
				tc = parent.span.TraceContext()
			} else {
				tc = parent.tx.TraceContext()
			}
			s.span = parent.tx.StartSpanOptions(spanName, "", apm.SpanOptions{
				Parent: tc,
				Start:  startTime,
			})
			s.tx = parent.tx
			return trace.ContextWithSpan(ctx, s), s
		}
	}

	var tranOpts apm.TransactionOptions
	if psc.HasTraceID() && psc.HasSpanID() {
		tranOpts.TraceContext = apm.TraceContext{
			Trace: [16]byte(psc.TraceID()),
			Span:  [8]byte(psc.SpanID()),
		}
	}
	s.tx = t.provider.apmTracer.StartTransactionOptions(spanName, "", tranOpts)
	return trace.ContextWithSpan(ctx, s), s
```

module/apmotel/span.go
```diff
@@ -0,0 +1,330 @@
...
+	s.events = append(s.events, e)
+	s.mu.Unlock()
+}
+
+func (s *span) IsRecording() bool {
+	if s.span != nil {
+		return !s.span.Dropped()
+	}
+
+	return true
```
marclop, line 103: If the span is nil, we should check if the tx is sampled (recording)
```suggestion
	return s.tx.Sampled()
```

module/apmotel/span.go
```diff
@@ -0,0 +1,330 @@
...
+	attributes  []attribute.KeyValue
+	events      []event
+	spanContext trace.SpanContext
+	status      status
+
+	tx   *apm.Transaction
+	span *apm.Span
+}
+
+func (s *span) End(options ...trace.SpanEndOption) {
```
marclop, line 67: Should end hold a write lock? it may be called concurrently twice and may attempt to set the duration and other span / tx fields.

Also, if end is called concurrently with `SetAttributes` (which holds a lock), It may cause race conditions when ranging over the stored attributes and attempting to set them at the same time.

module/apmotel/span_test.go
marclop, line 1: We should add a test for "grand child" spans:

```go
		{
			name: "with a grand child span",
			getSpan: func(ctx context.Context, tracer trace.Tracer) trace.Span {
				ctx, _ = tracer.Start(ctx, "parentSpan")
				ctx, _ = tracer.Start(ctx, "childSpan")
				ctx, s := tracer.Start(ctx, "grandChildSpan")
				return s
			},
			expectedSpans: []model.Span{
				{
					Name:    "grandChildSpan",
					Type:    "custom",
					Outcome: "success",
				},
			},
		},
```

module/apmotel/tracer.go
```diff
@@ -0,0 +1,112 @@
...
+
+	if !config.NewRoot() {
+		p := trace.SpanFromContext(ctx)
+		if _, ok := p.(*span); !ok {
+			// Try to find a transaction from the agent context
+			if _, ok := p.(*span); !ok {
+				if tx := apm.TransactionFromContext(ctx); tx != nil {
+					p = &span{tx: tx}
+				}
+			}
```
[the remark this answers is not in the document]
dmathieu: 🤦

module/apmotel/tracer.go
dmathieu: Pulling the diff locally looks good to me, it's indeed way more readable.

module/apmotel/span.go
dmathieu: Ah, I missed that method. Thanks.

module/apmotel/tracer.go
dmathieu: Thanks. I applied this with your other comment.

module/apmotel/span.go
```diff
@@ -0,0 +1,323 @@
...
+			haveHTTPContext = true
+			httpURL = v.Value.Emit()
+		case "http.method":
+			haveHTTPContext = true
+			httpMethod = v.Value.Emit()
+		case "http.host":
+			haveHTTPContext = true
+			haveHTTPHostTag = true
+			httpHost = v.Value.Emit()
+
```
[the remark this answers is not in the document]
dmathieu: It seems trickier than that. The logic between the one here (from apmot) and in that branch are quite different.
Some attributes are added, and others are removed.

I wonder if we need a spec to have a consistent behavior, as every language implementation is going to have this problem.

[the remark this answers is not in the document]
marclop: I wonder too. @felixbarny would you happen to know?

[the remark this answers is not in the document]
felixbarny: We don't have to map all attributes but we have to map some. We have a cross-agent spec for that: https://github.com/elastic/apm/blob/main/specs/agents/tracing-api-otel.md#attributes-mapping

marclop (commented):
Looks mostly good, just a few questions. I haven't looked at the tests in detail yet.

module/apmotel/span.go
```diff
@@ -0,0 +1,385 @@
...
+	s.span.Context.SetOTelAttributes(agentAttrs)
+	s.span.Context.SetOTelSpanKind(s.spanKind.String())
+}
+
+func (s *span) setTransactionAttributes() {
+	var (
+		isHTTP      bool
+		isRPC       bool
+		isMessaging = true
```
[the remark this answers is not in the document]
dmathieu: Ah, yes. Nice catch.

module/apmotel/span.go
```diff
@@ -0,0 +1,385 @@
...
+			// handles the case where the url.Host hasn't been set.
+			// Tries obtaining the host value from the "http.host" tag.
+			// If not found, or if the url.Host has a value, it won't
+			// mutate the existing host.
+			if url.Host == "" && haveHTTPHostTag {
+				url.Host = httpHost
+			}
+			s.span.Context.SetHTTPRequest(&http.Request{
+				ProtoMinor: 1,
+				ProtoMajor: 1,
```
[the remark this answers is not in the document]
dmathieu: We definitely can parse it, but the agent doesn't seem to do anything with them. So I've removed them from here.

module/apmotel/span.go
```diff
@@ -0,0 +1,385 @@
...
+		case "db.user":
+			dbContext.User = v.Value.Emit()
+			haveDBContext = true
+		case "db.name":
+			serviceTargetName = v.Value.Emit()
+
+		case "messaging.system":
+			s.span.Type = "messaging"
+			s.span.Subtype = v.Value.Emit()
+			serviceTargetType = v.Value.Emit()
```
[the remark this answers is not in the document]
dmathieu: Yeah. The spec is weird about that. `netName` is never used with messaging. And with an `if/else` case as they have the net case would never run if the messaging one does.
This section therefore seems like dead code. I've written a note to follow up on figuring out if the spec should be fixed, but didn't get to it yet.

module/apmotel/span.go
```diff
@@ -0,0 +1,385 @@
...
+
+	if serviceTargetType != "" || serviceTargetName != "" {
+		s.span.Context.SetServiceTarget(apm.ServiceTargetSpanContext{
+			Type: serviceTargetType,
+			Name: serviceTargetName,
+		})
+	}
+
+	s.span.Context.SetOTelAttributes(agentAttrs)
+	s.span.Context.SetOTelSpanKind(s.spanKind.String())
```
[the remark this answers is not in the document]
marclop: yep, thanks! The APM Server will map the `span.kind` to the right type based on the spec'd algorithm.

module/apmotel/span_test.go
```diff
@@ -0,0 +1,701 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
+
+	"go.elastic.co/apm/v2"
+	"go.elastic.co/apm/v2/model"
+	"go.elastic.co/apm/v2/transport/transporttest"
+)
+
+func TestSpanEnd(t *testing.T) {
```
marclop, line 41: We could add some `messaging.system` and `messaging.destination` tests here:

marclop (approved):
Looks good! Thanks for bearing with me with all the comments. It was a pretty big PR to review.

module/apmotel/span_test.go
dmathieu: Something more than what we have here? https://github.com/elastic/apm-agent-go/blob/73559701888014ae966538f1610ef614d09fe8a6/module/apmotel/span_test.go#L346-L371

module/apmotel/span.go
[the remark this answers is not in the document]
dmathieu: After talking with @SylvainJuge about it, I have opened a spec fix: https://github.com/elastic/apm/pull/787

module/apmotel/span_test.go
marclop: ah I missed that, thanks

PR #1420, opened 2023-04-27
Rely on the latest stable otel, not rc2

dmathieu:
I first hoped we could rely on the RC2 for the stable metrics API of otel, which provides more stability.
But the otelhttp instrumentation still relies on 0.37.0. So we need to use that version.


simitt (approved):

scripts/genmod/main.go
```diff
@@ -143,8 +143,6 @@ func updateModule(dir string, gomod *GoMod, modules map[string]*GoMod) error {
 			args = append(args, "-go", *goVersionFlag)
 		}
 		cmd := exec.Command("go", args...)
-		cmd.Env = append(os.Environ(), "GO111MODULE=on")
-		cmd.Env = append(cmd.Env, "GOPROXY=http://proxy.invalid", "GOSUMDB=sum.golang.org https://sum.golang.org")
```
dmathieu, line 147: I had to remove this, seemingly because of module version changes, which causes a download (before, the modules were in cache).
I'm not sure the reasoning for GOPROXY, and will merge to move forward. But I'm happy to rethink this also.

PR #1424, opened 2023-05-03
Otel Module: handle resources

dmathieu:
As I was digging through traces generated by this module, I would have wanted to see attributes set by it to identify where the traces were coming from. But I only was able to know the agent was sending them.

So this adds support for resources, as supported in the official OpenTelemetry SDK, which allows setting custom attributes on all spans.
By default, the following attributes are set:

* `service.name`: `unknown_service:<executable name>`
* `telemetry.sdk.language`: `go`
* `telemetry.sdk.name`: `apmotel`
* `telemetry.sdk.version`: `apm.AgentVersion`

This closely matches the default resource from the official SDK: https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/resource/resource.go#L203

### DIFF
--- a/module/apmotel/span.go
+++ b/module/apmotel/span.go
@@ -90,6 +90,10 @@ func (s *span) End(options ...trace.SpanEndOption) {
 		outcome = "unknown"
 	}
 
+	for iter := s.provider.resource.Iter(); iter.Next(); {
+		s.attributes = append(s.attributes, iter.Attribute())
+	}
+
 	if s.span != nil {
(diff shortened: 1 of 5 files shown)


kruskall:
For reviewers: https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/#telemetry-sdk

kruskall (approved):
LGTM!

marclop (approved):

PR #1437, opened 2023-05-12
apmotel: follow APM OTel spec and prefer delta temporality

### LINKED ISSUE #1436: module/apmotel: consider defaulting to delta temporality for metrics
Aggregating metrics in Elasticsearch tends to be easier when recorded as deltas. This might change if/when we migrate to TSDS, so this needs some investigation and discussion, but I think it may be a better default for now.

carsonip:
Make apmotel follow apm otel spec: https://github.com/elastic/apm/blob/main/specs/agents/metrics-otel.md#aggregation-temporality
- Prefer delta temporality on certain instruments
- Filter out zero values on delta temporality

Closes #1436


module/apmotel/gatherer_config.go
```diff
@@ -78,3 +86,17 @@ func customAggregationSelector(ik metric.InstrumentKind) aggregation.Aggregation
...
+// used.
+func WithTemporalitySelector(temporality metric.TemporalitySelector) GathererOption {
+	return func(cfg gathererConfig) gathererConfig {
+		cfg.temporality = temporality
+		return cfg
+	}
+}
+
+func defaultTemporalitySelector(_ metric.InstrumentKind) metricdata.Temporality {
+	return metricdata.CumulativeTemporality
```
axw, line 101: DeltaTemporality?

carsonip: :facepalm: was testing other stuff locally. Fixed now

carsonip:
> > For all instrument types with delta temporality, agents MUST filter out zero values before exporting. E.g. if a counter does not change since the last export, it must not be exported.
> 
> If you have time, could you take a stab at that too? Otherwise could you please open an issue?

Taken a stab at it, see if it makes sense to you.

module/apmotel/gatherer_config.go
```diff
@@ -78,3 +86,23 @@ func customAggregationSelector(ik metric.InstrumentKind) aggregation.Aggregation
...
+}
+
+func defaultTemporalitySelector(ik metric.InstrumentKind) metricdata.Temporality {
+	switch ik {
+	case metric.InstrumentKindCounter, metric.InstrumentKindObservableCounter, metric.InstrumentKindHistogram:
+		return metricdata.DeltaTemporality
+	case metric.InstrumentKindUpDownCounter, metric.InstrumentKindObservableUpDownCounter, metric.InstrumentKindObservableGauge:
+		return metricdata.CumulativeTemporality
+	}
+	panic("unknown instrument kind")
```
axw, line 107:
```suggestion
	}
	return metric.DefaultTemporalitySelector(ik)
```

I'd rather we didn't panic here, in case OTel grows new instrument kinds and we don't keep up to date.

axw (approved):
LGTM with the panic removed. Thanks!

axw (approved):
LGTM with the panic removed. Thanks!

dmathieu (approved):

PR #1453, opened 2023-06-06
Require Go 1.19+

### LINKED ISSUE #1447: Adopt opentelemetry-go's Go language compatibility policy
We have been considering removal of support for older versions of Go a breaking change, requiring a major version bump. It wasn't until we released v2 that we dropped support for Go 1.8, and when we introduced v2 we committed to supporting Go 1.15.

This is only for the core agent, though, and each instrumentation module may require more recent versions of Go, due to the external code with which we're integrating. For example (just to pick one), the recently introduced `module/apmotel` requires Go 1.20 since opentelemetry-go requires features introduced in recent Go versions.

It is very common in the Go ecosystem to align with the [Go release policy](https://go.dev/doc/devel/release#policy), and only support the two most recent versions of Go. Moreover, dropping older versions of Go is typically not considered a breaking change. If we continue to buck the trend, we make it more difficult to keep our instrumentation modules up to date. On the other hand if we were to make more frequent major releases, we would transfer burden onto users by requiring them to regularly update their import paths, as bumping the major version requires bumping the import path suffix.

Starting with v3, we will change our Go compatibility policy to align with the [opentelemtry-go policy](https://github.com/open-telemetry/opentelemetry-go#compatibility), where we will add support for a new Go version in one minor release, and drop testing and official support for older versions of Go in the following minor. We do this in v3 rather than v2 to give our users some time to adapt; once v3 is released, v2 will be supported until at least 2024-09-17 (see the [EOL policy](https://www.elastic.co/support/eol)).

axw:
Update docs to drop the "best effort" support for unsupported Go versions, and only test with the two most recent Go releases. Also, simplify the "check-modules" and "update-modules" make targets.

Closes #1447


marclop (approved):

kruskall (approved):

module/apmotel/go.mod
```diff
@@ -33,4 +33,4 @@ replace go.elastic.co/apm/v2 => ../..
 
 replace go.elastic.co/apm/module/apmhttp/v2 => ../apmhttp
 
-go 1.20
+go 1.19
```
dmathieu, line 36: In apmotel, this means we can remove the `// build` comments

dmathieu (approved):

PR #1466, opened 2023-06-12
Update golang.org/x/... to latest

axw:


marclop (approved):

dmathieu (approved):

axw:
precheck is failing because the version of golang.org/x/tools used by go-fastjson no longer matches what's in the top-level go.mod

I've got a PR up to update it (https://github.com/elastic/go-fastjson/pull/10), but I'll also update the generate script to use tools/go.mod for that code generation

PR #1477, opened 2023-06-28
FIX: properly set recording status when store apm span to otel

### LINKED ISSUE #1476: apmotel: native transaction are always considered non-recording when retrieved from Otel's `trace.SpanFromContext`
**Describe the bug**

After started a apm transaction, `apm.TransactionFromContext(ctx).Sampled()` and Otel's `trace.SpanFromContext(ctx).IsRecording()` returns different result.
the later is always `false`.

**To Reproduce**

```go
package main

import (
	"context"
	"fmt"
	"testing"

	"github.com/stretchr/testify/assert"
	"go.elastic.co/apm/module/apmotel/v2"
	"go.elastic.co/apm/v2"
	"go.elastic.co/apm/v2/apmtest"
	"go.opentelemetry.io/otel"
	"go.opentelemetry.io/otel/trace"
)

func Test_a(t *testing.T) {
	t.Parallel()
	tracer := apmtest.NewRecordingTracer()
	provider, err := apmotel.NewTracerProvider(apmotel.WithAPMTracer(tracer.Tracer))
	if err != nil {
		panic(err)
	}
	otel.SetTracerProvider(provider)

	ctx := context.Background()
	tx := tracer.StartTransaction("test1", "test")
	ctx = apm.ContextWithTransaction(ctx, tx)

	tx = apm.TransactionFromContext(ctx)
	sp := trace.SpanFromContext(ctx)
	fmt.Printf("apm transaction: %#v\n", tx)
	fmt.Printf("OTel span: %#v\n", sp)
	assert.Equal(t, tx.Sampled(), sp.IsRecording())
}
```

**Expected behavior**

Expect the assert to not fail.

wzy9607:
close #1476

### DIFF
--- a/module/apmotel/span.go
+++ b/module/apmotel/span.go
@@ -26,6 +26,7 @@ import (
 	"sync"
 	"time"
 
+	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
@@ -60,6 +61,12 @@ type span struct {
 
 	tx   *apm.Transaction
 	span *apm.Span
+}
+
+type noopSpan struct {
+	spanContext trace.SpanContext
+	tx          *apm.Transaction
+	span        *apm.Span
 }
 
 func (s *span) End(options ...trace.SpanEndOption) {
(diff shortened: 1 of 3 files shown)


dmathieu:
Thank you for the fix. The problem here is that while the span is marked as recording, it's unusable, since it's a noop.
If we provide a recording span, we should be able to set its attributes.

I have opened #1478.
However, it's not fully synced with the agent's spans/transactions. So if you create a span with the agent, then retrieve it using otel and make changes to it, and finally close the agent one, the changes made on the otel side won't be saved.

dmathieu:
We decided to go with #1478.

PR #1478, opened 2023-06-28
Return usable/recording spans from the agent context

dmathieu:
This is a fix for #1476.

### DIFF
--- a/module/apmotel/wrapper.go
+++ b/module/apmotel/wrapper.go
@@ -43,18 +43,28 @@ func init() {
 
 func contextWithSpan(ctx context.Context, apmSpan *apm.Span) context.Context {
 	ctx = oldOverrideContextWithSpan(ctx, apmSpan)
-	return trace.ContextWithSpanContext(ctx, trace.NewSpanContext(trace.SpanContextConfig{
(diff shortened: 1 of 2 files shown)


wzy9607:
This will likely result in nil pointer panic if some Otel instrumentation attempts to use the provider from `trace.SpanFromContext(ctx).TracerProvider()`.

I searched https://github.com/open-telemetry/opentelemetry-go-contrib, and found that one instrumentation indeed uses it. See 
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/df37a9d942c240f6c7ebdcee1900368728dd3adb/instrumentation/net/http/httptrace/otelhttptrace/clienttrace.go#L163-L176
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/df37a9d942c240f6c7ebdcee1900368728dd3adb/instrumentation/net/http/otelhttp/transport.go#L96-L102

dmathieu:
Good point. I've added the global tracer provider, if it's an apmotel one.
If it isn't one, we can't retrieve it from the global context.

axw (approved):

Issue #1484, opened 2023-07-07
fatal error: concurrent map writes


fridolin-koch:
**Describe the bug**

Just discovered this on our dev system, a services crashed due to:

```
fatal error: concurrent map writes

goroutine 6100 [running]:
go.elastic.co/apm/v2.droppedSpanTimingsMap.add(...)
	/go/pkg/mod/go.elastic.co/apm/v2@v2.3.0/transaction.go:457
go.elastic.co/apm/v2.(*Span).aggregateDroppedSpanStats(0xc001216fe8?)
	/go/pkg/mod/go.elastic.co/apm/v2@v2.3.0/span.go:613 +0x19a
go.elastic.co/apm/v2.(*Span).end(0xc001687440)
	/go/pkg/mod/go.elastic.co/apm/v2@v2.3.0/span.go:480 +0x47
go.elastic.co/apm/v2.(*Span).End(0xc001687440)
	/go/pkg/mod/go.elastic.co/apm/v2@v2.3.0/span.go:400 +0x351
go.elastic.co/apm/module/apmhttp/v2.(*responseBody).endSpan(0xc0045b0720)
	/go/pkg/mod/go.elastic.co/apm/module/apmhttp/v2@v2.3.0/client.go:198 +0x4f
go.elastic.co/apm/module/apmhttp/v2.(*responseBody).Read(0xc0045b0720, {0xc0045cd980?, 0xc0045cd600?, 0x0?})
	/go/pkg/mod/go.elastic.co/apm/module/apmhttp/v2@v2.3.0/client.go:187 +0x85
net/http.(*cancelTimerBody).Read(0xc0045b0740, {0xc0045cd980?, 0x19e9500?, 0xc0045cd600?})
	/usr/local/go/src/net/http/client.go:955 +0x33
io.(*LimitedReader).Read(0xc0045c86c0, {0xc0045cd980?, 0x0?, 0x7f30c89b25e8?})
	/usr/local/go/src/io/io.go:477 +0x45
io.ReadAll({0x1fb0ce0, 0xc0045c86c0})
	/usr/local/go/src/io/io.go:701 +0x103
io/ioutil.ReadAll(...)
	/usr/local/go/src/io/ioutil/ioutil.go:27
golang.org/x/oauth2/internal.doTokenRoundTrip({0x1fbe420, 0xc003c1aab0}, 0xc0000421af?)
[...]

PR #1487, opened 2023-07-11
fix: child span not acquiring transaction lock in some cases 

kruskall:
Spans from the same transaction share the dropped spans map. If they end concurrency a race condition could happen when updating the map.
Add a RW lock to prevent that.

Closes https://github.com/elastic/apm-agent-go/issues/1484

### DIFF
--- a/span.go
+++ b/span.go
@@ -400,6 +400,10 @@ func (s *Span) End() {
 			if s.tx != nil {
 				s.tx.mu.Lock()
 				defer s.tx.mu.Unlock()
+				if !s.tx.ended() {
+					s.tx.TransactionData.mu.Lock()
+					defer s.tx.TransactionData.mu.Unlock()
+				}
 			}
 			s.end()
 			return
--- a/span.go
+++ b/span.go
@@ -397,6 +397,10 @@ func (s *Span) End() {
 
 		if s.Type != parentType || s.Subtype != parentSubtype {
 			s.dropWhen(true)
+			if s.tx != nil {
+				s.tx.mu.Lock()
+				defer s.tx.mu.Unlock()
+			}
 			s.end()
 			return
 		}
--- a/span_test.go
+++ b/span_test.go
@@ -1048,6 +1048,7 @@ func TestDroppedSpanParentConcurrent(t *testing.T) {
 	// This test verifies there aren't any deadlocks on calling
 	// span.End(), Parent.End() and tx.End().
 	tracer := apmtest.NewRecordingTracer()
+	defer tracer.Close()
 	tracer.SetExitSpanMinDuration(0)
 
 	tx := tracer.StartTransaction("name", "type")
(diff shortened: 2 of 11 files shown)


dmathieu (approved):

marclop (commented):
Would you mind adding a test covering the bug?

kruskall:
I've added a test and discovered more issues. It seems this is affecting more than just the dropped spans.

I've pushed some changes and added a test.

span.go
```diff
@@ -397,6 +397,10 @@ func (s *Span) End() {
 
 		if s.Type != parentType || s.Subtype != parentSubtype {
 			s.dropWhen(true)
+			if s.tx != nil {
+				s.tx.mu.Lock()
+				defer s.tx.mu.Unlock()
```
marclop, line 402: Reading the godoc in `span.end()`, `s.tx.transactionData.mu` also needs to be held:

https://github.com/elastic/apm-agent-go/blob/326e364d6e52ea9601deb41a0004b633e1a03051/span.go#L466-L471

```suggestion
				defer s.tx.mu.Unlock()
				if !s.tx.ended() {
					s.tx.TransactionData.mu.Lock()
					defer s.tx.TransactionData.mu.Unlock()
				}
```

span_test.go
```diff
@@ -1044,6 +1044,40 @@ func TestCompressSpanSameKindConcurrent(t *testing.T) {
...
+			})
+			child.Context.SetDestinationService(apm.DestinationServiceSpanContext{Resource: fmt.Sprintf("foo%d", i)})
+			child.End()
+			wg.Done()
+		}(i)
+	}
+
+	// Wait until all the spans have ended.
+	wg.Wait()
+
```
marclop, line 1077: The parent and transaction should also end:

```suggestion
	parent.End()
	tx.End()
```

span_test.go
```diff
@@ -1044,6 +1044,40 @@ func TestCompressSpanSameKindConcurrent(t *testing.T) {
 	assert.Equal(t, 101, spanCount)
 }
 
+func TestDroppedSpanParentConcurrent(t *testing.T) {
+	// This test verifies there aren't any deadlocks on calling
+	// span.End(), Parent.End() and tx.End().
+	tracer := apmtest.NewRecordingTracer()
```
marclop, line 1050:
```suggestion
	tracer := apmtest.NewRecordingTracer()
	defer tracer.Close()
```

marclop (commented):
Thanks, this looks better!

simitt (commented):
Once this is ready to merge, please also add a changelog entry

kruskall:
@simitt could you clarify the above question ?

axw:
@kruskall in the past we have added entries to the Unreleased section. We've not been consistent about it though, and have sometimes just added the entries retrospectively at release time. Let's merge this and create a patch release - either adding the entry now, or creating the section at release time.

simitt:
Apologies @kruskall , I missed your message.

marclop (approved):

PR #1494, opened 2023-07-20
Update ELASTIC_APM_CLOUD_PROVIDER default value

yang-wei:
if i understand correctly the default value is "auto"
https://github.com/elastic/apm-agent-go/blob/main/utils.go#L179


axw (approved):
Can you please retarget this to main? We make all changes on main, and then merge back to 2.x on release.

axw:
Ah this probably needs to be recreated, there's a bunch of commits that we won't want in there.

yang-wei:
🙏 https://github.com/elastic/apm-agent-go/pull/1495

PR #1495, opened 2023-07-20
update default value of ELASTIC_APM_CLOUD_PROVIDER in doc

yang-wei:
if i understand correctly the default value is "auto"
https://github.com/elastic/apm-agent-go/blob/main/utils.go#L179


axw (approved):
thanks!

PR #1496, opened 2023-07-23
 docs: remove SECURITY.md to fallback to org-wide security policy

kruskall:
Align security policy with other repositories

Mention the public bug bounty program.

UPDATE: after looking into this, it seems other repositories are using the org-wide security policy from the health files repo. Remove the custom SECURITY.md to fallback to that.


axw (approved):
TIL you can do this. Would be good if the CoC were in there too.

PR #1503, opened 2023-08-29
[DOC+] Change example from elastic.com to elastic.co

stefnestor:
Per Elastic docs websites' feedback, changes elastic domain from ".com" to ".co" to match other examples.


axw (approved):
Oops. Thanks @stefnestor!

PR #1515, opened 2023-10-10
build: downgrade gin to 1.8.2

kruskall:


carsonip (approved):

Gh0u1L5:
@kruskall May I ask why are we downgrading the gin module? Do I have to downgrade gin version in my own projects if I rely on apmgin?

PR #1528, opened 2023-10-16
feat: add go workspace

kruskall:
Add support for go workspaces, getting rid of replace directives and making it easier to work with the repository (modern IDE and gopls have native support for workspaces)

### DIFF
--- a/.ci/updatecli.d/update-gherkin-specs.yml
+++ b/.ci/updatecli.d/update-gherkin-specs.yml
@@ -23,7 +23,15 @@ sources:
     transformers:
       - findsubmatch:
           pattern: "[0-9a-f]{40}"
-
+  pull_request:
+    kind: shell
+    dependson:
+      - sha
+    spec:
+      command: gh api /repos/elastic/apm/commits/{{ source "sha" }}/pulls --jq '.[].html_url'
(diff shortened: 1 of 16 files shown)


jamietanna:
I'd recently looked at adding a `go.work` to one of my own projects (https://github.com/deepmap/oapi-codegen) but there was a conversation about a `go.work` not necessarily being something you'd want to commit / distribute.

It's worth confirming this is what we want to do - it looks like we don't pull this as a library by anything else (internally at least) so it's probably safer than the use-case I've previously faced with this.

jamietanna:
I've traced back through the Gopher Slack, and there's a conversation https://gophers.slack.com/archives/C029RQSEE/p1692306223804279 and https://gophers.slack.com/archives/C029RQSEE/p1692373705463089?thread_ts=1692306966.974649&cid=C029RQSEE and see https://gophers.slack.com/archives/C029RQSEE/p1692374072836409 in particular and following messages around whether you should share a `go.work` with consumers of the library.

> Sorry, I might be misunderstanding this but the go-agent is heavily used throughout the organization 😅

You're absolutely right, I'd missed that we've got a nicer module import path, which threw off my searching.

PR #1531, opened 2023-10-25
perf: improve memory usage in apmhttp module

kruskall:
Few performance improvements related to memory usage:

- use simple string concat if there's no loop and we are concatenating a few strings (this is faster)
- use the Header map instead of calling `Set` if the key is a constant value (avoid alloc in canonical conversion)
- replace printf with more specific methods when possible (faster and less memory usage)
- let the runtime handle map realloc and do not prealloc a map with max size (this is wasted memory)

### DIFF
--- a/module/apmhttp/client.go
+++ b/module/apmhttp/client.go
@@ -139,11 +139,11 @@ func (r *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 func SetHeaders(req *http.Request, traceContext apm.TraceContext, propagateLegacyHeader bool) {
 	headerValue := FormatTraceparentHeader(traceContext)
 	if propagateLegacyHeader {
-		req.Header.Set(ElasticTraceparentHeader, headerValue)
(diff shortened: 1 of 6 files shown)


module/apmhttp/client.go
```diff
@@ -139,11 +139,11 @@ func (r *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 func SetHeaders(req *http.Request, traceContext apm.TraceContext, propagateLegacyHeader bool) {
 	headerValue := FormatTraceparentHeader(traceContext)
 	if propagateLegacyHeader {
-		req.Header.Set(ElasticTraceparentHeader, headerValue)
+		req.Header[ElasticTraceparentHeader] = []string{headerValue}
```
lahsivjar, line 142: +1 on the change but out of curiosity, won't they have the same allocs? Probably a bit high CPU cost in case of using `Set` due to canonical conversion of header keys but allocation wise I think they would be same.

lahsivjar (approved):

PR #1537, opened 2023-11-06
ci: use gvm instead of gimme 

v1v:
uses the script from https://github.com/elastic/beats/blob/024a9cec6608c6f371ad1cb769649e024124ff92/.ci/scripts/install-go.sh

`gimme` is not much maintained these days and 1.21 is not supported as per https://github.com/travis-ci/gimme/issues/210

I ran this [build ](https://buildkite.com/elastic/apm-agent-microbenchmark/builds/1002#018ba3c0-89a0-459f-bdf0-cccb4cd184a9) manually (it can only be accessed by Elastic) and so far it installed go using https://github.com/andrewkroh/gvm

### DIFF
--- a/scripts/ci/setenv.sh
+++ b/scripts/ci/setenv.sh
@@ -1,9 +1,47 @@
 #!/usr/bin/env bash
 set -euxo pipefail
 
-# Install Go using the same travis approach
-echo "Installing ${GO_VERSION} with gimme."
-eval "$(curl -sL https://raw.githubusercontent.com/travis-ci/gimme/master/gimme | GIMME_GO_VERSION=${GO_VERSION} bash)"
+# Install Golang using gvm
+echo "Installing ${GO_VERSION} with gvm."
+MSG="environment variable missing"
+GO_VERSION=${GO_VERSION:?$MSG}
+HOME=${HOME:?$MSG}
+OS=$(uname -s| tr '[:upper:]' '[:lower:]')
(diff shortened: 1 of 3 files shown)


reakaleek (approved):

scripts/ci/setenv.sh
```diff
@@ -1,9 +1,46 @@
...
+# Install Golang using gvm
+echo "Installing ${GO_VERSION} with gvm."
+MSG="environment variable missing"
+GO_VERSION=${GO_VERSION:?$MSG}
+HOME=${HOME:?$MSG}
+OS=$(uname -s| tr '[:upper:]' '[:lower:]')
+ARCH=$(uname -m| tr '[:upper:]' '[:lower:]')
+GVM_CMD="${HOME}/bin/gvm"
+
+if command -v go ; then
```
amannocci, line 13: Question: Does [`g`](https://github.com/stefanmaric/g) can be a good alternative?

v1v: Since `gvm` is maintained by @andrewkroh and it's well used for a few number of years by the ingest team, I'd say we can keep using `gvm` for now.

Probably, we could contribute to the `gvm` and facilitate the installation with a shell/bash script so all this magic won't be needed

v1v: https://github.com/andrewkroh/gvm/pull/67 is a tiny proposal

amannocci (approved):

PR #1575, opened 2024-02-17
security: add permissions block to workflows

reakaleek:
## Details

⚠️ This PR was created by an automated tool. Please review the changes carefully. ⚠️ 

We want to set the default permissions for workflows to read-only for contents. 
This is a security measure to prevent accidental changes to the repository.

This change adds a top-level permissions block to all workflows in the .github/workflows directory.
```yaml
permissions:
  contents: read
```

In some cases workflows might need more permissions than just `contents: read`.
Please checkout this branch and add the necessary permissions to the workflows.

If your workflow uses a Personal Access Token (PAT), we can still add the permissions block,
   but it will not have any effect.

Merging this PR as is might cause workflows that need more permissions to fail.

If there are any questions, please reach out to the @elastic/observablt-ci


axw (dismissed):

.github/workflows/labeler.yml
```diff
@@ -5,6 +5,9 @@ on:
   pull_request_target:
     types: [opened]
 
+permissions:
+  contents: read
```
v1v, line 9: `AlexanderWert/issue-labeler@v2.3` might require some write access?

reakaleek: Oh. Thanks for the warning. Because it's using `secrets.GITHUB_TOKEN` right?

reakaleek, line 9:
```suggestion
  contents: read
  issues: write
```

v1v (dismissed):

.github/workflows/snapshoty.yml
```diff
@@ -12,6 +12,10 @@ on:
     - '*.asciidoc'
     - 'docs/**'
 
+permissions:
+  contents: read
+  actions: read
```
v1v, line 17: what's the reason for `actions:read`? I don't recall we needed it

v1v (dismissed):

.github/workflows/snapshoty.yml
reakaleek: I re-ran the script on this repo, and for some reason, this was added. 
I don't know why it was missing since the workflow doesn't have a top-level permissions block.

Need to dig into this.

reakaleek: But yes, looking at the snapshoty-simple action, I also don't see a reason for it.

reakaleek:
```suggestion
```

dmathieu (approved):

PR #1581, opened 2024-03-12
Add write permission to labeler job

dmathieu:
The labeler needs to have the `pull-request: write` permission, so it can assign labels on PRs.

See https://github.com/elastic/apm-agent-go/actions/runs/8246861649/job/22553967383?pr=1580

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -8,6 +8,7 @@ on:
 permissions:
   contents: read
   issues: write
+  pull-requests: write
 
 jobs:
   triage:


dmathieu:
Since the labeler runs on `pull_request_target`, I think we won't see any change until the PR is merged (which is also why we didn't see any in #1575)

cc @reakaleek

reakaleek:
Thank you for the fix!

reakaleek (approved):

v1v (approved):

PR #1597, opened 2024-03-25
Slog Handler Elastic APM Integration

charliemenke:
Noticed that there is no included "built-in-instrumentation" for the ever more popular `slog` structured logging package. Wrote up a mostly simple implementation of a slog Handler that:

-  attaches any available trace/correlation ids to the log **when using context aware log methods**
- reports specific log level logs as errors to Elastic APM

Based my tests around the existing `apmlogrus` module so I hope I covered the most important cases.

LMK if I am missing anything critical, Thanks :) 

- Charlie


charliemenke:
signed CLA

module/apmslog/handler.go
```diff
@@ -0,0 +1,173 @@
...
+
+func (s *ApmHandler) tracer() *apm.Tracer {
+	if s.Tracer == nil {
+		return apm.DefaultTracer()
+	}
+	return s.Tracer
+}
+
+func (s *ApmHandler) levels() []slog.Level {
+	if s.ReportLevels == nil {
```
dmathieu, line 56: This value (and the one from `tracer()` above) is always set in `NewApmHandler`. So it shouldn't be possible for it to be nil.
How about removing those two methods and relying directly on the attributes then?

module/apmslog/handler.go
```diff
@@ -0,0 +1,173 @@
...
+	if slices.Contains(s.levels(), r.Level) && tracer.Recording() {
+
+		// attempt to find error/err attribute
+		// slog doesnt have a standard way of attaching an
+		// error to a record, so attempting to grab any attribute
+		// that has error/err as key and extracting the value
+		// seems like a likely way to do it.
+		var err error
+		r.Attrs(func(a slog.Attr) bool {
+			if a.Key == "error" || a.Key == "err" {
```
dmathieu, line 90: These values should at least be package constants.
Maybe this should also include any key which contains `err` (such as `http_error` is also an error)

module/apmslog/handler.go
```diff
@@ -0,0 +1,173 @@
...
+				if v, ok := a.Value.Any().(error); ok {
+					err = v
+					return false
+				}
+				if v, ok := a.Value.Any().(string); ok {
+					err = errors.New(v)
+					return false
+				}
+				return false
```
dmathieu, line 99: If the value is a string, it can be an error. But if somehow, it's an integer (or nil), it can't be?

```suggestion
				err = fmt.Errorf("%s", v)
				return false
```

module/apmslog/handler.go
```diff
@@ -0,0 +1,173 @@
...
+		// attempt to find error/err attribute
+		// slog doesnt have a standard way of attaching an
+		// error to a record, so attempting to grab any attribute
+		// that has error/err as key and extracting the value
+		// seems like a likely way to do it.
+		var err error
+		r.Attrs(func(a slog.Attr) bool {
+			if a.Key == "error" || a.Key == "err" {
+				if v, ok := a.Value.Any().(error); ok {
+					err = v
```
dmathieu, line 92: How about using `errors.Join` here to be able to handle multiple errors in the same log entry?

module/apmslog/handler.go
```diff
@@ -0,0 +1,173 @@
...
+				Message: r.Message,
+				Level:   strings.ToLower(r.Level.String()),
+				Error:   err,
+			}
+		} else {
+			errLogRecord = apm.ErrorLogRecord{
+				Message: r.Message,
+				Level:   strings.ToLower(r.Level.String()),
+			}
+		}
```
dmathieu, line 116: This can be simplified:

```suggestion
		errLogRecord := apm.ErrorLogRecord{
			Message: r.Message,
			Level:   strings.ToLower(r.Level.String()),
			Err: err,
		}
```

If `err` is nil, `errLogRecord.Err` will also be `nil`, which is the default value anyway.

dmathieu: You then don't even need to have an `errLogRecord` variable, you can just create `errlog`.

See the logrus implementation: https://github.com/elastic/apm-agent-go/blob/d6810ab0a60bfaaed9f35f98933877ebbfb3803c/module/apmlogrus/hook.go#L95

module/apmslog/handler.go
```diff
@@ -0,0 +1,173 @@
...
+
+		errlog := tracer.NewErrorLog(errLogRecord)
+		errlog.Handled = true
+		errlog.Timestamp = r.Time.UTC()
+		errlog.SetStacktrace(2)
+
+        // extract available trace info from context
```
dmathieu, line 123:
```suggestion
                // extract available trace info from context
```

module/apmslog/handler.go
```diff
@@ -0,0 +1,173 @@
...
+			errlog.TransactionID = tx.TraceContext().Span
+			errlog.ParentID = tx.TraceContext().Span
+		}
+		if span := apm.SpanFromContext(ctx); span != nil {
+			errlog.ParentID = span.TraceContext().Span
+		}
+		errlog.Send()
+	}
+
+	// attach trace context if exists and attach to record
```
dmathieu, line 135:
```suggestion
	// extract trace context if it exists and attach to record
```

module/apmslog/handler.go
```diff
@@ -0,0 +1,173 @@
...
+		errlog.Timestamp = r.Time.UTC()
+		errlog.SetStacktrace(2)
+
+        // extract available trace info from context
+		if tx := apm.TransactionFromContext(ctx); tx != nil {
+			errlog.TraceID = tx.TraceContext().Trace
+			errlog.TransactionID = tx.TraceContext().Span
+			errlog.ParentID = tx.TraceContext().Span
+		}
+		if span := apm.SpanFromContext(ctx); span != nil {
```
dmathieu, line 129: This transaction/span extraction logic happens twice, here and below to add the fields to the log record.
How about moving it slightly move to define local variables, so it's not duplicated?

module/apmslog/handler_test.go
```diff
@@ -0,0 +1,147 @@
...
+	return errors.New("new error")
+}
+
+func TestHandlerTracerClosed(t *testing.T) {
+	tracer, _ := transporttest.NewRecorderTracer()
+	tracer.Close() // close it straight away, hook should return immediately
+
+	h := newApmslogHandler(io.Discard, tracer)
+	logger := slog.New(h)
+	logger.Error("fail")
```
dmathieu, line 130: No expectation here? Wouldn't the test pass exactly the same way if we didn't call `tracer.Close()`?

dmathieu (commented):
My first feeling would be that we shouldn't add new instrumentations, since we're committing to OpenTelemetry.
But slog being a Go core logging library, I think having it makes sense.

Left a lot of comments, but the PR looks very nice overall. Thank you! 🌷

module/apmslog/handler.go
charliemenke: Since the `ApmHandler` Struct type is public someone could initialize is with a nil tracer which could cause panics when checking `tracer.Recording()` for example.

Though maybe I could see why we wouldn't want to override a user's selection unbeknownst to them.

if we remove  the `tracer()`  method I think we would then need to check if `h.Tracer != nil` before doing anything else with the `h.Tracer` struct  field.

We can remove the `levels()` method safely as a nil slice is just empty with no pointer to an underlying array so we should be able to safely use `slices.Contains`.

Thoughts?

module/apmslog/handler.go
charliemenke: I agree, having a hard time coming up with a regex or suffix match logic that would avoid false positives with Keys like `terror` or `dont_report_this_error`.

Also, not matter which direction we take, I should at least document the behavior of what key attributes are reported as APM Errors in the example file possible.

module/apmslog/handler.go
charliemenke: good catch!

module/apmslog/handler.go
charliemenke: agreed, I should be able to easily avoid this.

module/apmslog/handler_test.go
charliemenke: this is to test that the logger still works as expected (as a logger) even if the tracer closes. I will add an assert.

module/apmslog/handler.go
charliemenke: I am having a hard time managing to keep an error's StackTrace (if it exists) when using `errors.Join()`. I admit that my error handling skills are probably lacking in Go, but I dont see a great way of wrangling multiple errors in the slog record while keeping their respective stacktrackes.

Maybe instead of trying to consolidate all possible errors in one log into one `apm.ErrorLogRecord`, maybe if there are multiple "errors" in the log we send multiple `apm.ErrorLogRecord`s.

so for a log like: `{  "error_1":  "some error", "error_2":  "some  other error" }` we would report/send TWO `ErrorLogRecords` instead of trying to join them into one which would allow us to retain any specific error stack info.

module/apmslog/handler.go
dmathieu: We could also make the attributes private?

module/apmslog/handler.go
dmathieu: That does seem like a good idea.

dmathieu:
Note that there still seems to be a CLA issue.

charliemenke:
just a bump from my end.

- I will look into the CLA issue, not sure why its not saying i signed...
- ive been very busy with some other stuff this last week. Though I am still committed to working on this PR, hopefully will have some time soon to finish it up. Thanks for your patience  :)

charliemenke:
@dmathieu 
Thank you for your patience, I just pushed up changes to address the remaining comments.

I am not sure why the CLA is still not showing signed, I think I have signed it 4+ times at this point... If it is still showing unsigned, maybe it worth creating a new PR and see if that fixes it (or maybe I am doing something wrong).

If all looks good I will update this branch with master as it looks like there has been changes since I last created it.

per my commit message:
```
Implemented ability to report multiple apm errors from one log. If a user adds
multiple "reportable" error attributes to the log msg (default is "error" & "err"),
instead of trying to join the errors into one or discarding one, the apmslog
handler will report both errors.

Added ability for a user to define what slog attribute keys they want to report
as errors. Because there is no standard way in slog to attach an error to a msg
log, I wanted to add the ability for the user to decide what is and what is not
going to be reported. By default, slog attribute keys that are "error" or "err"
are reported, but with the new `WithErrorRecordAttrs(keys)` function a user
can define which keys will be reported.

Cleaned up `ApmHandler` struct and methods. Since we want the user to use the
included `NewApmHandler` function and its functional option functions, I
decided to make all Struct fields private.

Additionally added a check on if the `ApmHandler`'s `tracer` field is nill before
trying to use it. It is still possible for a user to pass in a nil tracer using
the `WithTracer` functional option.
```

module/apmslog/handler.go
charliemenke: I took your advice and made the fields private, though it is still possible to pass a nil `tracer` value using the `WithTracer` function so I just added a single check for `nil` where we use its value.

I did end up removing the `levels` method.

module/apmslog/handler.go
charliemenke: added a new field to `ApmHandler` so that users can define their own attribute keys that they want to report. Should give a good balance between ease of use and custom use cases.

module/apmslog/handler.go
charliemenke: implemented the ability to report multiple errors if available in the last commit.

module/apmslog/handler_test.go
```diff
@@ -0,0 +1,365 @@
...
+	logger.ErrorContext(ctx, "hello world", "error", mockFuncError())
+
+	span.End()
+	tx.End()
+
+	var msg MsgWithTrace
+	err := json.Unmarshal([]byte(buf.String()), &msg)
+	if err != nil {
+		t.Fatalf("unexpected error unmarshaling msg buff: %s", err)
+	}
```
dmathieu, line 168:
```suggestion
	assert.NoError(t, err)
```

module/apmslog/handler_test.go
```diff
@@ -0,0 +1,365 @@
...
+	logger.InfoContext(ctx, "hello world")
+
+	span.End()
+	tx.End()
+
+	var msg MsgWithTrace
+	err := json.Unmarshal([]byte(buf.String()), &msg)
+	if err != nil {
+		t.Fatalf("unexpected error unmarshaling msg buff: %s", err)
+	}
```
dmathieu, line 85:
```suggestion
	assert.NoError(t, err)
```

module/apmslog/handler_test.go
```diff
@@ -0,0 +1,365 @@
...
+	logger.ErrorContext(ctx, "hello world", "error", mockFuncError(), "err", errors.New("new error"))
+
+	span.End()
+	tx.End()
+
+	var msg MsgWithTrace
+	err := json.Unmarshal([]byte(buf.String()), &msg)
+	if err != nil {
+		t.Fatalf("unexpected error unmarshaling msg buff: %s", err)
+	}
```
dmathieu, line 232:
```suggestion
	assert.NoError(t, err)
```

charliemenke: fixed!

module/apmslog/handler_test.go
charliemenke: fixed

module/apmslog/handler_test.go
charliemenke: fixed!

charliemenke:
made the last changes and rebased into two relevant commits

dmathieu:
run docs-build

dmathieu (approved):

charliemenke:
Appreciate the reviews and all the suggestions! Have a good rest of the week :+1:

PR #1600, opened 2024-04-05
Internal package upgrade

kaanyalti:
Upgrading go-sysinfo package


kaanyalti:
> Hi @kaanyalti
> 
> the go agent is using go-sysinfo v1.7.1 which has no dependency on docker so it is not affected.
> 
> If you still want to get this merged in and bump the dependency could you rename the PR/commit title and fix the CI failure ?

I'm trying to fix the ci issues right now. I didn't realize that go-sysinfo 1.7.1 has no dependency on the docker package. I was mainly trying to go through the listed vulnerable dependencies in the snyk [report.](url) Although looking at it now, I may not have to do a more granular update, specifically update the apmhttp dependencies. Will convert this PR to a draft, and will resolve the issues first.

PR #1620, opened 2024-05-13
[Bugfix] gin web framework does not properly sanitize filename parameter of `Context.FileAttachment` function

sufatmawati:
The filename parameter of the Context.FileAttachment function is not properly sanitized. A maliciously crafted filename can cause the Content-Disposition header to be sent with an unexpected filename value or otherwise modify the Content-Disposition header. For example, a filename of "setup.bat";x=.txt" will be sent as a file named "setup.bat". If the FileAttachment function is called with names provided by an untrusted source, this may permit an attacker to cause a file to be served with a name different than provided. Maliciously crafted attachment file name can modify the Content-Disposition header.


dmathieu:
Thank you Suchi.
Note that the library is not impacted by this at all.
This PR upgrades gin (which should have happened within a few days from dependabot).

dmathieu (approved):

dmathieu:
run docs-build

dmathieu:
run docs-build

dmathieu (approved):

PR #1624, opened 2024-05-17
feat: update CODEOWNERS

kruskall:


dmathieu:
Should we then delete the @elastic/apm-agent-go team?
Alternative: we could keep the owners file as-is, and change the members of the team?

dmathieu:
cc @elastic/obs-ds-ebpf-go

rockdaboot:
> Should we then delete the @elastic/apm-agent-go team?
> Alternative: we could keep the owners file as-is, and change the members of the team?

The goal should be that people who are not directly involved in the maintenance of apm-agent-go or in the development of the lambda extension won't get GH notification emails ("reducing noise" for developers). And I don't think that every team member of @elastic/obs-ds-ebpf-go will be directly involved here. One way to achieve less noise is having more fine-grained GH groups.

Ideally, the decision on how to move forward is made by @AlexanderWert and @simitt.

dmathieu:
If we don't want everyone from obs-ds-ebpf-go to get notifications, then keeping the team and changing its members seems like the best way.

simitt:
>Ideally, the decision on how to move forward is made by @AlexanderWert and @simitt.

This is team internal to how the work is split up within the `obs-ds-ebpf-go` team, so leaving this to @AlexanderWert

PR #1625, opened 2024-05-27
added new instrumentation module for package `github.com/redis/go-redis/v9`

bars43ru:


dmathieu:
👎 
This agent is in maintenance mode, and does not accept new instrumentations anymore.
See the top of the readme: https://github.com/elastic/apm-agent-go/blob/main/README.md

Our recommendation is to migrate to OpenTelemetry instead. You can start using it alongside the elastic agent with our apmotel module.
Then, redis-go has its own otel instrumentation package: https://github.com/redis/go-redis/tree/master/extra/redisotel

PR #1645, opened 2024-08-17
feat: New Module for Go Redis client v9

Ishankhan21:
- Redis go client has new version https://pkg.go.dev/github.com/go-redis/redis/v9, but the elastic apm only has support for v8(https://pkg.go.dev/github.com/go-redis/redis/v8), 
- Created a new module in `module/apmgoredisv9` for the support.
- Added test case for same.


dmathieu:
See https://github.com/elastic/apm-agent-go/pull/1625

PR #1684, opened 2025-01-15
Bump minimum Go version to 1.22

rockdaboot:


dmathieu:
This is per the supported tech doc.
https://github.com/elastic/apm-agent-go/blob/main/docs/supported-tech.asciidoc

dmathieu (approved):

florianl:
From: https://github.com/elastic/apm-agent-go/blob/main/docs/supported-tech.asciidoc#go

```
Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release.
```

As the currently most recent Go version is Go 1.23, Go 1.21 should still be supported until the release of Go 1.24.

rockdaboot:
> From: https://github.com/elastic/apm-agent-go/blob/main/docs/supported-tech.asciidoc#go
> 
> ```
> Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release.
> ```
> 
> As the currently most recent Go version is Go 1.23, Go 1.21 should still be supported until the release of Go 1.24.

I read the docs as "we support version N and N-1", so 1.23 and 1.22 as of today.
"Go 1.5 was supported until the Go 1.7 release" means "with the Go 1.7 release, support for 1.5 is dropped" - so 1.6 and 1.7 (N and N-1) is supported.

florianl:
As a non-native speaker, my understanding of `Each major Go release is supported until there are two newer major releases.` might be different and would come down to N-2.

rockdaboot:
> As a non-native speaker, my understanding of `Each major Go release is supported until there are two newer major releases.` might be different and would come down to N-2.

I think the key is the "until" here. It means "to this point and not thereafter".
Example: 1.21 is supported until there are two higher major version. So as soon as we have two newer major versions, the support for 1.21 can be dropped. In this case it means if we have 1.22 and 1.23, the support for 1.21 can be dropped.

dmathieu:
When we set this up, we relied on the [otel-go](https://github.com/open-telemetry/opentelemetry-go/tree/main?tab=readme-ov-file#compatibility) policy, which matches the Go one.
It's as @rockdaboot describes it. N and N-1 versions are supported. So 1.21 is EOL.

rockdaboot:
> sorry for the confusion 🙇
> 
> the policy is the same as https://go.dev/doc/devel/release#policy, happy to improve the wording 🙂

@kruskall My suggestion would be to replace
```
For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release.
```
with
```
For example, Go 1.5 was supported until the Go 1.7 release and thus became EOL with the 1.7 release.
```

PR #1691, opened 2025-02-27
Run ci with fips check

dmathieu:
This does 3 things for FIPs compliance:

* Run the CI tests with the `GODEBUG` flag so the fips check is enabled.
  See https://go.dev/doc/security/fips140
* Disable custom certificate and CA chain when the fips build flag is enabled
* Prevent disabling certificate check when the fips build flag is enabled

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ docker-test:
 test-fips: ARGS=-tags=requirefips
 test-fips: test
 test:
-	@for dir in $(shell scripts/moduledirs.sh); do cd "$$dir" && go test -race -v -timeout=$(TEST_TIMEOUT) $(ARGS) ./... || exit $$?; done
+	@for dir in $(shell scripts/moduledirs.sh); do (cd $$dir && go test -race -v -timeout=$(TEST_TIMEOUT) $(ARGS) ./...) || exit $$?; done
 
 .PHONY: coverage
 coverage:
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ docker-test:
 test-fips: ARGS=-tags=requirefips
 test-fips: test
 test:
-	@for dir in $(shell scripts/moduledirs.sh); do (cd $$dir && go test -race -v -timeout=$(TEST_TIMEOUT) $(ARGS) ./...) || exit $$?; done
+	@for dir in $(shell scripts/moduledirs.sh); do cd "$$dir" && go test -race -v -timeout=$(TEST_TIMEOUT) $(ARGS) ./... || exit $$?; done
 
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -36,9 +36,11 @@ check-vet:
 docker-test:
 	scripts/docker-compose-testing run -T --rm go-agent-tests make test
 
-.PHONY: test
+.PHONY: test-fips test
+test-fips: ARGS=-tags=requirefips
+test-fips: test
 test:
-	@for dir in $(shell scripts/moduledirs.sh); do (cd $$dir && go test -race -v -timeout=$(TEST_TIMEOUT) ./...) || exit $$?; done
+	@for dir in $(shell scripts/moduledirs.sh); do (cd $$dir && go test -race -v -timeout=$(TEST_TIMEOUT) $(ARGS) ./...) || exit $$?; done
```
rockdaboot, line 43: - take care for `$dir` containing whitespace
- you don't need a subshell
```suggestion
	@for dir in $(shell scripts/moduledirs.sh); do cd "$$dir" && go test -race -v -timeout=$(TEST_TIMEOUT) $(ARGS) ./... || exit $$?; done
```

rockdaboot (dismissed):

Makefile
dmathieu: It's very unlikely we'll ever have a module with a whitespace in its name.

dmathieu: > /bin/sh: 1: cd: can't cd to ./module/apmazure

rockdaboot: OK, I didn't thought of that exiting the subshell restores to the previous working directory :+1:

rockdaboot: > It's very unlikely we'll ever have a module with a whitespace in its name.

Sure, but still let's do it the correct way. Code may be copied / spread.

rockdaboot, line 43:
```suggestion
	@for dir in $(shell scripts/moduledirs.sh); do (cd "$$dir" && go test -race -v -timeout=$(TEST_TIMEOUT) $(ARGS) ./...) || exit $$?; done
```

rockdaboot (approved):

transport/crypto.go
```diff
@@ -0,0 +1,97 @@
...
+package transport // import "go.elastic.co/apm/v2/transport"
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"os"
+
+	"github.com/pkg/errors"
```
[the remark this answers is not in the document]
dmathieu: I'd rather not do any behavior change in this PR. I'm happy to do that in another PR though.

PR #1693, opened 2025-02-28
Remove coverage ci check

dmathieu:
This check has been failing for months.
Its usefulness is also very limited, as the coverage results aren't sent anywhere, they're only displayed in the log output, making the data impossible to use in the current state.


rockdaboot (approved):
I agree that since the project is in maintenance mode, we don't need this check any more.

rockdaboot (approved):

PR #1697, opened 2025-03-13
Drop support for go 1.22

dmathieu:
Per our [Go version support](https://www.elastic.co/guide/en/apm/agent/go/current/supported-tech.html), we can drop support for Go 1.22, as it has reached EOL.
This upgrade also unblocks other dependency upgrades that require 1.23+.


dmathieu:
Note: I'll be cutting a new (minor, because of the fips PR) release after this.

PR #1801, opened 2026-04-09
Upgrade minimum version to go 1.25

dmathieu:
Go 1.24 has reached EOL.

### DIFF
--- a/internal/apmgodog/go.mod
+++ b/internal/apmgodog/go.mod
@@ -1,6 +1,6 @@
 module apmgodog/v2
 
-go 1.25.8
+go 1.25.0
 
 require (
 	github.com/cucumber/godog v0.12.2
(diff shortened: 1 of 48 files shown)


v1v (dismissed):

internal/apmgodog/go.mod
```diff
@@ -1,6 +1,6 @@
 module apmgodog/v2
 
-go 1.24.0
+go 1.25.8
```
kruskall: go agent is as a library, the go version here is the minimum supported version, we shouldn't enforce a toolchain version

dmathieu: I know. But there have been CVEs fixed there, and several folks complained in the otel area that we weren't enforcing them.
I'll go back to v0 is you strongly feel we shouldn't for the security fixes.

florianl (approved):

dmathieu:
@elastic/observablt-ci could one of you approve this?

fr4nc1sc0-r4m0n (approved):
LGTM

